PR #25, opened 2022-08-27
Added OpenWeatherMap widget.

AlexFullmoon:
I've little experience with JS, but looks like it's working. Still, please check the math in weather.jsx.

Using [current weather API](https://openweathermap.org/current), condition codes are [here](https://openweathermap.org/weather-conditions).

### DIFF
--- a/src/components/widget.jsx
+++ b/src/components/widget.jsx
@@ -1,8 +1,10 @@
 import Weather from "components/widgets/weather/weather";
+import OpenWeatherMap from "components/widgets/openweathermap/weather";
 import Resources from "components/widgets/resources/resources";
 
 const widgetMappings = {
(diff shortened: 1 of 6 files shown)


AlexFullmoon:
Okay, I've certainly missed something. Specifically, `widget.type` in pages/index.js and components/widget.jsx — should I add new or set it to `weather` for openweathermap?

benphelps:
> Okay, I've certainly missed something. Specifically, `widget.type` in pages/index.js — apparently, right now it is hardcoded to only one `weather` widget type, and I'm unsure on how to add another.

That's fine, it's an easy fix after I merge it.

Everything else looks fantastic, thank you so much!  I'm sure others will be happy to have options as well.

PR #34, opened 2022-08-28
Implement Jellyseerr integration

ilusi0n:
Hello!

Here is the implementation for jellyseerr.
Check it our since I never worked on next.js :D


benphelps:
Looks good to me!  Thank you :)

PR #100, opened 2022-09-11
Support Gotify Service

xicopitz:
Hi,

First of all, congratulations for the project.
I added this simple support for the self hosted notification service, Gotify (https://gotify.net/docs/pushmsg).

Requirements: 

1. Create a Client on Gotify web page
image

2. Add the service
```- Group Service:
    - One Service:
        href: http://gotify.host.or.ip
        icon: gotify.png
        description: Gotify
        widget:
            type: gotify
            url: http://gotify.host.or.ip
            key: clienttoken
```
3. Generated visual lists the configured applications, clients and messages unread
image

Note:  The file `src/utils/proxies/credentialed.js` should be reviewed because Gotify requires an header named `X-gotify-Key` and it should be handled dynamically not with a specific condition.

Thanks

### DIFF
--- a/src/components/widgets/resources/cpu.jsx
+++ b/src/components/widgets/resources/cpu.jsx
@@ -14,7 +14,7 @@ export default function Cpu() {
 
   if (error || data?.error) {
     return (
-      <div className="flex-none flex flex-row items-center justify-center">
+      <div className="flex-none flex flex-row items-center mr-3 py-1.5">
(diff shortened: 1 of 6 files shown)


benphelps:
Bring the locales up to date and only add the new entries to English and any other languages you can fluently translate.  Weblate will automatically make the other updates.

The change to the proxy is fine and is how I would have handled it.

PR #156, opened 2022-09-14
Add the Jackett widget

JazzFisch:
- add the follow-redirect package
- add the tough-cookie package

Jackett API uses a redirect mechanism to set a CSRF token. This CSRF token is stored in a cookie that is required to be present or the API won't work.


JazzFisch:
@benphelps hopefully the changes to http.js fits in with your vision of the http proxy.  I tried to find popular, well used, and rated modules to assist with handling redirects and cookies.

PR #183, opened 2022-09-16
Add backgroundOpacity option

andrii-kryvoviaz:
This fix improves readability with bright background images adding a tint.

Just add a new line to the `settings.yaml` (defaults to 1):
```yaml
backgroundOpacity: 0.5
```


benphelps:
Nice and simple, love it!

Issue #180, opened 2022-09-15
Lidarr widget not working


2ndwonder:
Lidarr widget is not working for me. Am I missing something? I don't see instructions for Lidarr.

```
    - Lidarr:
        icon: lidarr.png
        href: https://subdomain.domain.nl
        description: Music Managment
        server: docker
        container: lidarr
        widget:
          type: lidarr
          url: https://subdomain.domain.nl
          key: API KEY 
```
**If I uncomment the "key" section, it loads. But with API Error.**

<img width="1431" alt="image" src="https://user-images.githubusercontent.com/24813161/190519546-b711aac1-69e5-4d46-b2a6-1a601961d76c.png">

**with API key included**

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/24813161/190520139-eda6133e-9023-42f4-b1e7-bb6ceb49947b.png">

JazzFisch:
Is your Lidarr instance, perhaps, on a sub-path such as https://subdomain.domain.nl/lidarr?

2ndwonder:
No it Is not

JazzFisch:
Is there anything in the log output from the container?

benphelps:
The services proxy is reporting that the API key is unauthorized, it forwards the API response 1:1, so that's the actual error returned by the API.  I'd check and make sure your key is correct.

2ndwonder:
I double checked the API. It is correct. I'll try to generate another API key and see if that works

2ndwonder:
I changed the API and still the same results. 

@JazzFisch this is what I'm only getting from the logs from Homepage

<img width="717" alt="image" src="https://user-images.githubusercontent.com/24813161/190593779-cd3a94b6-f54b-45f0-9c49-1282d679fbff.png">

mtompkins:
I can confirm that the Lidarr widget chokes on large libraries. Perhaps this is in play for this issue.

benphelps:
Ohhh, it very well could be.  How large is your library?

This is also a problem with the Radarr and Sonarr integrations.

bjgeiser:
Mine is failing the same way for Lidarr and Readarr. Seeing this in the container logs. 

``type=lidarr&group=Media+Management+%28Cont%29&service=Lidarr&endpoint=album exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit` `

neildeadman:
My Lidarr is failling also:

[...]

PR #191, opened 2022-09-16
Proposal to add ability to map data in a proxy

JazzFisch:
This is a proof of concept on how to reduce the amount of data returned to the client via the existing proxy system, at the cost of additional CPU and memory utilization.  An upshot is that this could allow for clients to request data that you might not want to expose to those clients and potentially mask or remove any sensitive data.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,19 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Next.js: debug full stack",
+      "type": "node",
+      "request": "launch",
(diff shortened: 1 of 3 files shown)


JazzFisch:
@benphelps, when you have time, could you have a look and let me know what you think?  Would a better approach be to do what bjgeiser suggested in #180?

benphelps:
Okay, well that's not what I intended to do.

benphelps:
@JazzFisch I think this is the right direction, and will be needed for a lot of the pirate arr tools.

I made changes on how the endpoint is handled, which looks like this:

```js
readarr: {
    proxy: genericProxyHandler,
    maps: {
        book: (data) => jsonArrayMapper(data, (d) => ({ statistics: { bookFileCount: d.statistics.bookFileCount } })),
    },
},
```

That way we don't have to pass the endpoint around, and we can specify different maps for different endpoints easily.

benphelps:
The code inside of the proxy handler is still as simple,

```js
let resultData = data;
if (maps?.[endpoint]) {
  resultData = maps[endpoint](data);
}
```

Issue #152, opened 2022-09-13
[Feature Request] Qbittorrent widget


2ndwonder:
Could you make a widget for Qbittorrent? thank you in advance

Nonoss117:
Thanks👍 
![image](https://user-images.githubusercontent.com/79055577/190857135-31671261-8881-4ae2-a120-37c0c26a5a54.png)

Issue #123, opened 2022-09-12
[Feature Request] Widget for Kuma


PotatoPTT:

ShlomiPorush:
Add me to this request

ItsJustMeChris:
I will look into this over the weekend if there's nobody else interested in taking it on. 

Never really looked into AdGuard Home so I can't guarantee it will be done quickly, but I will try 😅

benphelps:
@eugenehw08 @PotatoPTT 

Added Adguard Home widget in the latest release.

eugenehw08:
hi,

can you update the information section in Service Widgets page?

benphelps:
Updated, https://github.com/benphelps/homepage/wiki/Service-Widgets#adguard-home

JazzFisch:
@ItsJustMeChris I didn't notice that you were interested on working on qBittorrent as well.  I was glossing over the titles and didn't read in to see your interest.  Sorry about that.

r0m1CH:
A widget for Kuma would be perfect

schnaxelhubr:
Being able to monitor the Uptime Kuma probes via a widget would be a very valuable addition

Alternatively - having a widget with a true IGMP ping would be also great.

IIIdefconIII:
would be very nice to have!

Jinxtrr:
+1 for uptime kuma as well!

NelsonDane:
Another +1 for uptime kuma

IIIdefconIII:
+1!

ozdeadmeat:
+4!!!!

joeryanh:
+1

ivybowman:
Extra +1 for uptime kuma

shamoon:
Appreciate all the interest but know that Kuma doesnt offer a JSON API yet which is almost a non-starter for this, until things change...

codeman1o1:
https://github.com/louislam/uptime-kuma/blob/a17084f75d1e5239ca2daf917d8a5002ef4953a8/server/routers/status-page-router.js#L29-L60
https://github.com/louislam/uptime-kuma/blob/a17084f75d1e5239ca2daf917d8a5002ef4953a8/server/routers/status-page-router.js#L62-L108
Could this be useful?

shamoon:
That doesnt look like something that could be used externally

See https://github.com/louislam/uptime-kuma/issues/118

[...]

PR #192, opened 2022-09-17
Add qBittorrent Widget

JazzFisch:
- extract cookie jar functionality into its own file
- use i18n for more strings in existing widgets

associated: #123
completes: #152


benphelps:
There is a specific i18n formatter for upload/download rates, `common.byterate`, you can see it in the speedtest and nzbget widgets for example.

Everything else looks good.

Issue #194, opened 2022-09-17
None of my API widgets are working


2ndwonder:
Did I miss something? using main tag. None of my API's are working suddenly 

<img width="1282" alt="image" src="https://user-images.githubusercontent.com/24813161/190847829-09374ccf-6ca3-49f3-95ac-ec69d1ff99e8.png">

<img width="1423" alt="image" src="https://user-images.githubusercontent.com/24813161/190847856-8c2cdf06-af8d-4a92-bbbf-3bfcc5d56d21.png">

Nonoss117:
I confirm, v0.3.51 broke radarr, sonarr, and tautully api. Overseerr, prowlarr is ok.

with v0.3.48 everything is working fine.
After update, without changing anything, widgets shows api error

![image](https://user-images.githubusercontent.com/79055577/190848067-bd43ae10-edcb-4cf2-93b3-b7de889e9498.png)

henkiewie:
Maybe switch to the latest instead of main label and they will come back. main is the for development and can change by the hour if i'm not mistaken. It worked for me.

benphelps:
@henkiewie is correct, do not use `main` unless you're fine with getting breaking changes.  `latest` is more stable and shouldn't break between updates.

benphelps:
@Nonoss117 no changes to API related stuff where part of the v0.3.51 update, so I'm unsure what the reason could be.  What are the container logs?

2ndwonder:
Yeah thought so, oh well I'll stay on main for now

Nonoss117:
> @Nonoss117 no changes to API related stuff where part of the v0.3.51 update, so I'm unsure what the reason could be. What are the container logs?

[homepage-2022-09-17T09-21-50.log](https://github.com/benphelps/homepage/files/9591343/homepage-2022-09-17T09-21-50.log)

benphelps:
The `main` tag should be functioning again now.

Nonoss117:
> The `main` tag should be functioning again now.

Yes, everthing is back to "normal" with main tag.
Thank's @benphelps

![image](https://user-images.githubusercontent.com/79055577/190852499-a9adb42c-c8b3-497d-b871-1b481fbac18e.png)

benphelps:
You've got a lot of movies, have you noticed a difference on `main` in how quickly it loads the stats information for Radarr?

@JazzFisch and I are trying out a new idea to consolidate API responses down into smaller formats for the widgets.

[...]

PR #198, opened 2022-09-17
Fix issues with incorrect values in widgets

JazzFisch:
associated: #180
associated: #194


benphelps:
Thank you, got tied up with Dockerfile shenanigans

Nonoss117:
Thank you

PR #211, opened 2022-09-18
Add authentik widget

JazzFisch:
I wanted to add disallowing proxied client calls to set HTTP verbs for generic and credentialed proxies to this PR as this could expose someone's authentik instance to calls that modify their authentication flows.  In the future I would like to investigate a better way to do more than the transform process we came up with in proxy.js and generic.js.


benphelps:
Agreed, it's not ideal, I added it because it's required for the Emby & Jellyfin remote control commands.

This will break those, but that's fine until we can figure out a better solution.

JazzFisch:
> Agreed, it's not ideal, I added it because it's required for the Emby & Jellyfin remote control commands.
> 
> This will break those, but that's fine until we can figure out a better solution.

Please hold off on this change then until we can come up with a way to proxy calls that would change underlying service state without allowing a malicious caller to proxy calls through homepage.

PR #228, opened 2022-09-20
Add Mastodon widget

DevPGSV:
* Added Mastodon service widget
* Also works with some other mastodon-like services like pleroma

Screenshot_20220920_034327


DevPGSV:
Added more commits in the same branch, closing

PR #229, opened 2022-09-20
Add Mastodon widget

DevPGSV:
* Added Mastodon service widget
* Also works with some other mastodon-like services like pleroma

191148909-1056d813-f367-41e5-94af-b066b52d1b12

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -168,5 +168,10 @@
     "jackett": {
         "configured": "Configured",
         "errored": "Errored"
+    },
+    "mastodon": {
+        "user_count": "Users",
+        "status_count": "Posts",
+        "domain_count": "Domains"
     }
 }
--- a/src/components/services/widget.jsx
+++ b/src/components/services/widget.jsx
@@ -27,6 +27,7 @@ import Gotify from "./widgets/service/gotify";
 import Prowlarr from "./widgets/service/prowlarr";
 import Jackett from "./widgets/service/jackett";
 import AdGuard from "./widgets/service/adguard";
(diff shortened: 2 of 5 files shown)


JazzFisch:
It appears that the Mastodon API uses an OAuth flow for authentication.  Is the `genericProxyHandler` the appropriate proxy handler?

DevPGSV:
The OAuth is only required for the endpoints that require authentication.
The endpoint used here is public, so I guess the genericProxyHandler might be the one to use.

benphelps:
I need an example widget entry, along with any details a user might need to know to get the widget hooked up, for the Wiki.

After that, I'll merge this in.

DevPGSV:
Currently in: https://homepage.librelabucm.org/
This is the service configuration we are using:

```yaml
    - Mastodon:
        href: https://mastodon.librelabucm.org/
        description: Mastodon Instance
        icon: mastodon
        server: [redacted]
        container: [redacted]
        widget:
            type: mastodon
            url: https://mastodon.librelabucm.org
```

To test it out with any other mastodon instance, change the `widget.url` to that of any server listed here: https://joinmastodon.org/servers
(Clicking "Join server" or "Get on waitlist" will open the instance url. The URL might contain "/about" in the login page, but the  `widget.url` is just `https://domain.tld` ).

For example, for: https://mastodon.social/about , the `widget.url` is `https://mastodon.social/`, and the widget appends the api endpoint to request https://mastodon.social/api/v1/instance


The 3 numeric values displayed by the widget ( user_count, status_count, domain_count ) are defined here: https://docs.joinmastodon.org/entities/instance/#stats

PR #230, opened 2022-09-20
Add Syncthing Relay Server widget

DevPGSV:
* Added Syncthing Relay Server widget

Screenshot_20220920_044309

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -172,7 +172,7 @@
     "strelaysrv": {
         "numActiveSessions": "Sessions",
         "numConnections": "Connections",
-        "dataRelayed": "Data Relayed",
-        "transferRate": "Transfer Rate"
+        "dataRelayed": "Relayed",
+        "transferRate": "Rate"
     }
 }
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -172,6 +172,7 @@
     "strelaysrv": {
         "numActiveSessions": "Sessions",
         "numConnections": "Connections",
-        "dataRelayed": "Data Relayed"
+        "dataRelayed": "Data Relayed",
+        "transferRate": "Transfer Rate"
     }
 }
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -168,5 +168,10 @@
     "jackett": {
         "configured": "Configured",
         "errored": "Errored"
+    },
+    "strelaysrv": {
+        "numActiveSessions": "Sessions",
+        "numConnections": "Connections",
+        "dataRelayed": "Data Relayed"
     }
 }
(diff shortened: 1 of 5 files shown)


JazzFisch:
Is `genericProxyHandler` the correct proxy to use?  According to https://docs.syncthing.net/dev/rest.html, this API is secured via the `X-API-Key` header.  Would `credentialedProxyHandler` not be a better choice?

DevPGSV:
That link point to the Syncthing client documentation.
The service used here wis the Syncthing Relay Server:

https://docs.syncthing.net/users/strelaysrv.html

The request is done to the address defined in:
--status-srv
That address only offers one endpoint: /status
That endpoint is public,it doesn't require authorization.
That endpoint is the one used by the following service:
https://relays.syncthing.net/

benphelps:
I need an example widget entry, along with any details a user might need to know to get the widget hooked up, for the Wiki.

After that, I'll merge this in.

benphelps:
I would suggest changing the last two stats to be just "Relayed" and "Rate", but I can do that after the merge, either way.

We try to keep them single words, since the translation is often much longer, but still needs to fit.

DevPGSV:
The service configuration I am using:

```yaml
    - Syncthing Relay Server:
        description: Syncthing Relay Server
        icon: syncthing
        server: [redacted]
        container: [redacted]
        widget:
            type: strelaysrv
            url: http://syncthing.librelabucm.org:22070
```

The command documentation is here: https://docs.syncthing.net/users/strelaysrv.html

We are currently running that Syncthing Relay Server by using: https://hub.docker.com/r/syncthing/relaysrv

Docker or not, the the flag "-status-srv=" defines the address for status endpoint with statistics. By default its value is `:22070`.
When run on a docker-compose, the flag can be set by using a custom entrypoint:

```yaml
services:
  strelaysrv:
    image: syncthing/relaysrv:latest
    network_mode: "host"
    entrypoint: ["/bin/entrypoint.sh", "/bin/strelaysrv", "-listen=0.0.0.0:22067", "-ext-address=syncthing.librelabucm.org:22067", "-status-srv=syncthing.librelabucm.org:22070"]
   volumes:
      - ./volumes/strelaysrv:/var/strelaysrv
   ports:
      - 22067:22067
      - 22070:22070
```

The site https://relays.syncthing.net/ periodically queries the endpoint of known Syncthing Relay Servers to display the statistics.
To test the widget with other servers, choose any other server in the list of https://relays.syncthing.net/ , and set `widget.url` to the Address of the service.  The port (usually :22067) displayed there is the port where the service runs. It has to be changed to `:22070`, the port where the statistics are available. In that port the statistics are shown at the endpoint `/status`.

For example, choosing from the list: `195.201.44.21:22067`, we would have to set `widget.url` to `http://195.201.44.21:22070`.
With that, the widget will query: http://195.201.44.21:22070/status

PR #253, opened 2022-09-23
Add global refresh interval for all SWR calls

carlosjdelgado:
Added global refresh interval for all widgets using SWR.

This allow to all widgets to auto refresh data each 1,5 seconds like in resource widgets except for weather widgets that not need refreshing data.

### DIFF
--- a/src/components/widgets/openweathermap/weather.jsx
+++ b/src/components/widgets/openweathermap/weather.jsx
@@ -11,8 +11,9 @@ function Widget({ options }) {
   const { t, i18n } = useTranslation();
 
   const { data, error } = useSWR(
-    `/api/widgets/openweathermap?${new URLSearchParams({ lang: i18n.language, ...options }).toString()}`
-  );
+    `/api/widgets/openweathermap?${new URLSearchParams({ lang: i18n.language, ...options }).toString()}`, {
+      refreshInterval: null
+    });
 
   if (error || data?.cod === 401 || data?.error) {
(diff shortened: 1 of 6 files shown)


benphelps:
We're aware that widgets don't auto refresh currently, and that's on purpose.

I'm hesitant to pull this in, as some people have a lot of widgets, and polling every 1.5 or even 3 to 5 seconds, can cause a waterfall of never ending requests.

As an example, let's take a homepage with 1 media streaming widget, 3 library monitors, 2 download clients and 1 request management tool.

That's 7 widgets, each with an average of 2.5 API calls needed to render.

That comes out to 17.5 API calls in total to render the widgets.

At a refresh of 1.5 seconds, that's 700 API calls per minute, or 12 API calls per second, on average.

Each of these API calls use extra CPU & memory on the platform hosting the API, as well as the platform hosting Homepage, as every API call is proxied.

benphelps:
I think this would need to be optional, and with a warning attached to it, for users with a lot of widgets.

shamoon:
Yes, definitely agree with benphelps, I would be pretty unhappy if updating my homepage install suddenly started making thousands of API calls per minute and personally I dont feel the need for refreshing most of them. Maybe I would use this at a much longer interval, but definitely would need to be user-controlled in some way (on/off, duration, etc).

Please feel free to re-open this if you are going to pursue that, and thank you for the contribution anyway!

PR #301, opened 2022-09-28
Feature: add category icons

maharsh9100:
Changes in this PR are as follows:
- Add custom icons to the title/category header[#215 ](https://github.com/benphelps/homepage/discussions/215)
- Updates the **services.yaml** file
- Export the **resolveIcon** function into Utils for reusability

Note: One of the commit message is "Update setting.yaml mapping" which is incorrect while the commit on is fine.


maharsh9100:
I'll work on resolving the conflicts.

maharsh9100:
I have resolved all the merge conflicts. Really great project!!! :smiley:

shamoon:
Thank you for the PR. I think some people may want this but this PR is a bit problematic imho as it alters the config schema i.e. it completely breaks an existing install's services... 

At the very least it would need some kind of migration (not a good idea I think), otherwise I would consider a different way to do this, for example making a separate `icons:` map in the yaml or something like that. Even though the current way is 'cleaner', a complete breaking change just for this seems incongruous.

Also, code needs cleanup but I will hold off on that for now as imho the above is a 'show-stopper', again just IMHO.

shamoon:
Have a look, I think its much simpler now to achieve this =)

`settings.yaml`:
```yaml
layout:
  Home Management:
    icon: https://cdn-icons-png.flaticon.com/512/25/25694.png
```

Of course please let me know if you disagree

shamoon:
Needs a squash, if merged.

PR #309, opened 2022-09-30
Allow widget field visibility to be configurable

JazzFisch:


benphelps (approved):
Very clever use of the i18n keys and children filtering.  Nice and clean, minimal changes required, love it :)

PR #366, opened 2022-10-10
Feature: Plex service widget without Tautulli

shamoon:
I didn't love the idea of installing Tautulli just for the widget goodness for Plex so I wrote a service widget that uses the Plex API directly. Its not as fancy as the Tautulli one but its still useful and is a little more 'standard'. If you guys dont like the approach here feel free to close this.

Screen Shot 2022-10-09 at 8 39 20 PM

Details:

- As noted in the wiki, the Plex API is not great so this merges a few different endpoints to get useful info (active streams, n movies and n tv shows).
- Caching is used to prevent too many calls. The active streams call is still every 5s (`refreshInterval`) but the libraries call is cached for 6 hours, the movie/tv counts for 10 minutes. Obviously these times can be tweaked but these seemed a reasonable balance.
- Plex API serves XML so this uses (lightweight!) https://www.npmjs.com/package/xml-js for conversion

`services.yaml:`
```yaml
    - Plex:
        href: https://app.plex.tv
        description: My Plex media server
        icon: plex.png
        widget:
            type: plex
            url: http://x.x.x.x:32400
            token: mytokenhere # see https://www.plexopedia.com/plex-media-server/general/plex-token/
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "shvl": "^3.0.0",
     "swr": "^1.3.0",
     "tough-cookie": "^4.1.2",
-    "winston": "^3.8.2"
+    "winston": "^3.8.2",
+    "xml-js": "^1.6.11"
   },
   "devDependencies": {
     "@tailwindcss/forms": "^0.5.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -40,6 +40,7 @@ specifiers:
   tough-cookie: ^4.1.2
   typescript: ^4.8.3
   winston: ^3.8.2
+  xml-js: ^1.6.11
 
 dependencies:
   '@headlessui/react': 1.7.2_biqbaboplfbrettd7655fr4n2y
@@ -65,6 +66,7 @@ dependencies:
   swr: 1.3.0_react@18.2.0
   tough-cookie: 4.1.2
   winston: 3.8.2
+  xml-js: 1.6.11
 
 devDependencies:
   '@tailwindcss/forms': 0.5.3_tailwindcss@3.1.8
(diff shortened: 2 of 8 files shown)


hunkyn:
Awesome. Thanks @shamoon

hunkyn:
It works for me as well. One minor suggestion, for all other widgets word "key" is used and this one is using word "token". Not a big deal, but if we are looking to be consistent then suggest changing to "key". Again not a big deal.

Nonoss117:
Same for me, lol.
I didn't read carefully the example above, and I first tried with "key". Then I changed to "token".

shamoon:
> It works for me as well. One minor suggestion, for all other widgets word "key" is used and this one is using word "token". Not a big deal, but if we are looking to be consistent then suggest changing to "key". Again not a big deal.

Ah yea I wasn't sure how strict that was, Plex calls it a token so I used that language. Let me know if you'd like a PR for that since this is already merged =)

hunkyn:
I will look for @benphelps to provide guidance in this.

JazzFisch:
@benphelps and @shamoon I wonder if we shouldn't work to standardize on a set of property names like:
* url
* key
* username
* password
* etc

benphelps:
> @benphelps and @shamoon I wonder if we shouldn't work to standardize on a set of property names like:
> 
> * key
> * username
> * password
> * etc

Yeah, that's basically what the unwritten rule has been, but perhaps its time to make it a written rule.

henryxrl:
Is it possible to show # of music as well? I believe many people use Plex for music organization as well (including me).

Milou020789:
Hello,

It's a nice widget, but can it count the movies from different libraries??? I have done it on my homepage but i only see my 39 series en not my 129 Movies

juanmanuelbc:
> Is it possible to show # of music as well? I believe many people use Plex for music organization as well (including me).

I would like it too, but I'm looking the Plex API and (if I'm not wrong) it seems that it only returns the number of artists...

I have some work done about this:
imatge

Maybe, if @shamoon can help me in obtaining the number of albums we can reach this goal (I hope so)...

juanmanuelbc:
I got it to work!!!

PR done: https://github.com/benphelps/homepage/pull/1418

PR #369, opened 2022-10-10
Feature: Glances info widget

shamoon:
One more! An info widget for [Glances](https://nicolargo.github.io/glances/) i.e. to remotely monitor another machine. Obviously this is built to match the `resources` info widget

Screen Shot 2022-10-11 at 8 15 57 AM

- Uses the [quicklook](https://glances.readthedocs.io/en/latest/api.html#get-quicklook) endpoint to make it just one http call.
- ~~Moved `UsageBar` to a common folder to share with resources.~~
- Allows basic auth if enabled for glances
- Allows multiple instances with optional `id` property

Part of #125

`widgets.yaml`:
```yaml
- glances:
    id: glances1 # if using multiple instances
    label: RaspPi # optional
```

`settings.yaml`:
```yaml
- glances1: # if using multiple instances or just 'glances' for a single instance
    url: http://x.x.x.x:61208
    username: user # optional
    password: pass # optional
```

### DIFF
--- a/src/pages/api/widgets/glances.js
+++ b/src/pages/api/widgets/glances.js
@@ -1,12 +1,19 @@
 import { httpProxy } from "utils/proxy/http";
 import createLogger from "utils/logger";
+import { getSettings } from "utils/config/config";
 
 const logger = createLogger("glances");
 
 export default async function handler(req, res) {
-  const { url } = req.query;
+  const settings = getSettings()?.glances;
+  if (!settings) {
+    logger.error("There is no glances section in settings.yaml");
+    return res.status(400).json({ error: "There is no glances section in settings.yaml" });
+  }
   
+  const url = settings?.url;
   if (!url) {
+    logger.error("Missing Glances URL");
     return res.status(400).json({ error: "Missing Glances URL" });
   }
 
--- a/src/pages/api/widgets/glances.js
+++ b/src/pages/api/widgets/glances.js
@@ -5,24 +5,35 @@ import { getSettings } from "utils/config/config";
 const logger = createLogger("glances");
 
 export default async function handler(req, res) {
-  const settings = getSettings()?.glances;
-  if (!settings) {
-    logger.error("There is no glances section in settings.yaml");
-    return res.status(400).json({ error: "There is no glances section in settings.yaml" });
+  const { id } = req.query;
+
+  let errorMessage;
+
+  let instanceID = "glances";
+  if (id) { // multiple instances
+    instanceID = id;
+  }
+  const settings = getSettings();
+  const instanceSettings = settings[instanceID];
+  if (!instanceSettings) {
+    errorMessage = id ? `There is no glances section with id '${id}' in settings.yaml` : "There is no glances section in settings.yaml";
+    logger.error(errorMessage);
+    return res.status(400).json({ error: errorMessage });
   }
   
-  const url = settings?.url;
+  const url = instanceSettings?.url;
   if (!url) {
-    logger.error("Missing Glances URL");
-    return res.status(400).json({ error: "Missing Glances URL" });
+    errorMessage = "Missing Glances URL";
+    logger.error(errorMessage);
+    return res.status(400).json({ error: errorMessage });
   }
 
   const apiUrl = `${url}/api/3/quicklook`;
   const headers = {
     "Accept-Encoding": "application/json"
   };
-  if (settings.username && settings.password) {
-    headers.Authorization = `Basic ${Buffer.from(`${settings.username}:${settings.password}`).toString("base64")}`
+  if (instanceSettings.username && instanceSettings.password) {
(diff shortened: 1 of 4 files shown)


src/pages/api/widgets/glances.js
```diff
@@ -0,0 +1,26 @@
+import { httpProxy } from "utils/proxy/http";
+import createLogger from "utils/logger";
+
+const logger = createLogger("glances");
+
+export default async function handler(req, res) {
+  const { url } = req.query;
```
JazzFisch, line 7: Rather than use URL as a param, can we instead look up the key (like provider or unifi_console) out of settings.yaml?  This would then disallow callers to essentially point your homepage instance at any glances server.  Attack vector is small, but I'd feel better operating from an allow list.

src/components/widgets/resources/memory.jsx
```diff
@@ -3,7 +3,7 @@ import { FaMemory } from "react-icons/fa";
 import { BiError } from "react-icons/bi";
 import { useTranslation } from "next-i18next";
 
-import UsageBar from "./usage-bar";
+import UsageBar from "../common/usage-bar";
```
JazzFisch, line 6: @benphelps would it make sense to move all of the common items now residing in src/components like errorboundry.jsx, etc. into the common folder?  Or would it make more sense to move usage-bar.jsx in to the src/components folder?

src/pages/api/widgets/glances.js
shamoon: Ah, fair enough. Yea this is a quirk of the current code that service widgets can get info without passing in the query whereas service widgets cant.

Which one do you prefer, I think a separate settings entry (like `unifi_console`) or `providers: glances: URL`? I like that the providers one feels more compact but the other providers seem to be API keys not URLs. Trying to abide the projects standards =)

shamoon: On second thought I think it should be a separate block as that will also allow other things. I should be able to make this change soon!

shamoon: All set on this part

JazzFisch: I think almost all of the settings we have in widgets.yaml should be moved in to settings.yaml.  That or we take an approach like there is in services.yaml where there is an allow list of properties from that is emitted to the client.

JazzFisch:
For configs, what about something like this.  Which seems like I'm advocating for a new property... _sigh_
`widgets.yaml`:
```yaml
- glances:
    label: Pi
    system: silliness

- glances:
    label: Sbc
    system: other
```

`settings.yaml`:
```yaml
glances:
  silliness:
    url: ...

  other:
    url: ...
```

I don't know what to use as a lookup key for `widgets.yaml` though.  `system`, `server`, `jason_was_here`?

src/pages/api/widgets/glances.js
```diff
@@ -0,0 +1,33 @@
+import { httpProxy } from "utils/proxy/http";
+import createLogger from "utils/logger";
+import { getSettings } from "utils/config/config";
+
+const logger = createLogger("glances");
+
+export default async function handler(req, res) {
+  const settings = getSettings()?.glances;
```
JazzFisch, line 8: Right now, this only allows for a single glances instance.  Was that intended?

shamoon: Actually, think I have a simple solution for this, at least for now

shamoon: Ok let me know what you think, I added an optional `id` property, if passed thats what it uses to look up the settings in `settings.yaml`. This would be a solution for this PR for now so it doesnt get snagged on any other changes.

shamoon: (I realize its quite similar to your idea, I guess pros / cons to nesting all instances under `glances`, lmk =)

JazzFisch: I have meetings pretty much nonstop for the next 7 hours.  I'm not going to be able to look for a while.  Sorry for the delay.  ☹️

shamoon: No worries of course, I should probably be doing other things too... 😅

src/components/widgets/resources/memory.jsx
benphelps: Without just refactoring everything, including the current resources, I think it's fine to leave it where it is now and reference it.

shamoon: As in you'd prefer I move it back and use `import UsageBar from "../resources/usage-bar";` ? Easy enough of course just clarifying!

shamoon: I went ahead and changed this back. Should be all set!

src/pages/api/widgets/glances.js
benphelps: Send me a DM or friend request on Discord @shamoon, phelps#0001

PR #370, opened 2022-10-10
MDI - Icons as YAML Setting

dieideeistgut:
Heya,

this is my addition of MDI icons to homepage. I am no expert in this field - feel free to decline. Maybe this serves ans an idea. MDI-Syntax i use is:

```
- Stockholm:
    - ShyNet:
        description: ShyNet Tracking
        icon: 'mdi mdi-chart-timeline-variant-shimmer'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.7.2",
+    "@mdi/font": "^7.0.96",
     "classnames": "^2.3.2",
     "compare-versions": "^5.0.1",
     "dockerode": "^3.3.4",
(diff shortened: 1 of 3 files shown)


shamoon:
I think react-icons has md icons? https://react-icons.github.io/react-icons/icons?name=md

dieideeistgut:
Didn't know that. What would be a decent syntax for implementing those via `icon` ?

benphelps:
Including the MDI font and CSS is too heavy, we use React Icons as that allows tree shaking so we only include the icons being used in the final bundled application.

I played around with this a bit and got something working that fetches the SVG icons from the jsdeliver CDN instead.

Since we can't style SVGs that are inside of an  tag, we have to resort to some more creative alternatives.  This is using the logo gradient stops to color a div and then using the SVG as a mask.

Screenshot 2022-10-11 at 2 19 17 PM
Screenshot 2022-10-11 at 2 18 50 PM
Screenshot 2022-10-11 at 2 18 43 PM
Screenshot 2022-10-11 at 2 18 36 PM
Screenshot 2022-10-11 at 2 18 21 PM

dieideeistgut:
Looks nice, works sweetly. Many thanks for this.

PR #378, opened 2022-10-12
Add option to specify locale for datetime widget

MrLemur:
I have found that i18n-next is wrongly setting the date in `en-US` format when it should be `en-GB`.

This should add an option to specify the locale, and then fall back if a value is not provided in the config:
```yaml
- datetime:
    text_size: sm
    locale: en-GB
    format:
      dateStyle: short
      timeStyle: short
      hour12: false
```

### DIFF
--- a/src/components/widgets/datetime/datetime.jsx
+++ b/src/components/widgets/datetime/datetime.jsx
@@ -13,7 +13,7 @@ const textSizes = {
 };
 
 export default function DateTime({ options }) {
-  const { text_size: textSize, format } = options;
+  const { text_size: textSize, locale, format } = options;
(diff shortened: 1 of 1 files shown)


benphelps:
Is there a reason to not just set the sites entire language to `en-GB`?

With the language set to `language: en-GB` inside my `settings.yaml` file, I get the following date format:

`12/10/2022, 17:06`

Where the default `en` or (`en-US` in this case, since that's the fallback) gives:

 `10/12/22, 5:06 PM`

MrLemur:
@benphelps I can't believe I missed that setting! Works as expected, sorry for the non-issue!

PR #379, opened 2022-10-12
Fix the page auto zoom when ios input focus

GodD6366:
Fix the page auto zoom when ios input focus


benphelps:
Thanks for this fix :)

PR #388, opened 2022-10-14
Feature: Service ping, redesign docker status

### LINKED ISSUE #78: [Feature Request] Site up or down indicator
Similar to what uptime kuma would do it would be nice to use the green light you use for container status to indicate if a website is up or down (404 / 502 / 300 / etc). I use uptime kuma as the example but it doesn't need to be that sophisticated a simple yes/no of if a site is up or down would be helpful.

### LINKED ISSUE #248: [Feature Request] Ping for bookmarks
it would be nice if the docker traffic lights could be used in the bookmarks to show a ping status.

so when the page is loaded a ping is sent out from the browser and then the traffic lights could show, green for 200 response  amber for awaiting response and red for non 200 responses

shamoon:
As discussed, this (draft for now) PR implements the ping which uses a `HEAD` request, also shows estimated latency if status is 200. Changes look of docker status too. As noted users would need to be careful about which URL to ping, if the public domain is behind a reverse-proxy this will potentially falsely always return status 200, instead local URL / ip could be specified to test http status.

Closes #248, closes #78

Screen Shot 2022-10-14 at 2 24 54 PM

`services.yaml`
```yaml
    - Paperless-ngx:
        href: https://paperless.domain.com
        description: My awesome Paperless-ngx instance
        icon: paperless.png
        ping: https://google.com
        server: docker
        container: paperless
```

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -47,7 +47,13 @@
         "tx": "TX",
         "mem": "MEM",
         "cpu": "CPU",
-        "offline": "Offline"
+        "offline": "Offline",
+        "error": "Error",
+        "unknown": "Unknown"
+    },
+    "ping": {
+        "error": "Error",
+        "ping": "Ping"
     },
     "emby": {
         "playing": "Playing",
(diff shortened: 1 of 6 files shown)


Shagon94:
@shamoon I can see how ping can be useful, would it be possible to have the latency be an optional field? The flexibility behind having the latency being optional seems great.

shamoon:
It could, sure, though more complexity isnt always better. Really just depends on what folks want. The current idea is that it communicates both status + latency with color / response code etc.

PR #408, opened 2022-10-18
Feature: support target in individual services/bookmarks

Facinorous-420:
Allows the user to add `target:` to specific individual service and bookmarks, so it will treat specific links similar to the `target:` that can be set in the settings file.

An example service with this PR in mind would be:

```
- Apps:
    - Grafana:
        href: https://example.com/
        icon: grafana
        description: System statistics and graphs
        server: docker2
        target: _self
```

If no target is added in the service/bookmark,  it continues to default to `_blank` like it normally does.

Will need to update the wiki to add this information to the services and bookmark pages, if this is accepted. Thanks.

### DIFF
--- a/src/components/bookmarks/item.jsx
+++ b/src/components/bookmarks/item.jsx
@@ -11,7 +11,7 @@ export default function Item({ bookmark }) {
       <a
         href={bookmark.href}
         title={bookmark.name}
-        target={settings.target ?? "_blank"}
+        target={settings.target ?? bookmark.target ?? "_blank"}
         className="block w-full text-left cursor-pointer transition-all h-15 mb-3 rounded-md font-medium text-theme-700 dark:text-theme-200 dark:hover:text-theme-300 shadow-md shadow-theme-900/10 dark:shadow-theme-900/20 bg-theme-100/20 hover:bg-theme-300/20 dark:bg-white/5 dark:hover:bg-white/10"
       >
         <div className="flex">
--- a/src/components/bookmarks/item.jsx
+++ b/src/components/bookmarks/item.jsx
@@ -11,7 +11,7 @@ export default function Item({ bookmark }) {
       <a
         href={bookmark.href}
         title={bookmark.name}
-        target={settings.target ?? bookmark.target ?? "_blank"}
+        target={bookmark.target ?? settings.target ?? "_blank"}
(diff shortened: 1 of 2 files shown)


src/components/bookmarks/item.jsx
```diff
@@ -11,7 +11,7 @@ export default function Item({ bookmark }) {
       <a
         href={bookmark.href}
         title={bookmark.name}
-        target={settings.target ?? "_blank"}
+        target={settings.target ?? bookmark.target ?? "_blank"}
```
JazzFisch, line 14: Please change from most specific to least specific.  In other words
```jsx
target={bookmark.target ?? settings.target ?? "_blank"}
```

src/components/services/item.jsx
```diff
@@ -71,7 +71,7 @@ export default function Item({ service }) {
             (hasLink ? (
               <a
                 href={service.href}
-                target={settings.target ?? "_blank"}
+                target={settings.target ?? service.target ?? "_blank"}
```
JazzFisch, line 74: Please change from most specific to least specific.  In other words
```jsx
target={service.target ?? settings.target ?? "_blank"}
```

src/components/services/item.jsx
```diff
@@ -84,7 +84,7 @@ export default function Item({ service }) {
           {hasLink ? (
             <a
               href={service.href}
-              target={settings.target ?? "_blank"}
+              target={settings.target ?? service.target ?? "_blank"}
```
JazzFisch, line 87: Please change from most specific to least specific.  In other words
```jsx
target={service.target ?? settings.target ?? "_blank"}
```

JazzFisch (changes requested):
Neat idea, though can you please update the ordering of the evaluations?

Facinorous-420:
> Neat idea, though can you please update the ordering of the evaluations?

My use case which brought this idea in was chaining homepage instances together, meanwhile leaving all other links as open in new tab. 
Thanks for the review! Ive done those changes. Everything should be golden now!

Facinorous-420:
> Neat idea, though can you please update the ordering of the evaluations?

My use case which brought this idea in was chaining homepage instances together, meanwhile leaving all other links as open in new tab. 
Thanks for the review! Ive done those changes. Everything should be golden now!

src/components/services/item.jsx
```diff
@@ -84,7 +84,7 @@ export default function Item({ service }) {
           {hasLink ? (
             <a
               href={service.href}
-              target={settings.target ?? "_blank"}
+              target={service.target ?? settings.target ??  "_blank"}
```
shamoon, line 87:
```suggestion
              target={service.target ?? settings.target ?? "_blank"}
```

shamoon (changes requested):
Sorry just spaces for my brain, feel free to squash these commits =)

PR #428, opened 2022-10-22
fixed issue causing preventing custom search from working

jshwrner:
When using custom search option, for example (search text=asdf):

```
- search:
    provider: custom
    url: https://qwant.com/?q=
    target: _blank
```

the the browser would try to redirect to `http(s)://myhost/undefinedasfe`. It appears this is because the `url` property is getting stripped out in the `cleanWidgetGroups` function. This causes the `url` property to get lost and not actually overwrite the `false` value set in `Search.jsx`


shamoon:
Thanks, we have a fix for this in the works already =)

jshwrner:
@shamoon, didn't realize, thanks!

PR #442, opened 2022-10-24
feature: add homebridge widget

Fernando-Neira:
Simple homebridge widget,

How to use:

```javascript
    - Homebridge:
        icon: homebridge.png
        container: homebridge
        service: homebridge
        href: http://your-ip-address:8581
        description:
        widget:
            type: homebridge
            url: http://your-ip-address:8581
            username: your-username
            password: your-password
```

Preview:

image

If you dont have child bridges:

image

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -284,5 +284,13 @@
         "96-night": "Thunderstorm With Hail",
         "99-day": "Thunderstorm With Hail",
         "99-night": "Thunderstorm With Hail"
+    },
+    "homebridge": {
+        "available_update": "System",
+        "update_available": "Update Available",
+        "up_to_date": "Up to Date",
+        "available_homebridge_update": "Plugins",
+        "plugins_updates_available": "{{quantity}} Available",
+        "plugins_up_to_date": "Up to Date"
     }
 }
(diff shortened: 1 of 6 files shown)


shamoon:
Nice, it works! =)

Couple thoughts:

1. The API (swagger) allows for basic auth, so you dont actually need the whole proxy here...
2. It could be more useful than the current items to show things like # of child bridges which many use (myself included). We could make it only show if n > 0? Only problem is its another API call. `/api/status/homebridge/child-bridges`. I'll have a think
3. Technically, this requires Config UI X (thats where the API comes from). Most people, especially at this point, have this, but its not actually required. We can note this in the docs.

Let me know your thoughts on # 2, I can help out with # 1 if you'd like

shamoon:
Sorry may have spoken too soon, 1 sec!

Fernando-Neira:
@shamoon  update the preview image, and add child_bridges api support :)

shamoon:
Love it! I'll take a look later when I have some real time, thanks for the pr!

shamoon:
Just some minor stuff, works great though =)

I will take a final look at it tonight, thanks again

shamoon:
Nice work!

PR #459, opened 2022-10-27
Feature: widget for displaying scripted information

StefanTT:
This widget executes a script and displays the script's output. It allows to create simple widgets without much programming knowledge.

Example display:
screenshot

The configuration for this is:
```yaml
  widget:
      type: scripted
      script: "mcstatus myserver:25565 json"
      fields: [ "online", "player_count", "ping" ]
      field_labels:
          player_count: "players"
      field_types:
          online: boolean
          ping:
              type: common.ms
              style: unit
              unit: millisecond
              unitDisplay: narrow
```

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -298,5 +298,9 @@
         "rejectedPushes": "Rejected",
         "filters": "Filters",
         "indexers": "Indexers"
+    },
+    "scripted": {
+        "yes": "Yes",
+        "no": "No"
     }
 }
--- a/src/widgets/components.js
+++ b/src/widgets/components.js
@@ -30,6 +30,7 @@ const components = {
   readarr: dynamic(() => import("./readarr/component")),
   rutorrent: dynamic(() => import("./rutorrent/component")),
   sabnzbd: dynamic(() => import("./sabnzbd/component")),
+  scripted: dynamic(() => import("./scripted/component")),
   sonarr: dynamic(() => import("./sonarr/component")),
   speedtest: dynamic(() => import("./speedtest/component")),
   strelaysrv: dynamic(() => import("./strelaysrv/component")),
--- a/src/widgets/scripted/README.md
+++ b/src/widgets/scripted/README.md
@@ -1,0 +1,50 @@
+# Widget for displaying scripted information
+
+This widget executes a script and displays the script's output. The executed script must return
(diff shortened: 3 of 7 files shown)


shamoon:
This is cool as a proof of concept but I have some concerns about it.

- The usability. Frankly, it doesnt feel particularly obvious or friendly, I've read it a couple times and still feel like it'd require trial-and-error.
- Security. Yes, this is something that would need to be setup by a user so theres a "use-at-your-own-risk" element to it, but frankly, just allowing passing random scripts to `execSync` makes me... nervous.

Would be curious to hear others' thoughts.

StefanTT:
The usability is not very obvious, especially when it comes to more complex formatting like the ping field in the example.

StefanTT:
Thank you very much for your comments. I would like to write down some of my thoughts, maybe it helps to better understand the points.

* The usability is not very obvious, I fully agree. Especially when it comes to more complex formatting like the ping field in the example. This can be changed of course, I tried to keep the code as small as possible.

* Security: I do not see a way how a web user could abuse the functionality as there are no variable parts in the script call. Sensible information might be exposed. But unused fields are filtered server side, so sensible information would be visible on the web page too and not be hidden in the traffic.

I see this widget as some sort of a helper to get some information for services that do not have an integration yet. And maybe never will get one because it's such an edge case that not many will need. If you are an admin but not a web developer than you can help yourself with it to get something going.

shamoon:
Thanks for your reply and for the effort, but I have to say I would not be in favor of merging this PR. It feels a bit out of the scope of this project as I see it and I foresee a **lot** of issues supporting this. Again, this is just my opinion.

JazzFisch:
> Thanks for your reply and for the effort, but I have to say I would not be in favor of merging this PR. It feels a bit out of the scope of this project as I see it and I foresee a **lot** of issues supporting this. Again, this is just my opinion.

I concur with shamoon on this one.  One of the appeals of homepage is it's simplified (as much as reasonable) configuration structure.

benphelps:
Very cool idea, but I do agree with @JazzFisch and @shamoon.  Supporting this would be a nightmare and the added security responsibility doesn't sit right.

What I'd recommend is creating your own new project that allows users to define function to run and expose the results through an API (much like what you have here).  We'd be more than happy to accept a PR that integrates that API, and you'd get your own little project that others could integrate with as well, not just homepage.

PR #464, opened 2022-10-29
Fix endpoint typo in emby widget

ForeverEndeavor:


shamoon:
Cool thanks. Is the play thing not working right now and guessing you can confirm this fixes it? (Looks like it would)

ForeverEndeavor:
Correct - PlayControl was looking for "endpoint" but got "enpoint" instead.

PR #468, opened 2022-10-30
Fix datetime widget size issue

stuffinator:
- the widget was missing a flex-col div around it and was therefore taking up more space than it needed as it was reported here https://github.com/benphelps/homepage/issues/235#issuecomment-1276209354


shamoon (dismissed):
Thanks

PR #485, opened 2022-11-04
Add OctoPrint widget

donslice:
- Tracking Job Status (state), File Name, Job Progress (pct)
firefox_stC2fKktKu

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -315,5 +315,10 @@
         "uptime": "Uptime",
         "alerts": "Alerts",
         "time": "{{value, number(style: unit; unitDisplay: long;)}}"
+    },
+    "octoprint": {
+        "job_status": "Job Status",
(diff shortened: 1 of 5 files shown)


shamoon:
Thanks for the PR, this seems like a rather specific one and sort of limited use since it only shows something _while_ printing is in progress...

We talked about it and if there are a lot of requests for this widget perhaps we can re-address. Thank you again.

donslice:
Thanks for the feedback. I'll continue to work on it and see if there is anything else I can add to expand the functionality.

shamoon:
Thanks for your understanding :) Yeah in general with these widgets, it’s not like there’s exact criteria of course but it’s some combination of community interest and “seems useful”. But clearly there is subjectivity to it.

Ps. one way to gauge interest is to check discussions/feature requests and see what things people are looking for.

Issue #291, opened 2022-09-27
[Feature Request] Use icons istead of abbrevation 


AdrienVIN:
Like the [Flame project](https://github.com/pawelmalak/flame) we could put a [materialdesignicons.com](https://materialdesignicons.com/) icon instead of a 2 letters abbreviation. 

But I don't really know the complexity behind it...

warmbo:
It would also be cool to just pull the favicon from the URL used as well.

JazzFisch:
This is for the bookmarks, yes?

AdrienVIN:
Yes for the bookmarks !

PR #487, opened 2022-11-04
Adds the option for a bookmark icon

donslice:
#291 

Same functionality as service icons. Falls back to abbreviation.

firefox_x414AUoKJk


shamoon:
Thanks for the PR, I updated this to use the shared functionality for `ResolvedIcon`, and rebased. Its a very straightforward PR now =)

PR #510, opened 2022-11-09
adding malay into translations list

eymankun:
updated Readme.md to add malay languange into translation list.


shamoon:
Nice! And thanks for translating

PR #511, opened 2022-11-09
Fix: openweathermap day/night icon issue

arimitx:
The day/night icon does not display properly in openwathermap widget due to the typo (sundown -> sunset) in parsing data from openweathermap api. The icon would be displayed properly after the fix.


shamoon:
Cheers thanks

PR #535, opened 2022-11-18
Added configurable setting for log folder

theeternalrat:
Added a setting `logpath` that when set, will allow the administrator to specify a log location on disk. If not set, the value will remain the config directory as before.

Why necessary:

1) Its a best practice to be able to configure the log place.
2) Logs probably shouldn't be stored in the config directory.
3) For certain platforms, the config directory might not be writable, no widgets appear to work when this is the case


shamoon (changes requested):
Please fix the lint errors

shamoon (approved):

PR #543, opened 2022-11-21
Fix error on refresh sending DATA var undefined on first time on widget ChangeDetection.IO

brunoccr:


shamoon:
Makes sense looking at the code, can you confirm the issue is you’re seeing an error at first?

shamoon (approved):

PR #556, opened 2022-11-25
Add paperless-ngx widget

Tarow:
Added a widget for [paperless-ngx](https://github.com/paperless-ngx/paperless-ngx). It should also be compatible with paperless-ng (and paperless?), as i didn't see any difference in the api endpoints, that this widget is using.

Config:
```
  - Paperless-ngx:
      description: Document Management
      icon: paperless
      href: https://paperless.somedomain.com
      widget:
        type: paperlessngx
        url: https://paperless.somedomain.com
        username: user
        password: password
```

Preview:
Unbenannt2

Currently the total number of documents and the number of documents in the Inbox can be displayed.
In order to get the number of documents in the inbox, the `inboxTag` must be specified in the config.

---

Now these are two metrics that i am particularly interested in. I am happy to hear your opinion/ideas (different metrics? more dynamic approach with custom queries? ...)

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -349,5 +349,9 @@
         "passed": "Passed",
         "failed": "Failed",
         "unknown": "Unknown"
+    },
+    "paperlessngx": {
+        "inbox": "Inbox",
+        "total": "Total"
     }
 }
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -118,6 +118,7 @@ export function cleanServiceGroups(groups) {
           container,
           currency, // coinmarketcap widget
           symbols,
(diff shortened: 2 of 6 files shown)


shamoon:
Ah, someone after my own heart =)

One thing, I updated this to use the `/statistics` endpoint which provides exactly the same data `{"documents_total":3,"documents_inbox":3}` at a much less 'expensive' endpoint, of course this also makes the widget much simpler. Let me know if youre good with that!

Screen Shot 2022-11-25 at 8 21 49 AM

Tarow:
Thanks alot, i had no idea about the `/statistics` endpoint as i didn't find it in the [documentation](https://paperless-ngx.readthedocs.io/en/latest/api.html#).
Looks good, way cleaner now!

PR #581, opened 2022-11-29
Added processing count to overseerr

myztillx:
The `processing` count is a count of the still in process (Requested) media. It is equal to the `approved` minus the `available`.

 It is available in the same API call as the other information.


shamoon:
Cheers thanks

Issue #444, opened 2022-10-24
[Feature Request] CoinMarketCap Default Interval


renardyre:
### Description

A new setting to choose the default interval of the widget, instead of always render the page with the "1Hour" interval selected.
 ```
widget:
    type: coinmarketcap
    currency: GBP # Optional
    symbols: [BTC, LTC, ETH]
    ** default_interval : last_hour, last_day, last_week, last_month # Optional
    key: apikeyapikeyapikeyapikeyapikey

rursache:
would like this too

PR #597, opened 2022-12-01
CoinMarketCap Default Interval

rursache:
Set the default interval in widget config (`default_interval`) as requested in #444

### DIFF
--- a/src/widgets/coinmarketcap/component.jsx
+++ b/src/widgets/coinmarketcap/component.jsx
@@ -16,8 +16,10 @@ export default function Component({ service }) {
     { label: t("coinmarketcap.7days"), value: "7d" },
     { label: t("coinmarketcap.30days"), value: "30d" },
   ];
+  
+  const { default_interval } = default_interval;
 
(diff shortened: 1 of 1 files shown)


src/widgets/coinmarketcap/component.jsx
```diff
@@ -16,8 +16,10 @@ export default function Component({ service }) {
     { label: t("coinmarketcap.7days"), value: "7d" },
     { label: t("coinmarketcap.30days"), value: "30d" },
   ];
+  
+  const { default_interval } = default_interval;
```
shamoon, line 20: This doesnt work, youre instantiating a const with itself, also its before the widget object has been instantiated, I think should be moved down and just be `widget.default_interval`

Also please address the failing lint test re camelcase

shamoon (changes requested):
`ReferenceError: Cannot access 'default_interval' before initialization 
`
Respectfully, did you test this? Maybe it got messed up before your commit but see my comment. Thanks

shamoon:
Sorry mate, seemed simpler to make a new PR, thanks for the effort

PR #598, opened 2022-12-01
Add flood widget

JigSawFr:
Add flood widget to homepage (https://github.com/jesec/flood)

image

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -70,6 +70,12 @@
         "bitrate": "Bitrate",
         "no_active": "No Active Streams"
     },
+    "flood": {
+        "download": "Download",
+        "upload": "Upload",
+        "leech": "Leech",
+        "seed": "Seed"
+    },
     "changedetectionio": {
         "totalObserved": "Total Observed",
         "diffsDetected": "Diffs Detected"
--- a/src/widgets/components.js
+++ b/src/widgets/components.js
@@ -11,6 +11,7 @@ const components = {
   diskstation: dynamic(() => import("./diskstation/component")),
   docker: dynamic(() => import("./docker/component")),
   emby: dynamic(() => import("./emby/component")),
+  flood: dynamic(() => import("./flood/component")),
   gluetun: dynamic(() => import("./gluetun/component")),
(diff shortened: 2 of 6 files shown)


shamoon:
Question: doesnt using flood require the use of [deluge](https://gethomepage.dev/en/services/deluge/) / [qBittorrent](https://gethomepage.dev/en/services/qbittorrent/) / [rTorrent](https://gethomepage.dev/en/services/rutorrent/) / [Transmission](https://gethomepage.dev/en/services/transmission/), _every one of which we already support_? So I guess the question is why not just use those directly?

shamoon:
Like kind seems redundant?
Screen Shot 2022-12-01 at 8 56 37 AM

Also: not working for me `error: <floodProxyHandler> HTTP 422 logging in to flood.  Data: {"message":"Validation error."}`

JigSawFr:
> Question: doesnt using flood require the use of [deluge](https://gethomepage.dev/en/services/deluge/) / [qBittorrent](https://gethomepage.dev/en/services/qbittorrent/) / [rTorrent](https://gethomepage.dev/en/services/rutorrent/) / [Transmission](https://gethomepage.dev/en/services/transmission/), _every one of which we already support_? So I guess the question is why not just use those directly?

Not the same thing. Flood is an UI on top of qbittorent / deluge / rtorrent and Transmission.
It's not redundant as if you don't expose these torrents downloaders and only the UI, you can't use them. (like rflood hotio docker container)
Also Flood-UI is using a JSON API, same for rtorrent from jsec. So XLMRPC can't be used.
Sonarr/Radarr/Overseer also implemented communication with Flood.

JigSawFr:
@shamoon is your test protected by an auth? or auth was disabled in Flood ?

JigSawFr:
@shamoon if you can retry, auth is now fixed. 💯

image

shamoon (approved):
Fair enough, its still a little inception-y =)

And cool that fixed it

Screen Shot 2022-12-01 at 4 44 27 PM

shamoon:
Lint errors were fixed in https://github.com/benphelps/homepage/commit/01b55a17f1aafdd8c4312dfb321b6b9caee55259

PR #620, opened 2022-12-04
Respect scrutiny device status threshold setting

Tarow:
This PR addresses the issue mentioned [here](https://github.com/benphelps/homepage/issues/596).

The scrutiny widget now respects the device status threshold setting when calculating the number of failed and passed devices.

### DIFF
--- a/src/widgets/scrutiny/component.jsx
+++ b/src/widgets/scrutiny/component.jsx
@@ -9,7 +9,10 @@ export default function Component({ service }) {
     passed: 0,
     failed_smart: 1,
     failed_scrutiny: 2,
-    failed_both: 3
+    failed_both: 3,
+
+    isFailed: function ( s ) { return s > this.passed && s <= this.failed_both},
+    isUnknown: function ( s ) { return s < this.passed || s > this.failed_both}
   }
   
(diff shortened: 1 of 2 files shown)


shamoon:
Thanks for the PR! This wasnt working as-submitted though, it was missing the "passing" condition where the status was > threshold, e.g. in my case 2 drives failing scrutiny but passing SMART, so status returns `2` and threshold is `1` yet it was still showing passed: 0.

I also simplified the code, I dont think we need the Map.

Let me know if youre good with these changes! Feel free, of course.

Tarow:
Good catch, thanks for the changes!

You are right when it comes to the map. It's not really necessary but i thought semantically it makes more sense to map the `DeviceStatusThreshold` back to a `DeviceStatus`, as now what we are doing is `[DeviceStatusThreshold,DeviceStatus].includes(DeviceStatus)`, which works ofc, as the numbers match in this case.

Right now, the failed condition can collide with the unknown condition as device status `4` (assuming that could happen) and threshold `both` would be counted both as `failed` and `unknown`. 

I made some minor adjustments, let me know what you think.

PR #631, opened 2022-12-06
Miniflux widget

henry40408:


shamoon (changes requested):
If the reason for the proxy is just because the app uses `X-Auth-Token` you can use the credentialed proxy (with an additional condition) instead of writing your own custom one, please see https://github.com/benphelps/homepage/blob/main/src/utils/proxy/handlers/credentialed.js

shamoon:
Went ahead and took care of it =)

```yaml
      type: miniflux
      url: http://localhost:8080
      key: myapikey
```

Screen Shot 2022-12-16 at 10 34 08 PM

PR #636, opened 2022-12-07
Omada Widget

Oupsman:
Pull request for Omada Widget. 

Configuration of the widget is as such : 

        widget:
          type: omada
          url: https://locahost:8043
          username: homepage
          password: password
          site: Maison
          legacy: true for controller v3, false for controller v4 and +


shamoon:
You got it, nice!

In order to test this, since I dont have the hardware, can you copy and paste some output from the API? Ideally we could get that for older vs newer versions.

Thanks

Oupsman:
[responses.txt](https://github.com/benphelps/homepage/files/10178206/responses.txt)

shamoon (changes requested):
Thanks again for the PR.

Ok so there are some small issues with the code but before I get into those there are some big ones before this can be considered for commit:

1. The new API doesnt work? If you cant test it (me neither) we need to get either someone to test or example API output at the very least. It doesnt make sense to merge without that...
2. Please dont just duplicate the entire logic in `legacy` vs not, they need to be combined or separated into two widgets if theyre truly totally different.

Oupsman:
1/ In the responses.txt file, you'll see samples for both modern and legacy controllers : I've setup a modern controller on my laptop for me to test. 

2/ I'll change my code, I agree with you, it's not ideal.

Oupsman:
Code is unstable.

PR #644, opened 2022-12-08
use unprivileged user in a container

ajgon:
A good practice is to avoid running processes in containers as root, unless absolutely necessary. Since homepage only uses it's own `/app` directory, and exposes port `3000`, I think it is safe to drop privileges on start :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,10 @@ WORKDIR /app
 
 COPY --link package.json pnpm-lock.yaml* ./
 
-RUN <<EOF
-    set -xe
-    apk add libc6-compat
-    apk add --virtual .gyp python3 make g++
-    npm install -g pnpm
-EOF
+SHELL ["/bin/ash", "-xeo", "pipefail", "-c"]
+RUN apk add --no-cache libc6-compat \
+ && apk add --no-cache --virtual .gyp python3 make g++ \
+ && npm install -g pnpm
 
 RUN --mount=type=cache,id=pnpm-store,target=/root/.local/share/pnpm/store pnpm fetch | grep -v "cross-device link not permitted\|Falling back to copying packages from store"
 
@@ -29,12 +27,10 @@ ARG REVISION
 COPY --link --from=deps /app/node_modules ./node_modules/
 COPY . .
 
-RUN <<EOF
-    set -xe
-    npm run telemetry
-    mkdir config && echo '-' > config/settings.yaml
-    NEXT_PUBLIC_BUILDTIME=$BUILDTIME NEXT_PUBLIC_VERSION=$VERSION NEXT_PUBLIC_REVISION=$REVISION npm run build
-EOF
+SHELL ["/bin/ash", "-xeo", "pipefail", "-c"]
+RUN npm run telemetry \
+ && mkdir config && echo '---' > config/settings.yaml \
+ && NEXT_PUBLIC_BUILDTIME=$BUILDTIME NEXT_PUBLIC_VERSION=$VERSION NEXT_PUBLIC_REVISION=$REVISION npm run build
 
 # Production image, copy all the files and run next
 FROM docker.io/node:18-alpine AS runner
@@ -50,12 +46,15 @@ ENV NODE_ENV production
 WORKDIR /app
 
 # Copy files from context (this allows the files to copy before the builder stage is done).
-COPY --link package.json next.config.js ./
-COPY --link /public ./public
+COPY --link --chown=1000:1000 package.json next.config.js ./
+COPY --link --chown=1000:1000 /public ./public/
 
 # Copy files from builder
-COPY --link --from=builder /app/.next/standalone ./
-COPY --link --from=builder /app/.next/static/ ./.next/static/
+COPY --link --from=builder --chown=1000:1000 /app/.next/standalone ./
+COPY --link --from=builder --chown=1000:1000 /app/.next/static/ ./.next/static/
+COPY --link --chmod=755 docker-entrypoint.sh /usr/local/bin/
+
+RUN apk add --no-cache su-exec
 
 ENV PORT 3000
 EXPOSE $PORT
@@ -63,4 +62,5 @@ EXPOSE $PORT
 HEALTHCHECK --interval=10s --timeout=3s --start-period=20s \
   CMD wget --no-verbose --tries=1 --spider --no-check-certificate http://localhost:$PORT/api/healthcheck || exit 1
 
+ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["node", "server.js"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,8 +2,22 @@
 
 set -e
 
+# Default to root, so old installations won't break
+export PUID=${PUID:-0}
+export PGID=${PGID:-0}
+
 # This is in attempt to preserve the original behavior of the Dockerfile,
 # while also supporting the lscr.io /config directory
 [ ! -d "/app/config" ] && ln -s /config /app/config
 
-node server.js
+# Set privileges for /app but only if pid 1 user is root and we are dropping privileges.
+# If container is run as an unprivileged user, it means owner already handled ownership setup on their own.
+# Running chown in that case (as non-root) will cause error
+[ "$(id -u)" == "0" ] && [ "${PUID}" != "0" ] && chown -R ${PUID}:${PGID} /app
+
+# Drop privileges (when asked to) if root, otherwise run as current user
+if [ "$(id -u)" == "0" ] && [ "${PUID}" != "0" ]; then
+  su-exec ${PUID}:${PGID} "$@"
+else
+  exec "$@"
+fi


shamoon:
Thanks. Im no docker guru so not sure about the pros / cons. One initial question I have is whether this will have ramifications for e.g. bind mounts to e.g. `/public/images` as is suggested in the docs?

(obviously I can test that out myself too =)

ajgon:
Hey, that's the faster PR response I have ever received in my life :D 

It won't be a problem, as long as mounted directory keeps the same permissions.

But now when I think of it.... it may be cumbersome for some users, to keep an eye of it... maybe it's a better idea, to use entrypoint which will drop privileges to provided `PUID`/`PGID` env variables (similar to linuxserver.io). We can set them to 1000:1000 (node/node) by default.

The advancement will be, that container will run as root, chown all necessary directories to given PUID/PGID (so it will take the burded off the user), and then run homepage with lower privileges.

What do you think?

shamoon:
Cool thanks. Yea that makes sense to me but again Im not all that knowledgeable re. docker so will let others weigh in before merging

JigSawFr:
Also agree on this, better to have this as popular containers provided by linuxservers or hotio. Thanks for the Pr

Macleykun:
> Thanks. Im no docker guru so not sure about the pros / cons. One initial question I have is whether this will have ramifications for e.g. bind mounts to e.g. `/public/images` as is suggested in the docs?
> 
> (obviously I can test that out myself too =)

I don't think this will be an issue, you only need read rights to the pictures and folders, and by default everyone has read rights and execute for folders.

While were at it, i noticed EOF is used, why not use && and \ to do multiple command's in one run?

And if you want to futher optimize/follow best practises, i would suggest checking https://hadolint.github.io/hadolint/

ajgon:
I updated Dockerfile to be more-or-less compliant with hadolint.

Macleykun:
> I updated Dockerfile to be more-or-less compliant with hadolint.

looks great! Thank you for looking into it :) I think this PR is ready for merging

Macleykun:
Something else that i only just noticed, but shouldn't we add USER somewhere near the bottom of the Dockerfile to run the container rootless?
I believe we already have a user (pid 1000) we can use for this.
However, this probably will break some intances then, as users can have their config's mounted where the permission isn't set correctly.
They would need to do: sudo chown -R 1000:1000 Homepage/ (aka the mount on the host).

Thoughs?

ajgon:
Yes, originally this PR had what you mentioned (USER 1000), but I decided against it, exactly because of the reasons you mentioned. Currently it runs as root, but only to set privileges, and drop them to non-privileged user. Thanks to that, container can handle incorrectly mounted directories (with wrong permissions) and still work as unprivileged.

Advanced users, can set permissions by themselves, and then run the container (either via `docker run -u node` or by adding `user: node` stanza to docker-compose) directly as an unprivileged user - this would also work, as there are checks in place.

Macleykun:
> Yes, originally this PR had what you mentioned (USER 1000), but I decided against it, exactly because of the reasons you also mentioned :). Currently it runs as root, but only to set privileges, and drop them to non-privileged user. Thanks to that, container can handle incorrectly mounted directories (with wrong permissions) and still work as unprivileged.
> 
> Advanced users, can set permissions by themselves, and then run the container (either via `docker run -u node` or by adding `user: node` stanza to docker-compose) directly as an unprivileged user - this would also work, as there are checks in place.

i only realised that's also an option now, very clever! I think this PR is ready to be merged if @shamoon is ok with that too

shamoon (approved):
Thanks. It looks fine to me but I'm not going to merge without other maintainers/Ben weighing in

JazzFisch:
This change seems to break permissions for existing installs.  I just ran this and without specifying `user: ####:####` or `PUID: $PUID` and PGID: $PGID` and it clobbered over the permissions I had on my already existing mounted folder.

benphelps:
How does this affect making connections to docker sockets?  The socket almost surely requires root access, or the user to be part of the docker group.

When I attempted this last time, socket connections (even as a user with docker access) broke.

JazzFisch:
> How does this affect making connections to docker sockets? The socket almost surely requires root access, or the user to be part of the docker group.
> 
> When I attempted this last time, socket connections (even as a user with docker access) broke.

Great question.  I had this working, but I have moved to using ghcr.io/tecnativa/docker-socket-proxy.

ajgon:
Using sockets in containers is considered as anti-pattern - docker can expose its API via SSL endpoint, which can be used to perform same actions, with proper authentication etc.........

..... having said that, we live in a real world, and sockets are fast way which "just works", and most people use them, so I agree, they should be supported. I can think of few solutions to this problem:

- add support for `PUID=0`/`PGID=0`. In that case, no privileges will be dropped, no permissions changed, and everything will work
- it is also possible to set `PUID`/`PGID` to the uid/gid of docker user in the host system. Then, the container would have access to docker socket. The downside is, that all other files will be chowned to docker user as well.
-  just drop support for sockets, and use proper SSL way (not gonna happen :D )

I think the best approach is `PUID=0`/`PGID=0` - if user wants to use docker.sock, they know what they sign for - and that privileges cannot be dropped... We can even set PUID/PGID to 0 as default, and add big warning in documentation "if you gonna change this, and using docker.sock, you're gonna have a bad time..."

benphelps:
Yeah, a large portion of the homepage users are already using sockets, and breaking most of the installs would be awful.  

Defaulting to `PUID=0`/`PGID=0` would essintally preserve the current functionality, right?  If so, then I think this could go out at any time.

Like @JazzFisch mentioned, using `tecnativa/docker-socket-proxy` is a proper way to handle this, and we can direct users to that from the very start, leaving sockets to those users who explicitly want to use them.

ajgon:
All rightie, I updated the script to default to (and support) `0:0` (effectively being backwards compatible). In this configuration, socket is available, and permissions of mounted files are not changed. To do something differently, user explicitly has to change PUID/PGID.

JazzFisch:
Your most recent change results in the following message in the logs:
```
sh: missing ]

Listening on port 3000
```

ajgon:
Sorry, my bad - fixed.

JazzFisch:
@ajgon, that fixed the error I saw in the logs, however the PGID values appears to be ignored during the chown step.  All of the file ownership changes from (for example) `user:group` to `user:user`.  Can you fix the chown problem?

ajgon:
`export PGID=${PGID:-0}` :man_facepalming: 

Stupid mistake, I juggled with this too many times :) Fixed.

JazzFisch:
That did the trick.  I tried this with and without PUID/PGID and both succeeded without altering my existing file permissions.

JazzFisch:
@benphelps and @shamoon we'll need to update the docs site as well to reflect this change.

shamoon:
Yep! can go in https://github.com/benphelps/homepage-docs/pull/20

shamoon:
Hey @ajgon , unfortunately as-is this seems to have broken the docker integration via docker sock, which as noted, if not ideal is currently in use by a lot of users. I've reverted the PR for now but would love to try again and troubleshoot. Our bad for not testing this, of course. Again, if youre interested in opening a new one please do.

ajgon:
Hey, my bad as well... should've test it more thoroughly. I'll take a look on a weekend, and see what's wrong. Can you share, what issues occurred (permission issues, problems with connections?).

shamoon:
No worries! The stakes are low here haha

I haven’t taken the time to look more deeply into why exactly, to be honest I just wanted to fix it before everyone’s watchtower updated and we got bombarded 😝 But it seemed to just be docker connections via socket that borked. (The EOF stuff is fine I’m sure btw, again I was aiming for speed)

As mentioned, for better or for worse, a lot of our users are using docker integration via socket (especially since that was the recommended way to setup the app for a long time). Also, dunno if this came up, but there is a huge variation in the technical ability of our users and many seem pretty novice (no judgements!) which is to say I think we want to make this as easy as possible. If that just means releasing it with clear notes / documentation that’s cool with me, I just don’t want it taking folks completely by surprise!

Again we have a moment to investigate, I will try to find some time for that too but for sure we’d like these changes in a future release, appreciate your effort.

Macleykun:
Just my 2 cents, is it b/c we don't use root but smth else (uid/gid is different) which can't read and write to the /var/run/docker.socket?
```bash
srw-rw----    1 root     995              0 Jan 14 14:24 docker.sock
```
The only way i can imagine to fix this, is by creating a docker group in the container with the same ID: 995
Assign that group to the rootless user, and it should be able to read/write again.

ajgon:
I did some tests today, and I can't reproduce an issue, where `docker.sock` is inaccessible with default settings (i.e., not tinkering with PUID/PGID params):

```bash
# notice explict use of v0.6.0, which introduce unprivileged users
$ docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock ghcr.io/benphelps/homepage:v0.6.0 sh

# inside docker
/app # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)

/app # apk add curl
# ...

/app # curl --silent -XGET --unix-socket /var/run/docker.sock http://localhost/version
{"Platform":{"Name":""},"Components":[{"Name":"Engine","Version":"20.10.23","Details":  ... and so on
```

Not sure about EOF related stuff, but it shouldn't matter here - again, on default settings (no custom `PUID`/`PGID`) it runs a root as it did before, so I see no reason why it should fail :thinking:

shamoon:
Thanks, I havent had the chance to dig into this (and have to run RN but wanted to post quickly) but the curl check works fine for me too on 0.6.0, just not from the app itself. Perhaps something about permissions on the nodejs process? Im definitely not an expert here, again happy to keep investigating when I can or if there are specific things to try.

Screen Shot 2023-01-22 at 9 15 51 AM

ajgon:
Hmm... but API error shouldn't be related to `docker.sock` no? It's just problem with communication between containers (invalid network configuration?), which should be unrelated...

shamoon:
Sorry if thats not clear, its https://github.com/benphelps/homepage/blob/cdbdb86c9e67751c73f8e33459bee7b79386d808/src/pages/api/docker/stats/%5B...service%5D.js#L75-L77 which yes will look like that. Currently that failure has no helpful output (error isnt caught) but I think its just EACCES, again need to dig into it.
Screen Shot 2023-01-22 at 9 23 02 AM

shamoon:
Oh wait, damn my bad this might be on me

shamoon:
My sincere apologies all, I was too hasty in diagnosing this. You are (obviously) correct ajgon about defaulting to root, I will re-merge. Thanks for your patience

ajgon:
No worries :D So what's the issue?

shamoon:
I was looking at the wrong config (too many testing installs / versions hanging around) and I was truly specifying UID/GID which was (correctly) not using root which of course will break the socket

Nonoss117:
I'm using Unraid, and homepage docker container
Homepage version : [v0.6.3 (c74b467, 22 janv. 2023)](https://github.com/benphelps/homepage/releases/tag/v0.6.3)

with user : nobody
PUID : 99
PGID : 100
API error and status unknown
image

image

with user root :
PUID : 0
PGID : 0
API ok and status ok
image

shamoon:
If you are using docker socket the app will need root permissions for docker stats/status, if you want to change that have a look at other means of docker integration (don’t have a link on hand but in the docs of course)

ajgon:
Here is the link for future reference: https://gethomepage.dev/en/configs/docker/

Nonoss117:
thank's @shamoon for your reply & @ajgon for the link. 

Yes I'm using docker socket :
```yml
/var/run/docker.sock:/var/run/docker.sock:ro # Mounted as read-only
```
image

I think I'll keep my homepage configuration

PR #674, opened 2022-12-13
Support for Omada widget

Oupsman:
Adding support for Omada Widget, with compatibility with v3, v4 and v5 controllers. 

Will display 3 fields : number of devices connected to Wifi, number of AP connected to the controller and number of alerts. 

On v4 and v5 controllers, if the numbers of connected switches and gateways are greater than 0, they are displayed as well. 

Configuration is as such :

        widget:
          type: omada
          url: url of the controller
          username: username
          password: password
          site: the site you want the stat from.


shamoon:
Thanks, again when someone has time we will review.

also, same as last time please remove the extraneous files and less important would be to cleanup the commit history here (squash and force-push)

Oupsman:
It seems I can't squash any commit, I'm working on it. 

SOrry, I'm new to collaborating with github.

Issue #404, opened 2022-10-17
[Feature Request] Tdarr widget


jbeck22:
would love to have a tdarr widget

irlam:
+1

mokshmridul:
+1

wazzirr:
+1

rshide:
Submitted [PR687](https://github.com/benphelps/homepage/pull/687) for this.  Will update the docs once this is approved and merged.

PR #687, opened 2022-12-16
added tdarr widget

rshide:
Adding a widget for Tdarr to close [#404](https://github.com/benphelps/homepage/issues/404).

This will display 4 fields: 
- Count of items currently in queue
- Count of items that have been processed
- Count of items that have errored
- Total amount of disk space saved

Configuration example:
```        
widget:
    type: tdarr
    url: http://tdarr.host.or.ip
```

Preview:
image

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -192,6 +192,12 @@
         "stopped": "Stopped",
         "total": "Total"
     },
+    "tdarr": {
+        "queue": "Queue",
+        "processed": "Processed",
+        "errored": "Errored",
+        "saved": "Saved"
+    },     
     "traefik": {
         "routers": "Routers",
         "services": "Services",
--- a/src/widgets/components.js
+++ b/src/widgets/components.js
@@ -43,6 +43,7 @@ const components = {
   speedtest: dynamic(() => import("./speedtest/component")),
   strelaysrv: dynamic(() => import("./strelaysrv/component")),
   tautulli: dynamic(() => import("./tautulli/component")),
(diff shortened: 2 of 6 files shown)


shamoon:
Thanks, we'll take a look. There are a couple comments in the code if you could remove but one actual thing is re. `headerCacheKey`, I dont see the purpose of this, its just setting the `content-type`? That shouldn't be necessary. And an odd API that the call is a `POST`

wazzirr:
Would be great to also see what's currently being transcoded, and time left

rshide:
> Thanks, we'll take a look. There are a couple comments in the code if you could remove but one actual thing is re. `headerCacheKey`, I dont see the purpose of this, its just setting the `content-type`? That shouldn't be necessary. And an odd API that the call is a `POST`

@shamoon I've removed the comments and the headerCacheKey as well.  Also added a preview and config example. I agree that this doesn't make much sense as a `POST`, but that is the only call that [their docs](https://tdarr.readme.io/reference/post_cruddb)  provide. Let me know if anything else needs to be cleaned up.


> Would be great to also see what's currently being transcoded, and time left

@wazzirr I think it makes sense to also allow time remaining to be displayed.  I can look into this more in the future and see if there is an API available for that.  

For what is being currently transcoded, are you thinking a list of files that are currently being transcoded similar to what is displayed under the Nodes section in Tdarr?  If so, are there any other widgets that currently display a list of items?

wazzirr:
Nvm, I forgot about an option to transcode more than one file at once.

shamoon:
Cheers thanks

PR #688, opened 2022-12-16
Adds widget for NextDNS

luiseduardobrito:
Adds a simple widget to get the main analytics from the [NextDNS.io API](https://nextdns.github.io/api/), similar to the PiHole widget.

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -372,5 +372,9 @@
     "paperlessngx": {
         "inbox": "Inbox",
         "total": "Total"
+    },
+    "nextdns": {
+        "wait": "Please Wait",
+        "no_devices": "No Device Data Received"
     }
 }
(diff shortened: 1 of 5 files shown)


shamoon:
Thanks, never heard of this but cool service. Widget works great. Always helpful to have an example (i.e. to show what fields are needed)

```
     - NextDNS:
          icon: nextdns.png
          widget:
               type: nextdns
               profile: profileid
               key: yourapikeyhere
```

One quibble, if no devices are setup the widget shows nothing which might be confusing (make people think its not working) so I changed it to show "No Device Data Received". Also adjusted the case when data hasn't been received yet (that's presumably a waiting state and shouldn't break the visual layout of a container with just "no data").

Let me know your thoughts, feel free to adjust.

Screen Shot 2022-12-16 at 2 24 08 PM
Screen Shot 2022-12-16 at 2 08 37 PM

luiseduardobrito:
Thanks for the review and feedbacks, you're absolutely right I should have put an example :)

PR #721, opened 2022-12-23
fixed number formatting in widgets: radarr, sonarr, overseerr

kimpenhaus:
I've fixed some of the widgets to give them some proper number formatting while using the i18n-translator

widgets:

- sonarr
- radarr
- overseerr

sample when having culture information setup to: de-DE

before:
image

after:
image

### DIFF
--- a/src/widgets/overseerr/component.jsx
+++ b/src/widgets/overseerr/component.jsx
@@ -1,8 +1,11 @@
+import { useTranslation } from "next-i18next";
+
 import Container from "components/services/widget/container";
 import Block from "components/services/widget/block";
 import useWidgetAPI from "utils/proxy/use-widget-api";
 
 export default function Component({ service }) {
+  const { t } = useTranslation();
   const { widget } = service;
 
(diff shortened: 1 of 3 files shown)


Nonoss117:
Good idea, I agree with you, numbers formatting for radarr, sonarr, overseerr widgets can be improved.
But I think number formatting is different depending on the country.
In France we use space for this formatting.

This is an example of TubeArchivist widget and its fine for me.
image

kimpenhaus:
It uses the provided i18n translator - so I think it should respect the given culture information.

shamoon:
Nice thanks. Yes as this uses the translation function (like in other widgets) it will respect localization

shamoon (approved):

Nonoss117:
Arrr to late, I wanted to ask the same for prowlarr widget :
image

shamoon:
> Arrr too late, I wanted to ask the same for prowlarr widget :

No worries, added in 1cac02f

PR #724, opened 2022-12-24
Torrent listing in qBittorrent

dbtdsilva:
I've added torrent listing in the qBittorrent widget. Let me know your thoughts.
I based myself a little bit from Tautulli.

This is the end result (the torrents names were auto-generated, but I guess you can image how it looks like):

qbittorrent_torrent_list

### DIFF
--- a/src/widgets/qbittorrent/component.jsx
+++ b/src/widgets/qbittorrent/component.jsx
@@ -137,7 +137,7 @@ export default function Component({ service }) {
   }
 
   const leech = torrentData.length - completed;
-
+  const torrentsEnabled = service.widget.fields === null || service.widget.fields?.includes('torrents');
   return (
     <>
       <Container service={service}>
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks for the PR. This is cool as a proof-of-concept but I’m not sure if I’d be in favor of merging this, it doesn’t really feel in line with other widgets, although Tautulli perhaps. It also feels a little unwieldy?

I think of the homepage widgets as aiming to be quick useful glances of info, not a comprehensive alternative to the actual service.

Of course, happy to hear others’ perspective

Oupsman:
Maybe just display the top 3 active torrents (aka downloading or uploading) sorted out by speed ?

dbtdsilva:
> Maybe just display the top 3 active torrents (aka downloading or uploading) sorted out by speed ?

We could even add those kinds of display options into the widget configuration

shadofall:
I like it, i'd also would love to see something similar with SAB, if its something that can be made optional for if users want to see it or not, Combine it with a nice calendar widget with "today, this week, this month in sonarr/radarr" and you'd have to only two features from organizrr i'm missing ;)  but just my two bits as a new user :P

FullMetalEd:
Recently switched to Homepage, and I think this would be a good feature too. My suggestion is that you make the list start collapsed and click to reveal. loading the full list at page load is unwieldy.

JazzFisch:
Thank you for the feedback everyone.  We are weighing the usability / visibility aspects of this PR against issues from other users who want more consistency in the sizing of widgets so that they line up and align in a way that is aesthetically pleasing for them.  A possible solution we've considered would be for the list to be off by default with an opt in mechanism.  We've done this in other areas so that older behavior is the default and new behavior is behind an opt-in configuration.

shamoon (changes requested):
Yes if you can put behind a flag

shamoon:
Please re-open if / when you're ready to proceed with this, just closing to tidy things up. Thanks

Issue #643, opened 2022-12-08
[Feature Request] Add Support for Xteve in a widget


iceman73:
### Description

I would love to get the status of Xteve and how many streams its playing

xteve converts IPTV into a mock hdhomerun to play inside of plex.

It has an API
https://github.com/xteve-project/xTeVe-Documentation/blob/master/en/configuration.md#api

And its Status API returns, which is very useful in knowing its status:
  "status": true,
  "streams.active": 2,
  "streams.all": 21,
  "streams.xepg": 2,

HamzaM:
This would be really good !

iceman73:
Thankyou works great

HamzaM:
> Author

Just wondering, what repo of Xteve do you use ? I found these 2 that are still in development : 
https://github.com/dnsforge-repo/xteve
https://github.com/SenexCrenshaw/xTeVe

shamoon:
Those are just docker wrappers for the actual project. Hopefully it doesn’t matter? I tested his code with dnsforge one

PR #731, opened 2022-12-26
added widget for xteve

kimpenhaus:
This PR is adding a widget `xteve` that can be either configured to use `username/password` credentials or `none` as provided by the docs of xteve.

image

It is addressing #643 

As I am not a JS developer it is very much appreciated if you could give that PR a deep check :) Things I wasn't aware how to do is:

 - translation (I've added English defaults)
 - documentation

If you could give me a link or some information I will happily provide the missing parts.


shamoon:
Cool we’ll take a look. The translations happen via another tool so this is perfect, you just supply the English one. Docs are in a separate repo but not necessary to update, I can handle that (obviously, feel free it’s homepage-docs)

shamoon:
Awesome, really minor changes like just syntax stuff

PR #770, opened 2023-01-04
Giving Read Only rights to homepage

HaroldVB:
Giving Read Only rights to homepage container in docker-compose example.
Adding :RO to the docker.sock volume. When the container gets compromised the intruder will have root access basically. The container doesn't need the write privileges.

This measure will stop inexperienced people from exposing their docker.socket to the public internet.


shamoon:
Sure thanks

PR #771, opened 2023-01-04
Feature: allow setting locale for datetime widget directly

shamoon:
E.g.:
Screen Shot 2023-01-04 at 1 50 31 PM

Screen Shot 2023-01-04 at 1 51 45 PM

Screen Shot 2023-01-04 at 1 51 26 PM

Fixes #751


Mega-Volti:
It would be great to also have support for the ISO 8601 date format. I think en_DK might be a decent workaround but directly supporting "isodate" or similar would be more clear.

PR #777, opened 2023-01-07
Change TrueNAS to use API key

jhollowe:
Using an API key is better than username/password and it is easy to generate an API key in TrueNAS.

Documentation updates in https://github.com/benphelps/homepage-docs/pull/18

### DIFF
--- a/src/utils/proxy/handlers/credentialed.js
+++ b/src/utils/proxy/handlers/credentialed.js
@@ -30,6 +30,8 @@ export default async function credentialedProxyHandler(req, res, map) {
         headers["X-gotify-Key"] = `${widget.key}`;
       } else if (widget.type === "authentik") {
         headers.Authorization = `Bearer ${widget.key}`;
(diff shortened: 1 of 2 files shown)


shamoon (changes requested):
Hmm, but this is a breaking change for literally all users with this widget setup. I don’t use Truenas myself but seems hard to justify unless we know that most folks agree. 

So yea, without a clear mandate to change I think this would need to support both, which can be done

shamoon:
I have fixed this but I cant push to your fork. Can you either grant me access or I can recreate the PR, up to you.

jhollowe:
You should be able to push to it now.

shamoon:
Sweet thanks, take a look and let me know, this should support both

Screen Shot 2023-01-07 at 10 10 00 AM

jhollowe (commented):
Looks good!

shamoon:
Awesome thanks

PR #793, opened 2023-01-09
Added support for custom themes

libkothimira:
I renamed the `theme-white` to `theme-custom` and used the original `theme-white` colors as fallback values. 

To set a custom theme, create `public/custom-colors.css` and put something like this in it (adjusting colors as needed)

```css
* {
    --custom-color-50: 220 224 232;
    --custom-color-100: 230 233 239;
    --custom-color-200: 205 214 244;
    --custom-color-300: 124 127 147;
    --custom-color-400: 140 143 161;
    --custom-color-500: 92 95 119;
    --custom-color-600: 88 91 112;
    --custom-color-700: 76 78 105;
    --custom-color-800: 30 30 46;
    --custom-color-900: 65 69 85;
    --custom-color-logo-start: 255 0 0;
    --custom-color-logo-stop: 0 0 255;
}
```

### DIFF
--- a/public/custom-colors.css
+++ b/public/custom-colors.css
@@ -1,0 +1,14 @@
+* {
+    --custom-color-50: ;
+    --custom-color-100: ;
+    --custom-color-200: ;
+    --custom-color-300: ;
+    --custom-color-400: ;
+    --custom-color-500: ;
+    --custom-color-600: ;
+    --custom-color-700: ;
+    --custom-color-800: ;
+    --custom-color-900: ;
+    --custom-color-logo-start: ;
+    --custom-color-logo-stop: ;
+}
--- a/src/components/toggles/color.jsx
+++ b/src/components/toggles/color.jsx
@@ -27,7 +27,7 @@ const colors = [
   "pink",
   "rose",
   "red",
-  "white",
+  "custom",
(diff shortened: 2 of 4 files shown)


shamoon (changes requested):
Thanks. I like the idea here but as-is without adding that file this PR will cause a 404 out-of-the-box which I don’t love. Then of course including it by default is a sort of pointless extra request, but perhaps the lesser of two not ideal choices. Open to other ideas ofc

also wondering about this PR opening the custom CSS can if works a bit. Hmm

libkothimira:
I added `public/custom-colors.css` to avoid the 404, but I also see an issue with linking the stylesheet in `_document.jsx` since that results in `Warning: Do not include stylesheets manually. See: https://nextjs.org/docs/messages/no-css-tags  @next/next/no-css-tags` being emitted. Not sure if there's a better way to handle including a file that we don't necessarily use at build time

shamoon:
Yes I should have identified that from the jump, as this is sort of antithetical to the whole static site thing...

shamoon:
@libkothimira let us know if you come up with another solution here (of course feel free to re-open), but again it kinda goes against the static site principle

PR #798, opened 2023-01-10
Add Moonraker (Klipper) widget

Eizock:
I've added support for displaying prinstatus from Klipper via Moonraker.
To use it:
`type: moonraker`
`url: http://ip.to.your.moonraker:port`

### DIFF
--- a/src/widgets/moonraker/component.jsx
+++ b/src/widgets/moonraker/component.jsx
@@ -9,16 +9,16 @@ export default function Component({ service }) {
 
   const { widget } = service;
 
-  const { data: mr_print_stats, error: mre_print_stats } = useWidgetAPI(widget, "print_stats");
-  const { data: mr_display_status, error: mre_display_status } = useWidgetAPI(widget, "display_status");
-  const { data: mr_webhooks, error: mre_webhooks } = useWidgetAPI(widget, "webhooks");
+  const { data: printStats, error: printStatsError } = useWidgetAPI(widget, "print_stats");
+  const { data: displayStatus, error: displayStatsError } = useWidgetAPI(widget, "display_status");
+  const { data: webHooks, error: webHooksError } = useWidgetAPI(widget, "webhooks");
 
-  if (mre_print_stats || mre_display_status || mre_webhooks) {
-    const finalError = mre_print_stats ?? mre_display_status ?? mre_webhooks;
+  if (printStatsError || displayStatsError || webHooksError) {
+    const finalError = printStatsError ?? displayStatsError ?? webHooksError;
     return <Container error={finalError} />;
   }
 
-  if (!mr_print_stats || !mr_display_status || !mr_webhooks) {
+  if (!printStats || !displayStatus || !webHooks) {
     return (
       <Container service={service}>
         <div class="inline-flex flex-col" style={{width: '100%'}}>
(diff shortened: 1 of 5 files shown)


shamoon (changes requested):
There are some linting things to fix but also this is another one it looks like we wont be able to test easily. Would you mind posting some sample output from the API at least?

Eizock:
Apologies, for the avoidable linting errors. I'm quite new to linting in general.
The linting errors should now be fixed.

Here are sample outputs of the API endpoints.
<details>
  <summary>print_stats</summary>

  ```json
  {
    "result": {
      "status": {
        "print_stats": {
          "info": {
            "total_layer": null,
            "current_layer": null
          },
          "print_duration": 4885.865378969,
          "total_duration": 5133.010960750005,
          "filament_used": 9796.927220000001,
          "filename": "SC846_SSD_HDD_Holder.gcode",
          "state": "complete",
          "message": ""
        }
      },
      "eventtime": 248270.603765604
    }
  }
  ```
</details>

<details>
  <summary>display_status</summary>

  ```json
  {
    "result": {
      "status": {
        "display_status": {
          "progress": 1,
          "message": null
        }
      },
      "eventtime": 248284.897741016
    }
  }
  ```
</details>

<details>
  <summary>webhooks</summary>

  ```json
  {
    "result": {
      "status": {
        "webhooks": {
          "state": "shutdown",
          "state_message": "Lost communication with MCU 'mcu'\nOnce the underlying issue is corrected, use the\n\"FIRMWARE_RESTART\" command to reset the firmware, reload the\nconfig, and restart the host software.\nPrinter is shutdown\n"
        }
      },
      "eventtime": 253881.55679112
    }
  }
  ```
</details>

JazzFisch:
@Eizock 5 fields can be pretty busy on a widget.  Can you also provide some images of it running?

Eizock:
image

I tried styling it so it doesn't look too busy.
Alternatively i could only display Printer Status if it's offline and display the rest instead if it's online.

shamoon:
Ah yea we shoot for 4 max as horizontal blocks, perhaps drop one

Eizock:
image
image

I ditched the filename entirely and made it, that if the printer is offline it only shows status (upper image).
And if it is online, it shows layers, progress and the status of the print (lower image).

JazzFisch:
As you are no longer using the filename field, can you please remove the entry from common.json?

Eizock:
Yeah, i noticed aswell. Sorry for that.

shamoon:
My only other question is how much interest there is for this (besides you =), I honestly dont know. Theres only one discussion about 3d print monitoring and sounds like they weren't using this tool. #250

Eizock:
The mentioned Mainsail and/or Fluidd is the UI, Moonraker provides the actual API to connect to Klipper and Klipper is the firmware. :D 
See [Mainsail-Docs](https://docs.mainsail.xyz/setup/manual-setup/moonraker) and [Fluidd-Docs](https://docs.fluidd.xyz/configuration/moonraker) and [Moonraker](https://github.com/Arksine/moonraker).

shamoon:
Thanks for clarifying that, you 3d printers weave a tangled web (pun?)

One thing in testing, can you send an example API output for printStats when things are going? are the `total_layer` and `current_layer` properties passed as strings or numbers? there may be an issue there

src/widgets/moonraker/component.jsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <Container service={service}>
+        <Block label="moonraker.printer_state" value={webHooks.result.status.webhooks.state} />
+      </Container>
+    );
+  }
+
+  let currentLayer = "-";
+  let totalLayer = "-";
+  if(!printStats.result.status.print_stats.info.total_layer === "") {
```
shamoon, line 39:
```suggestion
  if (printStats.result.status.print_stats.info.total_layer !== "") {
```

I think should be this, at the moment youre (unintentionally) coercing the value to a boolean which doesnt really work. Depending on the type of the API output though there could be another way

shamoon (changes requested):

Eizock:
We don't weave it, we print it. :P

Here is the requested sample output.
<details>
  <summary>print_stats</summary>

  ```json
  {
    "result": {
      "status": {
        "print_stats": {
        "info": {
            "total_layer": 361,
            "current_layer": 7
          },
          "print_duration": 0,
          "total_duration": 0,
          "filament_used": 0,
          "filename": "",
          "state": "standby",
          "message": ""
        }
      },
      "eventtime": 410808.673944955
    }
  }
  ```
</details>

src/widgets/moonraker/component.jsx
Eizock: Yeah, i think you are right. TBH i don't know what my brain was thinking while i wrote that.
```suggestion
  if(printStats.result.status.print_stats.info.total_layer !== null) {
```
This should be even better since I am using strict equality and the API returns null instead of 0 if it has no layer information.

shamoon:
Sweet thanks!

PR #834, opened 2023-01-17
OctoPrint Widget

williamwoldum:
Made myself this widget that displays info from my 3d printer running OctoPrint, and thought others may enjoy it as well.

The widget has 3 primary display modes based on the printer state:

norfound
operational
printing

Most of the time my printer isn't even on and the API requests will fail.
I don't want to see the red API error box showing `500` just because the printer is off.
So, I overwrote the generic proxy to ignore `500`, `502`, and `409` errors and instead display the printer state "Not found" (first image).
I made it so that this silencing can be enabled/disabled from the `services.yaml` file via the setting `silencePrinterNotFound`.

```
- OctoPrint:
        icon: https://octoprint.org/assets/img/logo.png
        href: http://example.com
        description: 3d printing
        widget:
          type: octoPrint
          url: http://example.com
          key: 758.....F968
          silencePrinterNotFound: true
```

### Refs
* https://octoprint.org/
* https://docs.octoprint.org/en/master/


shamoon (changes requested):
Thanks for the widget.

- Yes, the translation thing =)
- Widgets shouldn't have more than 4 blocks, please lop 2 off
- It looks like your proxy has no purpose other than the error 500 thing right? This is an odd one, but I get the thought process. My issue is a user will never be able to tell if they've configured it wrong, it will just always show shutdown. Not sure what the solution is here. Either way, instead of overwriting the proxy (which will complicate upkeep etc) you can just catch this in the component. But yea, still a little unsure about the "500 is expected" thing. I see how your `silencePrinterNotFound` setting helps but not sold on that either. Hmm

williamwoldum:
I have now removed the concern completely, if the printer isn't on the red API error box just shows.
Hence I also removed the custom proxy and the `silencePrinterNotFound` yaml option.

The 3 display modes now looks like this

notfound2
operational2
printing2

And the yaml file like this
```
- OctoPrint:
        icon: https://octoprint.org/assets/img/logo.png
        href: http://example.com
        description: 3d printing
        widget:
          type: octoPrint
          url: http://example.com
          key: 758.....F968
```

shamoon:
Cool thanks! One last request, since we can’t meaningfully test this without a 3d printer lying around 😉 would you be able to provide a couple of relevant sample outputs from the API for us to test with?

thanks again

williamwoldum:
Of course 

### Endpoint `printer` 
`printer.com/api/printer?apikey=752...68` [Docs](https://docs.octoprint.org/en/master/api/job.html#retrieve-information-about-the-current-job)
```
{
    "sd": {
        "ready": false
    },
    "state": {
        "error": "",
        "flags": {
            "cancelling": false,
            "closedOrError": false,
            "error": false,
            "finishing": false,
            "operational": true,
            "paused": false,
            "pausing": false,
            "printing": false,
            "ready": true,
            "resuming": false,
            "sdReady": false
        },
        "text": "Operational"
    },
    "temperature": {
        "bed": {
            "actual": 24.37,
            "offset": 0,
            "target": 0.0
        },
        "tool0": {
            "actual": 24.61,
            "offset": 0,
            "target": 0.0
        }
    }
}
```

For a brief moment after the printer is turned on this endpoint will not return temperatures
```
{
    "sd": {
        "ready": false
    },
    "state": {
        "error": "",
        "flags": {
            "cancelling": false,
            "closedOrError": false,
            "error": false,
            "finishing": false,
            "operational": true,
            "paused": false,
            "pausing": false,
            "printing": false,
            "ready": true,
            "resuming": false,
            "sdReady": false
        },
        "text": "Operational"
    },
    "temperature": {}
}
```

---

### Endpoint `job` 
`printer.com/api/job?apikey=752...68` [Docs](https://docs.octoprint.org/en/master/api/printer.html#retrieve-the-current-printer-state)

```
{
    "job": {
        "averagePrintTime": 2152.3367760640003,
        "estimatedPrintTime": 1678.1762885939777,
        "filament": {
            "tool0": {
                "length": 829.7786100000014,
                "volume": 0.0
            }
        },
        "file": {
            "date": 1621853434,
            "display": "CE3_3d print.gcode",
            "name": "CE3_3d_print.gcode",
            "origin": "local",
            "path": "CE3_3d_print.gcode",
            "size": 2380119
        },
        "lastPrintTime": 2152.3367760640003,
        "user": "wilwol"
    },
    "progress": {
        "completion": 0.00886510296333923,
        "filepos": 211,
        "printTime": 0,
        "printTimeLeft": 2152,
        "printTimeLeftOrigin": "average"
    },
    "state": "Printing"
}
```

In case no job is active
```
{
    "job": {
        "averagePrintTime": null,
        "estimatedPrintTime": null,
        "filament": null,
        "file": {
            "date": null,
            "display": null,
            "name": null,
            "origin": null,
            "path": null,
            "size": null
        },
        "lastPrintTime": null,
        "user": null
    },
    "progress": {
        "completion": null,
        "filepos": null,
        "printTime": null,
        "printTimeLeft": null,
        "printTimeLeftOrigin": null
    },
    "state": "Operational"
}
```

JazzFisch:
For those of us with a, shall we say unique measuring system, could you find a way of indicating the temperature scale?  I mean, I'm pretty sure it isn't in Kelvin, but I can't tell if those number are C or F.

williamwoldum:
😆 They are indeed celcius. And the time is listed in seconds.
Let me know if I should add a metric/imperial boolean option in the yaml config

JazzFisch:
Maybe a units setting like the Open Meteo service widget?
https://gethomepage.dev/en/widgets/openmeteo

shamoon:
Or if you 3d printers are all sane and using C is pretty accepted just indicating the unit would be fine (guessing thats not the case tho =)

JazzFisch:
@williamwoldum can you remove the `, { refreshInterval: 1500, }` portion of the useWidgetAPI calls?  We're working on guidance for how and when to do refreshes, but currently (with one exception we're going to review) the only apps allowed to do refreshes are the service widgets and the media widgets.  The concern comes from potentially dozens of requests per second being sent through homepage.  Again, this is something the team is reviewing on how we can provide guidance and possibly configuration on.

williamwoldum:
In the years I've been 3d printing I've yet to come across a video, guide, or tutorial using Fahrenheit when talking 3d printer stuff. So for simplicity, I went with @shamoon's "idea" of just adding the 'C' unit to the temps. That's just my experience of course.

image

Also, I tried the units metric/imperial option but could not figure out how to access the yaml settings inside the `components.jsx` file 🤷‍♂️ I understand how it can be accessed inside the `widgets.js` file but the Octoprint API does not support any unit conversion queries. 
I tried to look at the openmeteo informationwidget but I did not understand how to properly apply getting the yaml options on my ServiceWidget.

JazzFisch:
I think your solution for °C looks fine.  Didn't realize that the unit conversion stuff wasn't more accessible.  Sorry for the wild goose chase.  Seems like a facility we might want to look at generalizing.

src/widgets/octoPrint/component.jsx
```diff
@@ -0,0 +1,64 @@
...
+        <Block label="octoPrint.temp_bed" value={`${printerStats.temperature.bed.actual} °C`} />
+        <Block label="octoPrint.job_completion" value={`${completion.toFixed(2)}%`} />
+      </Container>
+    );
+  }
+
+  return (
+    <Container service={service}>
+      <Block label="octoPrint.printer_state" value={printerStats.state.text} />
+      <Block label="octoPrint.temp_tool" value={`${printerStats.temperature.tool0.actual}°`} />
```
JazzFisch, line 60: Minor quibble, can you format these two lines like the two lines at 50 and 51?

williamwoldum: No problem, should be fixed now. Didn't realize I missed those 🤦

JazzFisch: Happens to the best of us.  😁

shamoon:
I think we missed the fact that this uses the case-sensitive `octoPrint` as it’s key, nothing else does that and we should probably change this, unfortunately, even though it’s been released 🤔

dwolfuk40:
> Thanks for the widget.
> 
> * Yes, the translation thing =)
> * Widgets shouldn't have more than 4 blocks, please lop 2 off
> * It looks like your proxy has no purpose other than the error 500 thing right? This is an odd one, but I get the thought process. My issue is a user will never be able to tell if they've configured it wrong, it will just always show shutdown. Not sure what the solution is here. Either way, instead of overwriting the proxy (which will complicate upkeep etc) you can just catch this in the component. But yea, still a little unsure about the "500 is expected" thing. I see how your `silencePrinterNotFound` setting helps but not sold on that either. Hmm

I get what you're saying about the widget "standard" but the extra blocks would be so usefull for many, myself included. Is there no way to have both. A standard setting and expanded setting like with the resource widgets maybe so it can conform to the standard but offer more if required. I just think there's so much potential useful info and thats the whole point of the homepage. I don't want to see "some" info like percentage complete and then have to visit octoprint just to find more such as estimated completion time as that defeats the purpose for me when it could (and probably should) be all there :)

jdrusso:
I do agree it would be great to have a way around the big red error box. The printer being off is not an error state, and the API does return a valid status even when the printer is off.

Thanks for the great plugin!

shamoon:
@jdrusso please create a FR for that if you’d like, there’s no way to track comments on merged PRs :)

PR #847, opened 2023-01-19
Synology Diskstation Widget

Oupsman:
It will break the alias between downloadstation and diskstation widget.

Configuration is as such : 

    - Name:
        icon: synology.png
        href: URL
        description: description
        widget:
          type: diskstation
          url: url
          username: username
          password: password

It shows 5 fields : Uptime, Volume Usage (in percent), Total volume size, CPU Load and memory usage

EDIT : it is tested against DSM 5, 6 and 7.

### DIFF
--- a/src/widgets/opnsense/component.jsx
+++ b/src/widgets/opnsense/component.jsx
@@ -31,9 +31,7 @@ export default function Component({ service }) {
 
   const cpuIdle = activityData.headers[2].match(/ ([0-9.]+)% idle/)[1];
   const cpu = 100 - parseFloat(cpuIdle);
-  const memoryInfos = activityData.headers[3].split(" ");
-  const totalMemory = parseFloat(memoryInfos[1]) + parseFloat(memoryInfos[3]) + parseFloat(memoryInfos[5]) + parseFloat(memoryInfos[7]) + parseFloat(memoryInfos[9])/1024 + parseFloat(memoryInfos[11]);
-  const memory = ( 1 - parseFloat(memoryInfos[11]) / totalMemory) * 100;
+  const memory = activityData.headers[3].match(/Mem: (.+) Active,/)[1];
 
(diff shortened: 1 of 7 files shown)


JazzFisch:
I rather wonder if work shouldn't be done to consolidate this widget's proxy and download stations proxy.  Both are talking to endpoints that operate essentially identically with regards to session management, etc.

Oupsman:
Indeed a generic Synology widget can be a good idea, you just select the module you want to get infos from : diskstation, downloadstation, vmm, etc etc. Basically the logic will be the same.

JazzFisch:
> Indeed a generic Synology widget can be a good idea, you just select the module you want to get infos from : diskstation, downloadstation, vmm, etc etc. Basically the logic will be the same.

I wasn't referring to a single consolidated widget, I just meant a shared Synology proxy.  I'd rather stick to an approach where each widget is focused and simple to configure rather than one widget to rule them all.

src/widgets/opnsense/component.jsx
```diff
@@ -24,11 +24,12 @@ export default function Component({ service }) {
         <Block label="opnsense.memory" />
         <Block label="opnsense.wanUpload" />
         <Block label="opnsense.wanDownload" />
+        <Block label="opnsense.wanUploadRate" />
+        <Block label="opnsense.wanDownloadRate" />
```
shamoon, line 28: This doesnt belong here. In general the commit history here is messy, as we've discussed in the past, it would be helpful to squash these commits.

shamoon:
> I wasn't referring to a single consolidated widget, I just meant a shared Synology proxy. I'd rather stick to an approach where each widget is focused and simple to configure rather than one widget to rule them all.

Yea great idea 💯

Oupsman:
I'm going to close the PR, until I figure how to squash those commits

PR #850, opened 2023-01-20
Revert "use unprivileged user in a container"

### LINKED ISSUE #849: Docker status broken in 0.6.0
_Originally posted by @KillahB33 in https://github.com/benphelps/homepage/issues/848

shamoon:
Reverts benphelps/homepage#644

Closes #849

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,12 @@ WORKDIR /app
 
 COPY --link package.json pnpm-lock.yaml* ./
 
-SHELL ["/bin/ash", "-xeo", "pipefail", "-c"]
-RUN apk add --no-cache libc6-compat \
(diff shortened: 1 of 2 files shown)


KillahB33:
I don't think it's necessary to revert this change, like the entrypoint states it will run by default as 0:0 unless specified other wise. Maybe just update the docs to include info on what permissions are needed.

shamoon:
Even if its just making the release notes and documentation clearer Im ok with having to re-release for that, its one thing to expect an update to break your install, its another to not! Even if its just to buy us a couple of days to sort it out its nbd to re-release those changes next time

KillahB33:
LOL very good point!

Ok sounds good!

PR #871, opened 2023-01-24
Enhancement: support specifying widget fields in docker labels / kubernetes annotations

jameswynn:
At least in Kubernetes, the field list is not being parsed. This PR detects if the fields list is a string and if it is parses it as JSON. This resolves the issue for Kubernetes widgets and should not disturb existing functionality.

### DIFF
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -218,9 +218,11 @@ export function cleanServiceGroups(groups) {
           wan // opnsense widget
         } = cleanedService.widget;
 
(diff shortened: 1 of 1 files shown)


shamoon (approved):
Thanks. Yea change seems fine tho I’m curious why this would be getting parsed as a string, maybe a nesting thing

jameswynn:
@shamoon In Kubernetes the annotations can only be strings, so the array has to be stringified. From what I can tell, Docker labels need the same treatment, but I've not tried Docker discovery.

PR #873, opened 2023-01-24
Services are now sorted by a new 'weight' field

### LINKED ISSUE #673: [Feature Request] Fixed order of docker defined services?
### Description

Currently, it seems services that are defined via docker are ordered in the reverse order to how they were spun up, i.e. the last service you brought up via docker will be first on the list. This creates some confusion as services on the home page can sometimes jump around.

Would be great if we could define a sorting method for docker services, this could even be as simple as a new field in the docker.yaml file like:

```
my-docker:
    host: homepage-socket-proxy
    port: 2375
    order: alphabetical
```

### Other

_No response_

jameswynn:
* Default for discovered services is 0
* Default weight for configured services is their index within their group scaled by 100, i.e. (index + 1) * 100
* Should be backwards compatible with current loose ordering
* If two items have the same weight value, then they will be sorted by name

closes #673

### DIFF
--- a/src/utils/config/api-response.js
+++ b/src/utils/config/api-response.js
@@ -13,6 +13,17 @@ import {
 } from "utils/config/service-helpers";
 import { cleanWidgetGroups, widgetsFromConfig } from "utils/config/widget-helpers";
 
+/**
+ * Compares services by weight then by name.
+ */
+function compareServices(service1, service2) {
+  const comp = service1.weight - service2.weight;
+  if (comp !== 0) {
+    return comp;
+  }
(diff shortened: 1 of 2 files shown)


JazzFisch:
I think you are right @jameswynn and that this would fix #673.  Can you update your commit message so that GitHub will auto close that issue when this is merged?

JazzFisch:
@jameswynn hope you don't mind, I updated your commit message with the verbiage and merged.

shamoon:
Great idea mate, you been crushing it :)

jameswynn:
Thanks @shamoon and @JazzFisch!

PR #879, opened 2023-01-25
Add Proxmox Backup Server Widget

Eizock:
I've added support for displaying info from Proxmox Backup Server.
To use it:
````
type: proxmoxbackupserver
url: https://your.pbs.ip:8007
username: api_token_id
password: api_token_secret
````
image

### CODE REVIEW

File: src/utils/proxy/handlers/credentialed.js [hunk 1]
```diff
@@ -34,6 +34,9 @@ export default async function credentialedProxyHandler(req, res, map) {
         headers.Authorization = `Bearer ${widget.key}`;
       } else if (widget.type === "proxmox") {
         headers.Authorization = `PVEAPIToken=${widget.username}=${widget.password}`;
+      } else if (widget.type === "pbs") {
+        headers = {};
+        headers.Authorization = `PBSAPIToken=${widget.username}:${widget.password}`;
```
shamoon, line 39: Did you do this because including the `Content-Type` doesnt work with proxmox? If so, please just remove the property rather than re-creating an empty object. Otherwise, leave it alone. Id be very surprised if it did since this is a json api...

shamoon (changes requested):
Thanks for the PR. Generally looks good, I left a couple comments as you see.

Since I dont have proxmox to test with, can you provide relevant sample API output for this?

Also, Im again flagging that I dont think this widget has been requested by anyone (other than yourself =) That's something for us to decide about

src/widgets/widgets.js
```diff
@@ -92,6 +93,7 @@ const widgets = {
   opnsense,
   overseerr,
   paperlessngx,
+  pbs,
```
shamoon, line 96: Sorry this comment got left off.

Please use a more meaningful name, e.g. presumably `proxmoxbackupserver`. Its a mouthful but more useful than just e.g. `pbs`

src/utils/proxy/handlers/credentialed.js
Eizock: Yeah, sadly this currently seems broken.
I'll probably report this to Proxmox later, until then imma have to delete it.

src/widgets/widgets.js
Eizock: I changed it, but the type changed aswell.
So to use it, it has to be done like this:
```
type: proxmoxbackupserver
url: https://your.pbs.ip:8007
username: api_token_id
password: api_token_secret
```

shamoon:
Thanks for api output, we’ll give it a test. 

Yea no worries low stakes :) It’s come up a couple times so far, generally we’ve been pretty willing to merge widgets that don’t have high (or any) demand. But to answer your question it’s because anything we include in the codebase comes with an expectation that it works (obviously :), which means supporting it, which can be significantly time-consuming or challenging especially for things like this which aren’t a simple “spin up a docker container to test against”.

The good news in this case is that widgets like this which don’t have a custom proxy etc are generally way less error-prone so less to “worry” about.

I’m def not saying we won’t merge (tbh, we prob will), just giving you some context as to why it’s worth considering.

Eizock:
Sweet.

I totally get your worries. If you want, I could open a issue or similar, bevore putting in a pr?

JazzFisch:
> Sweet.
> 
> I totally get your worries. If you want, I could open a issue or similar, bevore putting in a pr?

That can be really helpful for us, and also give the community an opportunity to provide feedback regarding features and demand as well.  Thanks for wanting to contribute!

shiftoptk:
As a PROXMOX VE user, I'm glad to see PBS has a widget in development. Can't wait to try it.

shamoon:
Alright, no reason to quibble here, again it’s a simple-enough one. Thanks for the PR

shiftoptk:
Apologies, didn't mean to sound terse; I just set up Homepage and got all my widgets (to include the PVE widget) humming along. I went seeking any PBS widget feature request and was pumped to see the pull request. Very much appreciate the project, thanks you for you effort!

kloknibor:
How do you need to create the API key and user. I tried to make a user like with proxmox, but I keep getting the following error:

API Error: Invalid data
URL: https://xxx.xxx.xxx.xxx:8007/api2/json/nodes/localhost/status
Response Data:
permission check failed

Eizock:
@kloknibor 
I have it like this, the user and the API-Key needs the permission (atleast on PBS):
grafik

kloknibor:
I got it to work :D I choose the every Audit I could find, but missed the normal Audit... Thanks!

0xAF:
> @kloknibor I have it like this, the user and the API-Key needs the permission (atleast on PBS): grafik

Please consider adding this to the widget info, since it's not that obvious. Thanks for the great work!

PR #889, opened 2023-01-28
Don't include LXC templates in LXC count

### LINKED ISSUE #864: [Feature Request] Proxmox Service - LXC Templates Included In Count
Just a minor thing, but the counter for LXC containers includes LXC templates. 

In my case, I have 1 template and 5 containers. In Homepage, this appears as 5 / 6 containers running. 

Could LXC templates be excluded from this counter? In my case this would mean a counter of 5 / 5.

JazzFisch:
fixes #864


sgastol:
We do need same for VMs, not to count templates.

PR #897, opened 2023-01-30
Added search to quicklaunch

avdept:
Hey folks. So I've added this little part that adds option to search for item entered to quick launch panel. This is critical to those who needs google for something without moving hands away from keyboard to change focus on search input field. Please let me know if there needs something else to be added.

Screenshot 2023-01-30 at 10 19 20

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -283,7 +283,7 @@
     "quicklaunch": {
         "bookmark": "Bookmark",
         "service": "Service",
-        "searchGoogle": "Search Google"
+        "search": "Search"
     },
     "wmo": {
         "0-day": "Sunny",
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -282,7 +282,8 @@
     },
     "quicklaunch": {
         "bookmark": "Bookmark",
-        "service": "Service"
+        "service": "Service",
+        "searchGoogle": "Search Google"
     },
     "wmo": {
         "0-day": "Sunny",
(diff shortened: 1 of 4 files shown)


shamoon (changes requested):
Thanks for the PR. I've been meaning to add this.

We can't accept a hard-coded Google search here though, our users wouldn't really stand for that =)

This needs to use the configured search widget

avdept:
> Thanks for the PR. I've been meaning to add this.
> 
> We can't accept a hard-coded Google search here though, our users wouldn't really stand for that =)
> 
> This needs to use the configured search widget

Yeah that makes sense. Just wanted to find out if this kind of change generally allowed. I will add more code to allow users to select preferable engine based on config.

shamoon:
You can always open a discussion to discuss

Just to clarify, I don’t mean a separate configuration for this, I mean whatever search provider is configured as a widget

we can, of course, help with that if you’d like

JazzFisch:
It would be nice if there was a way to not have the internet search provider checked when doing a quick search.

shamoon:
Take a look, I updated this to:

- Use the configured search provider
- Make quick launch internet search hideable (`hideInternetSearch: true`)
- Also changed what the title shows
- Also shows type `Search` not service

Screenshot 2023-01-30 at 11 34 39 PM
Screenshot 2023-01-30 at 11 34 45 PM

avdept:
Great work! Was in middle of doing it myself, but glad you made it first!

avdept:
Also I've noticed its not going to work with custom search provider, only with default

shamoon:
> Great work! Was in middle of doing it myself, but glad you made it first!

Sorry about that mate!

> Also I've noticed its not going to work with custom search provider, only with default

Ah, good point. Gotta test this.

shamoon:
Ok custom should be sorted too, thanks for pointing that out

PR #899, opened 2023-01-30
Support multi search providers

chain710:
support multiple `provider` for option in search widget

Example:

```yaml
- search:
    provider: [baidu, google, duckduckgo]
    target: _blank
```

When `providers` is configured, the search button is replaced with a ListBox, from which a search engine can be selected.

Look like this

image

### DIFF
--- a/src/components/widgets/search/search.jsx
+++ b/src/components/widgets/search/search.jsx
@@ -41,33 +41,12 @@ function classNames(...classes) {
   return classes.filter(Boolean).join(" ");
 }
 
-function useProviderState() {
-  const key = "search-name";
-
-  const [value, setValue] = useState(providers.google);
-  useEffect(() => {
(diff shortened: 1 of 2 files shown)


shamoon:
Thanks for the PR, Im not sure how much interest there is in this but seems like a reasonable "progressive enhancement". A couple things:

- I refactored this to keep using just the `provider` key. If someone passes one value, e.g. as most have, nothing changes, but this can now be changed to an array like you specified. But in general I think its cleaner to keep the same key (even though its "singular"
- I refactored the component to always use the listbox, just disable the dropdown if only 1 provider
- Finally, initially this was falling back to using Google in (I think) unintended ways, also fixed that

Of course let me know your thoughts!

chain710:
👍  LGTM

shamoon:
Cool!

Though I realized this will conflict with #897, should be sorted

Issue #235, opened 2022-09-20
[Topic] Layout & Positioning Issues


benphelps:
This issue will serve as a single point of discussion for anything related to layout or positioning issues, bugs, suggestions, etc.

AmadeusGraves:
CUSTOM CSS 
Apart from the .yml files that exist in the configuration folder, a css class could also be instantiated to allow editing of the page style. This way the developer will be more free from design issues like on certain wallpapers with certain colours, it is not possible to read the data displayed on the screen.

Spiritreader:
It would be wonderful to be able to display a custom logo to the left of the resource monitoring information.
![image](https://user-images.githubusercontent.com/18438899/191610769-ef0fd104-0af6-46c0-888e-674039d61e0f.png)

ysarkain:
2 requests:
1) Adjustable width of page - on my 2560x1440 monitor it looks like its ~75% so there is lots of space on both sides.
2) Also would be cool if we could put the speedtest widget in the header bar area with the disk/cpu/etc. statistics.

iamernie:
> Hi, could you please tell me how I can get this layout config?
> 
> <img alt="190466646-8ca94505-0fcf-4964-9687-3a6c7cd3144f" width="1260" src="https://user-images.githubusercontent.com/106855541/193499239-1969527e-b44b-4459-af4a-b4c86f4f762c.png">
> 
> I have the current layout settings in my yml file:
> 
> layout: Media: style: row columns: 4 Downloads: style: row columns: 1 Utilities: style: row columns: 1
> 
> I have tried a few different variations of the above, but nothing that will replicate the above picture :(

Try this:
layout:
Media:
style: row
columns: 4
Downloads:
style: column
Utilities:
style: column

dieideeistgut:
Heya. Please don't over-complicate this one. I'd vote for a simple single-css-file-include via volume/bind. If there is no file, there will be nothing included.

chuckiton:
<img width="1568" alt="Screenshot 2022-10-12 at 14 51 06" src="https://user-images.githubusercontent.com/90910365/195360997-27706564-e8dc-4af7-8f49-e83f6c32a541.png">
Search bar doesn't extend to the boundary of the date/time info widget
Contents of widgets.yaml:
```
- resources:
    label: System
    cpu: true
    memory: true
- resources:
    label: mSD Storage
    disk: /
- resources:
    label: Disk Storage
    disk: /mnt/storage
- search:
[...]

PR #904, opened 2023-01-31
feat: Custom CSS

lholota:
Hi @benphelps,
based on the discussions in #235, #139  and similar issue I have encountered myself (trying to insert a logo which I want to be wider than 48px), I would suggest adding support for custom css to cater for unique and custom usecases users may have. This way everyone can simply override the style they want to customize.

I'd be of course happy to update the docs accordingly in the benphelps/homepage-docs repo if this change is accepted.

Side note: I have considered adding custom styles via settings but it's troublesome because the imports must be static otherwise Webpack build fails. There could be probably ways to use dynamic import but tbh this seems much easier and transparent.

### DIFF
--- a/src/pages/_app.jsx
+++ b/src/pages/_app.jsx
@@ -6,6 +6,7 @@ import Head from "next/head";
 import "styles/globals.css";
 import "styles/theme.css";
 import "styles/manrope.css";
+import "styles/custom.css";
 import nextI18nextConfig from "../../next-i18next.config";
 
 import { ColorProvider } from "utils/contexts/color";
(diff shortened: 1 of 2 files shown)


shamoon:
Simple enough, thanks. Yea feel free to add PR to the docs too. Since homepage uses tailwind the css classes are not going to be all that straightforward but its there for folks who want to

shamoon:
Honestly I was on auto-pilot when I approved this, adding CSS this way doesn't work unless you re-build the docker container because its pre-compiled by next-js, which clearly wouldn't be the expectation for this feature. We're going to refer the change.

Sorry for confusion.

IIIdefconIII:
would be aweome to add something in setting.yml to choose how many columd in widith is allowed, myself would love to see 5 colums:

PR #909, opened 2023-02-01
Split jellyfin out to its own widget as it has diverged from emby

cjd:


shamoon:
I get the idea here but this is kinda pointless at the moment, no? This is literally the exact current version of the emby widget so what is the purpose of separating them besides giving us two copies of the code to maintain? If the two services diverge in a meaningful way then sure, but the semantic 'benefit' of this doesn't seem to outweigh the technical debt.

cjd:
There is a (very) slight difference in that the endpoint used to access the api changed from  "emby/Sessions/{sessionId}/Playing/{command}" to just "Sessions/{sessionId}/Playing/{command}" and so the emby widget no longer works for jellyfin.
This is in line 13 of src/widgets/jellyfin/widget.js but since this diff shows it as a new file the diff doesn't show up clearly :)

shamoon:
Ah, thanks. Well that's lame (of them). Will have a think about how to handle this

cjd:
Some other options would be to add a test to the emby-handling code which checks if you are accessing it a jellyfin, and adjust the endpoint accordingly, or to pass the endpoint prefix somehow (ie: pass 'emby/' if running emby and pass '' if running jellyfin)

shamoon:
Yea I think something like that would be best, should be straightforward. Again it's very un-DRY to have the entire widget code duplicated for 1 line change. Let me know if you'd like to update this / open a new PR or of course we can take care of that

shamoon:
So I just tested the current versions of both and seems to work fine...

Screenshot 2023-02-01 at 11 34 45 PM
Screenshot 2023-02-01 at 11 36 38 PM

?

PR #912, opened 2023-02-01
Fix: Show container status if health is an empty string

eanmclaughlin:
If data.health is an empty string, ?? will accept it and not use the right side value; instead, we should use || which treats empty string as false and returns the right side value. This will show "RUNNING" in the status box if health is "".

### DIFF
--- a/src/components/services/status.jsx
+++ b/src/components/services/status.jsx
@@ -30,8 +30,8 @@ export default function Status({ service }) {
     }
 
     return (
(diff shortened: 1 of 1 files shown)


eanmclaughlin:
Exemplary screenshot:

image

shamoon:
In what situation is health ""?

shamoon:
Sorry misread this! =)

shamoon:
Thanks

PR #914, opened 2023-02-02
Implement uptime-kuma widget

chain710:
Inspired by [homer's UptimeKuma service](https://github.com/bastienwirtz/homer/blob/main/src/components/services/UptimeKuma.vue)

config like:
```
widget:
  type: uptimekuma
  url: http://uptime-kuma-host:port
  slug: statuspageslug
```

look like:

Edit:
Screenshot 2023-02-02 at 12 27 43 AM


shamoon:
Thanks for this. Quick thing: You should only update the en/common language file and the rest are updated via weblate

Question: I thought uptime doesn’t have an [public] api yet? Is this just checking the status of a single site?

chain710:
> You should only update the en/common language file and the rest are updated via weblate

Sorry I'm unfamiliar with weblate, I'll revert other language files. Btw: where can I add chinese translation?

> Question: I thought uptime doesn’t have an [public] api yet? 

You mean whether api stable enough? Based on my personal experience using Homer's UptimeKuma service for half a year and upgrading it several times, it seems to be working pretty well.

> Is this just checking the status of a single site?

It checks all monitors of  a single [status-page](https://github.com/louislam/uptime-kuma/wiki/Status-Page)

> edit: it’s also not using our regular widget block structure

Yeah, I also tried Block, but after comparing I feel that the current style is more suitable for UptimeKuma. If you have any ideas for adjustments, let me know.

shamoon:
I cleaned up the commits, you still had changes to 40 files.

Ah, Im not an uptime user but last time I looked IIRC their documentation indicated no api but I do see you're hitting /api/... endpoints of course.

We have to think about the layout thing, we very much strive for standardization around here...

shamoon:
Hah, you fixed the commmits too, thanks =)

shamoon:
So "Partially degraded service" just means "some of your sites are down"? Is that very helpful without more specifics?

chain710:
> We have to think about the layout thing, we very much strive for standardization around here... 

Then I shoud switch back to block structure, will update later

shamoon:
Nice, definitely like that more 👍

I think there's still a question of the utility of this info but I suppose we can see what users have to say.

As for the code, I haven't dug into it too much but I don't think you really need the proxy? In general we try to avoid them since they require much more individual maintenance etc. It looks like you're just hitting to endpoints and manipulating the data, which can be done in the component code and moving the endpoints into widget.js. Let me know if you think that's not possible for some reason or of course if you'd like help with that

shamoon:
Ps. it looks like this code is literally copypasta'd from home with some variable changes, which should absolutely be noted in the code (that project uses Apache 2.0 so should be ok but...)

shamoon:
See my commit =)

Screenshot 2023-02-02 at 12 27 43 AM

shamoon:
That's going to be a problem because the icon is a relative URL, you'd need the full URL, which you won't have (you can take from the href but that's not guaranteed to be the same thing I suppose)

shamoon:
So, what about this:

Screenshot 2023-02-02 at 1 08 59 AM
Screenshot 2023-02-02 at 12 51 11 AM

shamoon:
Will you test #916 ? Its the different version of this

shamoon:
`widget.url` gets stripped out

PR #922, opened 2023-02-02
Komga Widget

KillahB33:
I have tried everything I can think of. Network is showing me the correct values for Komga but they never populate in homepage. I completely reworked everything last night and it's still failing to populate so I think I need a second set of eyes.


shamoon:
Please stop fiddling with the commits, they're kinda messed up now but I'll take a look

shamoon:
Think you just had the wrong variable for when data hasn't returned yet

Screenshot 2023-02-02 at 11 06 51 AM

shamoon:
@KillahB33 let me know if any issues, overall widget works great, problem was just a small error in the code

KillahB33:
Feel like a goof that I mixed that up, thanks!!
Everything looks good on my side! Glad I made the changes to where I grab the numbers from, looks like actuator/metrics isn't up to date or delayed some how so this approach is better.
Also my bad on all the commits, saw your message.

PR #953, opened 2023-02-06
Revert "feat: Custom CSS"

shamoon:
Reverts benphelps/homepage#904

Honestly don't know what I was thinking. This doesn't work and has been discussed before because CSS gets compiled at buildtime.

Sorry for the space cadet moment


dieideeistgut:
I was confused too when i saw you merging this - added an idea that may be more in line with your overall philo for this project - https://github.com/benphelps/homepage/pull/1003

PR #963, opened 2023-02-08
Fix settings after container update

### LINKED ISSUE #576: [Bug] Layout settings not picked up after update

shamoon:
Ok, this is another approach to try and fix #576 (and it reverts the other attempt).

The idea here is to include `NEXT_PUBLIC_BUILDTIME` in the config hash, which in effect will force revalidation after the container is updated even if the settings haven't changed.

It's not perfect, if someone recreates the container it'll still happen and folks building their own image will need to supply the build time arg for this to work. But for the majority of users who are seeing this e.g. after watchtower updates this should help.

Still welcome other solutions here, especially to the post-recreation part

Closes #576 (again)

### DIFF
--- a/src/pages/api/hash.js
+++ b/src/pages/api/hash.js
@@ -19,7 +19,10 @@ export default async function handler(req, res) {
     return hash(readFileSync(configYaml, "utf8"));
   });
 
-  const combinedHash = hash(hashes.join(""));
+  // this ties hash to specific build which should force revaliation between versions
+  const buildTime = process.env.NEXT_PUBLIC_BUILDTIME?.length ? process.env.NEXT_PUBLIC_BUILDTIME : '';
+
+  const combinedHash = hash(hashes.join("") + buildTime);
 
   res.send({
     hash: combinedHash,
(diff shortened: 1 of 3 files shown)


jacksgt:
Hi there, 

I'm running `ghcr.io/benphelps/homepage:v0.6.10` (which should include commit https://github.com/benphelps/homepage/commit/45946ddfe176b7ea58cbd00d537fad8bd781087a from this MR), but I'm still seeing the behavior that I need to manually reload the config to apply `settings.yaml`.

Any ideas? Should I create a new issue for this or should we re-open https://github.com/benphelps/homepage/issues/576 ?

shamoon:
Between version updates you have to reload or when re-creating the container? The first thing should be sorted, the second is starting to feel more like a known limitation…

jacksgt:
The issue appears when "re-creating" the container - as well as starting it for the first time.

PR #964, opened 2023-02-08
Fix: cookieJar.setCookieSync ignore error

chain710:
cookieJar.setCookieSync crash node when error, like:
```
uncaughtException: Error: Cookie not in this host's domain. Cookie:example.com Request:127.0.0.1
```

### DIFF
--- a/src/utils/proxy/cookie-jar.js
+++ b/src/utils/proxy/cookie-jar.js
@@ -33,8 +33,7 @@ export function addCookieToJar(url, headers) {
     cookies = [cookie];
   }
 
-  const options = { ignoreError: true };
   for (let i = 0; i < cookies.length; i += 1) {
-    cookieJar.setCookieSync(cookies[i], url.toString(), options);
+    cookieJar.setCookieSync(cookies[i], url.toString(), { ignoreError: true });
   }
 }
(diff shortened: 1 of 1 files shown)


src/utils/proxy/cookie-jar.js
```diff
@@ -33,7 +33,8 @@ export function addCookieToJar(url, headers) {
     cookies = [cookie];
   }
 
+  const options = { ignoreError: true };
```
shamoon, line 36:
```suggestion
```

src/utils/proxy/cookie-jar.js
```diff
@@ -33,7 +33,8 @@ export function addCookieToJar(url, headers) {
...
   for (let i = 0; i < cookies.length; i += 1) {
-    cookieJar.setCookieSync(cookies[i], url.toString());
+    cookieJar.setCookieSync(cookies[i], url.toString(), options);
```
shamoon, line 38:
```suggestion
    cookieJar.setCookieSync(cookies[i], url.toString(), { ignoreError: true });
```

shamoon (changes requested):
Thanks. Under what circumstances are you seeing this error? I never have

See the code changes, no reason to create a const if it never gets used again

PR #988, opened 2023-02-12
Move the footer to the bottom

vorotynsky:
I moved the footer to the bottom of the screen.

If the home page is smaller than a screen (for example, without bookmarks), the footer is in the middle of the screen. I made the max page height to move the footer down.


shamoon:
Sure thanks

Curios, do you actually have so few services that this is happening or did you just notice it in testing?

vorotynsky:
I skipped bookmarks and placed services in 2 rows.

PR #994, opened 2023-02-13
Fix: improve swarm status for replicated services & prefer stats for local containers

### LINKED ISSUE #970: Swarm Mode
Hi,

I have swarm mode setup as per the instruction manual 

```
docker-swarm:
  socket: /var/run/docker.sock
  swarm: true
```

I then reference the Server and the Service name under my service.

This works, but only when the service I'm referencing is on the same node as the dashboard container.

For example, I have a 3 node swarm cluster.

The dashboard service sits on Node 1
Adguard is also on Node 1
Unifi is on node 2

The dashboard will only show Adguard as healthy/online as its on the same node as the dashboard service.

Unifi container shows as Unknown as its on another node.

If I force Unifi onto the same node as the dashboard service it shows as available.

So.....am I doing something wrong here? Shouldn't this work across all nodes? Or have I setup the Docker service wrong in the config?

shamoon:
Digging into swarm a bit, currently the code just chooses the 0th task for a service which might be on a different node, in which case we can't get container stats, so this PR tries to find one running locally which fixes the status and stats badge (though stats are still only for 1 task as current).

Closes #970 (which may actually be partly a separate issue too)

@vinaydawani would be great if you could check this out, again Im just playing with swarm for testing purposes...

Edit: this now also contains changes from #998 which improved swarm status with N replicas info

### DIFF
--- a/src/components/services/status.jsx
+++ b/src/components/services/status.jsx
@@ -12,7 +12,7 @@ export default function Status({ service }) {
     </div>
   }
 
-  if (data && data.status === "running") {
+  if (data && data.status.includes("running")) {
     if (data.health === "starting") {
       return (
         <div className="w-auto px-1.5 py-0.5 text-center bg-theme-500/10 dark:bg-theme-900/50 rounded-b-[3px] overflow-hidden" title={data.health}>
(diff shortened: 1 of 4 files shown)


Zareix:
Looks good to me 👍

PR #998, opened 2023-02-14
Improve Docker Swarm status check

Zareix:
This PR improves the docker swarm status by checking the mode (replicated or global) and if the number of tasks corresponds to the expected one.

This also removes the `Docker.Container.inspect()` method that couldn't work if the container was not on the manager node (addressing PR #970)

Remark:
The only problem that remains is for the stats where it will only work if the container is on the manager. There is (as far as I know) no way to prevent this considering the actual Docker API.

### DIFF
--- a/src/components/services/status.jsx
+++ b/src/components/services/status.jsx
@@ -7,15 +7,21 @@ export default function Status({ service }) {
   const { data, error } = useSWR(`/api/docker/status/${service.container}/${service.server || ""}`);
 
   if (error) {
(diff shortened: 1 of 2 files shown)


Zareix:
Just saw #994. I think we should keep #994's stats part (as there is no other way) and my status part (using task info)

shamoon:
I can see how the partial info is valuable in swarm. How about we bring this change into that PR and merge the two ideas

shamoon:
I can "cherry pick" the commit, still retaining you as the author

PR #1003, opened 2023-02-15
Proposal / Idea for customCSS loading

dieideeistgut:
This would be my idea for implementing custom CSS loading.

from widgets.yaml:

```
- customcss:
    src: /customcss/styles.css
```

I did some research - afaik is loading css in body valid nowadays (body-ok).

Please feel free to deny this PR - it should serve more as an idea.

### DIFF
--- a/src/components/widgets/customcss/customcss.jsx
+++ b/src/components/widgets/customcss/customcss.jsx
@@ -1,7 +1,7 @@
 export default function Customcss({ options }) {
   if (options.src) {
     return (
-      <link href={options.src} type="text/css" rel="stylesheet"></link>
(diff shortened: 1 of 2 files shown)


shamoon:
Thanks but this has the same issue as #793. Externally loaded css doesn’t perform well in a react app and kinda goes against the static site concept

dieideeistgut:
True & valid point. Would you be fine with creating more / custom themes though?

NiciTheNici:
Is there currently a way to change the default colors?

I'd love to customize so that this app too uses [catppuccin](https://github.com/catppuccin/catppuccin) :)

PR #1026, opened 2023-02-20
Pterodactyl widget: initial commit

### LINKED ISSUE #930: [Feature Request] Pterodactyl Panel
### Description

would love to have a widget like for pterodactyl panel 

showing servers, servers online/offline etc..

### Other

_No response_

Oupsman:
Widget has two fields:
Nodes (Online/Total)
Servers (Running/Total)

## Proposed change

Closes #930 (issue) : https://github.com/benphelps/homepage/issues/930

## Type of change

- [X] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/40
- [ ] 


shamoon:
Please provide sample API output, this doesn't seem trivial to setup which is always challenging with a custom proxy.

Speaking of which, there are some issues with it, I will leave comments but again ideally we could test against the app.

src/widgets/pterodactyl/proxy.js
```diff
@@ -0,0 +1,88 @@
...
+
+      const nodesData = JSON.parse(data);
+      const nodesTotal = nodesData.data.length;
+      let nodesOnline = 0;
+      const serversIds = [];
+
+      for (let nodeid = 0; nodeid < nodesData.data.length; nodeid += 1) {
+        // check if node is online
+        const nodeURL = `${nodesData.data[nodeid].attributes.scheme}://${nodesData.data[nodeid].attributes.fqdn}:${nodesData.data[nodeid].attributes.daemon_listen}/api/system`;
+        // eslint-disable-next-line no-await-in-loop
```
shamoon, line 43: The reason this throws a lint error is because using await inside a loop doesn't work like you think it does. You likely need to use `Promise.all`. In general with these if you google the error you'll get a good explanation, e.g. https://eslint.org/docs/latest/rules/no-await-in-loop

src/widgets/pterodactyl/proxy.js
```diff
@@ -0,0 +1,88 @@
...
+          nodesOnline += 1;
+        }
+
+        for (let serverid = 0; serverid < nodesData.data[nodeid].attributes.relationships.servers.data.length; serverid += 1) {
+          serversIds.push(nodesData.data[nodeid].attributes.relationships.servers.data[serverid].attributes.identifier);
+        }
+      }
+
+      let online = 0;
+      /* eslint-disable no-await-in-loop */
```
shamoon, line 56: Same as above https://eslint.org/docs/latest/rules/no-await-in-loop

src/widgets/pterodactyl/proxy.js
```diff
@@ -0,0 +1,88 @@
...
+        logger.error("Unable to retrieve Pterodactyl nodes' list");
+        return res.status(status).json({error: {message: `HTTP Error ${status}`, url: nodesURL, data}});
+      }
+
+      const nodesData = JSON.parse(data);
+      const nodesTotal = nodesData.data.length;
+      let nodesOnline = 0;
+      const serversIds = [];
+
+      for (let nodeid = 0; nodeid < nodesData.data.length; nodeid += 1) {
```
shamoon, line 40: So do we have any idea of the potential number of API calls this could make? Obviously I don't use the software but Im guessing its generally low, but...

shamoon (changes requested):

JazzFisch:
@Oupsman can you please provide some guidance on how we can test all of this?

src/widgets/pterodactyl/proxy.js
Oupsman: In the current state of the API, there is no way to get all the statuses of the servers in one API call, sadly. 

So this code makes 2 calls per node (one to get the general information and the server list, the other to verify if it is online) and one API call per server, to get the server status. On very busy pterodactyl servers,it can make a great number of API calls.

src/widgets/pterodactyl/proxy.js
Oupsman: OK got it, I'll work on that.

Oupsman:
For testing, I can give you an access to my personnal server, just give me an email address for me to send the configuration to.

shamoon:
Thanks for the offer to use your setup, we may need to take you up on that but before that I think there’s still a fundamental question about the plan to hit each node multiple times. Simply put, I don’t think something like that belongs in homepage. These widgets are meant to be quick performant informational widgets.

If you remove the node calls, can you still get any useful information from the main server?

shamoon:
Hey @Oupsman , please see my last commit. As usual **this is just a suggestion, you can always revert the change**. Essentially it just shows node and server count. I realize it makes the widget less useful but it also avoids the issue above and doesn't require a custom proxy any more.

I did get a very basic pterodactyl instance setup but nothings really working so I can only test it out so much.

Again, let us know your thoughts, if you're not happy with this the change can be reverted but I don't currently see an alternate pathway to get the 'online' info that isn't so cumbersome...

Oupsman:
I liked the idea of having an information of the online nodes and the started servers, but I really understand your concerns and I agree with you : better not to use any proxy code. 

Why don't we publish the widget as you propose, and see if someone requests some more informations ?

shamoon:
That sounds reasonable, thanks for your understanding. Does the code work as-is for you? Again my instance is kind of a 'dummy' one.

shamoon (approved):

Issue #957, opened 2023-02-07
[Bug] Kubernetes Ingress Annotations with External Apps 


Aesop7:
### Description

The Ingress annotations only work if there is a corresponding Kubernetes app.  This causes an issue (at least for me) for apps that are hosted outside of the cluster, but utilize Kubernetes Ingresses (e.g. an externally-hosted version of PiHole that uses ingress-nginx for its SSL certificate)

When there is no corresponding k8s app for the Ingress, the service ends up getting rendered like this:

![image](https://user-images.githubusercontent.com/9644862/217161427-1e7a3c6b-3f4c-4ef4-9732-7a206355ca3f.png)

The issue is that small, empty box on the top right:

![image](https://user-images.githubusercontent.com/9644862/217161968-077c664a-4c93-46ef-b7f3-5674567c9c73.png)

I really like being able to use ingress annotations for Homepage, so I tried to adjust the annotations to have it display as a [ping](https://gethomepage.dev/en/configs/services/) instead, but that also failed to work.

Here's a copy of the external service's Kubernetes manifests:

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: pihole
spec:
  ports:
    - targetPort: 443
      port: 80
      protocol: TCP
  type: ClusterIP
  clusterIP: None
---
apiVersion: v1
kind: Endpoints
metadata:
  name: pihole
subsets:
  - addresses:
    - ip: 10.10.10.9
    ports:
    - name: http
      port: 80
      protocol: TCP
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: pihole
  annotations:
    kubernetes.io/ingress.class: "internal-nginx"
    cert-manager.io/issuer: "letsencrypt-prod"
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    gethomepage.dev/enabled: "true"
    gethomepage.dev/name: PiHole
    gethomepage.dev/description: Ad Blocker and Local DNS
    gethomepage.dev/group: Admin
    gethomepage.dev/ping: "https://pihole.my.network/admin"
    gethomepage.dev/icon: pihole.png
    gethomepage.dev/widget.type: "pihole"
    gethomepage.dev/widget.url: "https://pihole.my.network"
    gethomepage.dev/widget.key: "secretwidgetkey"
spec:
  rules:
  - host: pihole.my.network
    http:
      paths:
      - pathType: Prefix
        path: /
        backend:
          service:
            name: pihole
            port:
[...]

PR #1040, opened 2023-02-22
Added "external" boolean annotation to specify external services

jameswynn:
## Proposed change

External services define a kubernetes service pointing to an application hosted outside of the cluster. These services should not attempt to get the status based on kubernetes deployments. The new boolean flag disables this functionality. Since this is an edge case, the default value is "false".

Prior to this change the system would attempt to find a matching deployment and fail, resulting in a "not found" error and small box in the top right corner.

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: uptimekuma
  annotations:
    gethomepage.dev/enabled: "true"
    gethomepage.dev/name: Uptime Kuma
    gethomepage.dev/description: Service Uptime Monitoring
    gethomepage.dev/group: Admin
    gethomepage.dev/url: "https://uptime.k3d.localhost"
    gethomepage.dev/icon: uptime-kuma.png
    gethomepage.dev/external: "true"
```

This is a follow up to #957

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/item.jsx
+++ b/src/components/services/item.jsx
@@ -91,7 +91,7 @@ export default function Item({ service }) {
                   <span className="sr-only">View container stats</span>
                 </button>
               )}
-              {service.app && (
+              {(service.app && !service.external) && (
                 <button
                   type="button"
                   onClick={() => (statsOpen ? closeStats() : setStatsOpen(true))}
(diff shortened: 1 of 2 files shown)


shamoon:
Cool. Seems straightforward thanks

PR #1043, opened 2023-02-23
Added support for environment variable substitution

### LINKED ISSUE #60: [Feature Request] Allow variables in config file
Would it be possible to allow variables in the config files?

E.g.

```
- Media:
    - Sonarr:
        href: http://192.168.0.254:8989
        #description: Homepage is awesome
        server: myserver
        container: Sonarr
        widget:
          type: sonarr
          url: http://192.168.0.254:8989 # no slash at the end
          key: $SONARRKEY
```

Then in the compose file, we can do something like:
```
  homepage:
    image: ghcr.io/benphelps/homepage:latest
    container_name: Homepage
    environment:
      - SONARRKEY=xNYcIvYlbEiBXiIXJo3HDw32
    ports:
      - 8045:3000
    volumes:
      - $USERDIR/Homepage:/app/config
      - /var/run/docker.sock:/var/run/docker.sock # (optional) For docker integrations
```

(note: key used was randomly generated for this request)

jameswynn:
## Proposed change

* Only environment variables starting with HOMEPAGE_VAR_ and HOMEPAGE_FILE_ are supported
* The value of env var HOMEPAGE_VAR_XXX will replace {{HOMEPAGE_VAR_XXX}} in any config
* The value of env var HOMEPAGE_FILE_XXX must be a file path, the contents of which will be used to replace {{HOMEPAGE_FILE_XXX}} in any config
* If a substituted value contains a variable reference it may also be replaced, but the behavior is non-deterministic

Closes #60 

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/utils/config/api-response.js
+++ b/src/utils/config/api-response.js
@@ -4,7 +4,7 @@ import path from "path";
 
 import yaml from "js-yaml";
 
-import checkAndCopyConfig, { getSettings } from "utils/config/config";
+import checkAndCopyConfig, { getSettings, substituteEnvironmentVars } from "utils/config/config";
 import {
   servicesFromConfig,
   servicesFromDocker,
@@ -28,7 +28,8 @@ export async function bookmarksResponse() {
   checkAndCopyConfig("bookmarks.yaml");
 
   const bookmarksYaml = path.join(process.cwd(), "config", "bookmarks.yaml");
-  const fileContents = await fs.readFile(bookmarksYaml, "utf8");
+  const rawFileContents = await fs.readFile(bookmarksYaml, "utf8");
+  const fileContents = substituteEnvironmentVars(rawFileContents);
   const bookmarks = yaml.load(fileContents);
 
   if (!bookmarks) return [];
--- a/src/utils/config/config.js
+++ b/src/utils/config/config.js
@@ -2,7 +2,12 @@
 import { join } from "path";
 import { existsSync, copyFile, readFileSync } from "fs";
 
+import cache from "memory-cache";
 import yaml from "js-yaml";
+
+const cacheKey = "homepageEnvironmentVariables";
+const homepageVarPrefix = "HOMEPAGE_VAR_";
+const homepageFilePrefix = "HOMEPAGE_FILE_";
 
 export default function checkAndCopyConfig(config) {
   const configYaml = join(process.cwd(), "config", config);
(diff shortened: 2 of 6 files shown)


shamoon:
Awesome change, thanks.

This works great in my testing but one q: currently this would look like e.g.
```yaml
          widget:
               type: unifi
               ...
               password: UNIFI_PASSWORD
```

Since there isn't a standard for this in YAML, could we implement our own syntax for these like `{{UNIFI_PASSWORD}}` or `${{UNIFI_PASSWORD}}` or even `$UNIFI_PASSWORD` to make them more obvious? You could also then just loop through the variables that are found in the yaml and replace with the corresponding ENV as opposed to currently looping through every ENV each time (Im sure the performance is not a noticeable difference there, but still).

I don't feel super strongly about the above, just a thought / q for the group.

jameswynn:
It should be `{{HOMEPAGE_VAR_UNIFI_PASSWORD}}` right now.

Finding all of the variable references may also be expensive, but I can look into it. This approach can definitely be optimized further. We could store a cache of the usable variables and values as well. A simple search to see if a file contains any variable references at all would be a good step.

shamoon:
Ha, whoops yes you can ignore my comment.

Yea I don't feel strongly about doing it in the reverse, my guess is some kind of regex would be similarly expensive maybe even worse than just looping through the env array

shamoon (approved):
Will let the other folks take a look too.

shamoon:
In the meantime I can fiddle with caching / maybe optimizing a bit, unless you'd prefer to do that, of course.

jameswynn:
Sure, have fun :)

jameswynn:
This implementation allows for some pretty weird substitutions too. You could theoretically use it to smash multiple files together into the final config. I don't think its advisable, but you definitely can. It would probably be a good idea to sort the environment variable keys before doing the substitutions so that it is more deterministic.

shamoon:
Please let me know your thoughts, of course feel free to revert/modify. 

I added caching so we don't loop through them all and a crude check `result.includes('{{')`(went with `includes` because my understanding is its fastest of substring methods) which of course would 'fail' if someone has e.g. a password with `{{` in it or whatever but in that case substitution will still prevent any error.

jameswynn:
I think it looks good. Those changes should definitely make it faster.

JazzFisch:
I must be doing something wrong, but I get an error when I try the following:

```yaml
    - Service:
        href: https://service.domain.local
        container: service
        widget:
            type: service
            url: http://service:1234
            key: {{HOMEPAGE_VAR_SERVICE_API_KEY}}
```

The widget shows:
```
API Error: Invalid data
URL: http://service:1234/api?apikey=***
```

Going back to replace {{...}} with the actual key removes the error.

cmorg3:
Are there any updates on when this will be merged?

jameswynn:
I'm looking into JazzFisch's issue. Once we've resolved that it will be much closer.

jameswynn:
@JazzFisch I just verified on my side that this is working. You may try adding quotes around the substituted value: "{{HOMEPAGE_VAR_SERVICE_API_KEY}}"

JazzFisch:
> @JazzFisch I just verified on my side that this is working. You may try adding quotes around the substituted value: "{{HOMEPAGE_VAR_SERVICE_API_KEY}}"

Argh.  Problem exists between keyboard and chair.  Turns out that if one doesn't check out the PR, one can't test the changes.  Everything here looks great to me.

cmorg3:
I want to follow up on this... How does this implementation work with a .env passed to the container?

I have a .env that contains my secrets. It follows this format: 
```
RADARR_KEY=APIKEYAPIKEYAPIKEYAPIKEYAPIKEY
```
Then my service.yaml includes this however its still not recognizing it....
```
- Media Management:
    - Radarr:
        href: http://radarr.local.example.net/
        icon: radarr.png
        description: Movie Management
        server: my-remote-docker
        container: radarr
        widget:
            type: radarr
            url: http:http://radarr.local.example.net
            key: {{HOMEPAGE_VAR_RADARR_KEY}}
```
Any suggestions or pointers to the correct place for this is very welcome.

shamoon:
> * Only environment variables starting with HOMEPAGE_VAR_ and HOMEPAGE_FILE_ are supported

Also this is only available in `main` at the moment but will be in `latest` after the next release.

PR #1044, opened 2023-02-23
Migrated helm chart from jameswynn/helm-charts

jameswynn:
## Proposed change

Moving my unofficial helm chart to the official repo. In addition to merging this, a new `gh-pages` branch needs to be created first which is where the resulting chart will be published to.

This is based on the official docs: https://helm.sh/docs/howto/chart_releaser_action/

## Type of change

- [ ] New service widget
- [X] Other (please explain)

## Checklist:

### DIFF
--- a/.github/workflows/chart-publish.yml
+++ b/.github/workflows/chart-publish.yml
@@ -1,0 +1,25 @@
+name: Chart Publish
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.1.0
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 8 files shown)


shamoon:
I created the gh-pages branch. Am I correct in thinking this is the same chart that you're using in your repo? Because otherwise I don't have much to add in terms of testing but if we know it works over there then this is pretty straightforward. I've had some trouble with the chart-releaser action in another project but seems simpler here.

jameswynn:
Yeah, this is the exact same chart. All I did was change the urls, readme, and version number. I set the chart version back to 0.1.0 until we get a couple of test builds out of the way.

jameswynn:
Looks like there is an issue. I'll investigate it.

shamoon:
https://github.com/benphelps/homepage/actions/runs/4263538743/jobs/7420407285#step:4:36 :)

jameswynn:
Fix in #1047

shamoon:
@jameswynn I mistakenly thought that the charts-releaser action only published changes to that branch, but it also creates a release package? This of course breaks the software, see #1049 .

I'll revert this while we investigate but perhaps this does need its own repo? Which we can of course setup.

Haha, second time having issues with this one, from now on will just keep the heml charts separated. Thanks as always for your contributions

eanmclaughlin:
I was just writing a reply on the bug thread then saw that you mentioned it here before I posted so I'll put my reply here:

@shamoon regarding the naming of the helm action's releases, looks like we'd need to configure `--release-name-template` (or `CR_RELEASE_NAME_TEMPLATE`):

```
      --release-name-template string   Go template for computing release names,
      using chart metadata (default "{{ .Name }}-{{ .Version }}")
```

([CR configuration docs page](https://github.com/helm/chart-releaser#create-the-repository-index-from-github-releases))

I'm not sure how having two actions (the existing docker build/publish and the new helm publish) each creating a release will work out though (as you've just mentioned).

PR #1067, opened 2023-02-27
Minecraft server status widget

### LINKED ISSUE #1042: [Feature Request] Minecraft server status
### Description

This would be a widget that sends a request to the following URL:

https://mcapi.us/server/status?ip={IP_ADDRESS}&port={PORT}. 

This returns a json indicating if the sever is online, how many players are online and what the IP/name of the server is.

This would be similar to the smart app on the fenrus dashboard, I have a link to their code below:
https://github.com/revenz/Fenrus/tree/master/apps/Smart/Minecraft

### Other

I got quite a bit of a busy upcoming weeks, but if I find the time I could do it.

Sairenity:
## Proposed change

Adds a minecraft server status widget
image

Closes #1042 

## Type of change

- [x] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/45

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "js-yaml": "^4.1.0",
     "json-rpc-2.0": "^1.4.1",
     "memory-cache": "^0.2.0",
+    "minecraft-ping-js": "^1.0.2",
     "next": "^12.3.1",
     "next-i18next": "^12.0.1",
     "node-os-utils": "^1.3.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,6 +22,7 @@ specifiers:
   js-yaml: ^4.1.0
   json-rpc-2.0: ^1.4.1
   memory-cache: ^0.2.0
+  minecraft-ping-js: ^1.0.2
   next: ^12.3.1
   next-i18next: ^12.0.1
   node-os-utils: ^1.3.7
@@ -53,6 +54,7 @@ dependencies:
   js-yaml: 4.1.0
   json-rpc-2.0: 1.4.1
   memory-cache: 0.2.0
+  minecraft-ping-js: 1.0.2
   next: 12.3.1_biqbaboplfbrettd7655fr4n2y
   next-i18next: 12.0.1_azq6kxkn3od7qdylwkyksrwopy
   node-os-utils: 1.3.7
@@ -2152,6 +2154,13 @@ packages:
     engines: {node: '>=6'}
     dev: false
 
+  /minecraft-ping-js/1.0.2:
+    resolution: {integrity: sha512-h9QYG2n+fBKgp520tXBwR354XRzR/w5wXe8CJCmxKm6jbLpAoLODM8Nj5+ssuIVQF8rtxkAnjwv7PH+7ehFzQQ==}
+    dependencies:
+      node-int64: 0.4.0
+      varint: 6.0.0
+    dev: false
+
   /mini-svg-data-uri/1.4.4:
     resolution: {integrity: sha512-r9deDe9p5FJUPZAk3A59wGH7Ii9YrjjWw0jmw/liSbHl2CHiyXj6FcDXDu2K3TjVAXqiJdaw3xxwlZZr9E6nHg==}
     hasBin: true
@@ -2286,6 +2295,10 @@ packages:
     transitivePeerDependencies:
       - '@babel/core'
       - babel-plugin-macros
+    dev: false
+
+  /node-int64/0.4.0:
+    resolution: {integrity: sha512-O5lz91xSOeoXP6DulyHfllpq+Eg00MWitZIbtPfoSEvqIHdl5gfcY6hYzDWnj0qD5tz52PI08u9qUvSVeUBeHw==}
     dev: false
 
   /node-os-utils/1.3.7:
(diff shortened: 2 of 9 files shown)


Sairenity:
This is the first time I'm contributing to an OSS project, so if you have any feedback at all, please let me know :)

shamoon:
Thanks for the PR, welcome to FOSS :)

So just to clarify this widget hits a third-party service which is what actually pings the server? Is there a reason you’re not bypassing that and hitting the server directly? And did you check the TOS of that service to see if this would be ok?

Sairenity:
Hi shamoon, thanks for the warm welcome :)

> So just to clarify this widget hits a third-party service which is what actually pings the server? Is there a reason you’re not bypassing that and hitting the server directly?

Yep, exactly. The reason I'm not hitting the server directly is because minecraft itself sadly doesn't provide a REST api to query server status information. [The protocol minecraft uses](https://wiki.vg/Query) is a bit more... involved, and if I don't absolutely have to re-implement it, I'd rather use a proxy to get the infos 😅 

> And did you check the TOS of that service to see if this would be ok?

The service I used doesn't have an explicit TOS, and even after some digging, the only thing I'm able to find which implicitly grants permission is the [API Docs](https://api.mcsrvstat.us/)

If you don't feel comfortable with this service, which I would totally understand, I could go hunt for another with more acceptable explicit TOS?

Kind regards, 
Sairenity

Sairenity:
of course, I could also pull in something like [minecraft-ping-js](https://www.npmjs.com/package/minecraft-ping-js), and write a proxy in src/utils/proxy/handlers... both would work for me.

shamoon:
Ok thanks. As long as they don’t explicitly say this isn’t allowed it should be ok. The custom proxy thing can get messy. it’s probably worth noting in the docs how this works and your widget.js can probably just hard-code the URL part no? Or is the software that runs that service available to self-host (in which case the main url could be different I suppose)

Thanks so much, we’ll take a look at the actual code soon.

Sairenity:
> Ok thanks. As long as they don’t explicitly say this isn’t allowed it should be ok. 

great to hear :)

> The custom proxy thing can get messy.

I cobbled together a proxy that actually works... I'm now quite tempted to just use this instead. [Here's the gist to the minecraft proxy handler](https://gist.github.com/Sairenity/0c69dd0ab76a1ac52bcaf90e4f7270d8)

> your widget.js can probably just hard-code the URL part no? Or is the software that runs that service available to self-host (in which case the main url could be different I suppose)

yeah, that's fair criticism. I left the URL configurable, just in case somebody wanted to build a compatible drop-in replacement.

shamoon (changes requested):
Hey, we talked about this and even though we usually like to avoid custom proxies we agreed the version where uses can connect to their own servers without relying on this third party service would be preferable. Let us know if you need help working this out, shouldn't be too bad.

Thanks for your contribution

shamoon:
Nice thanks! Is this ready to test?

Sairenity:
heya, yep this is now ready from my side. mind taking a look?

Sairenity:
fwiw, here's a sample `services.yaml` to quickly check if everything is behaving well.

```yaml
- Minecraft Servers:
    - Hypixel:
        description: mc.hypixel.net
        icon: https://api.mcsrvstat.us/icon/mc.hypixel.net
        widget:
          type: minecraft
          domain: mc.hypixel.net
    - Inexistant:
        description: some inexistant server
        icon: https://api.mcsrvstat.us/icon/does.not.exist
        widget:
          type: minecraft
          domain: does.not.exist
```

shamoon:
Cool, is this also compatible with something like https://github.com/itzg/docker-minecraft-server that a user would use for self-hosted?

Sairenity:
> Cool, is this also compatible with something like https://github.com/itzg/docker-minecraft-server that a user would use for self-hosted?

Yep, it totally should be! I've verified it's working for a locally-hosted server by plugging in the IP directly.
I'll test the docker scenario a bit later, because this is also interesting for the documentation. Being able to directly reference the docker container like such would be quite neat:

```yaml
- Minecraft Servers:
    - Local Docker Minecraft Server:
        widget:
          type: minecraft
          domain: docker-container-name-here
```

shamoon:
Awesome, this works great with your examples and self-hosted. The thing you mentioned above about referencing containers should already work, thats a docker thing

Self-hosted:
Screenshot 2023-02-28 at 8 50 02 AM

Others:
Screenshot 2023-02-28 at 8 35 43 AM
Screenshot 2023-02-28 at 8 35 47 AM

See my changes, obviously one thing is I moved the proxy since it won't be used by other widgets. I also changed the URL stuff to use the `url` key which is standard here. Your examples above will need to have: `url: http://mc.hypixel.net` as a result, but for self-hosted it'd be `url: http://localhost:25565` or similar.

Please let me know any thoughts, feel free to tweak more of course. Docs will need to be updated per all this.

shamoon:
> _technically,_ what we're using to try connect to a minecraft server isn't an URL, since we're not using http or https, but that's just me splitting hairs 🤓

Hmm, I was going to go and look if it does use http or not, wasn’t sure. Yea, I agree it bugs me too :) if it’s not technically correct but wondering if that outweighs the benefit of using the `url` key that is otherwise pretty standard

21st-centuryman:
@Sairenity Just wanna say great job on this. Welcome to FOSS.

shamoon:
Lets go with `url: udp://mc.hypixel.net` =)

Sairenity:
> Lets go with `url: udp://mc.hypixel.net` =)
that's a great suggestion!

Thank you for all the help, also with the docs.

mgrimace:
The minecraft ping / proxy handler used here seems to reference service port 25565, which appears to work for java edition, but does not work (at least for me) for a self-hosted bedrock server (e.g., xbox compatible) which instead uses port 19132 as the default port. For example, I use 
udp://[serverIP]:19132 but it results in: 

Screenshot 2023-06-19 at 8 47 04 PM

Despite the server being online/connectable on PC/xbox, etc.
Any idea on the widget side to add bedrock compatibility?

shamoon:
There’s an open FR for that https://github.com/benphelps/homepage/discussions/1179

Sairenity:
I considered bedrock out-of-scope initially.

I could take a look how easy it's API is to integrate 👍

SX86:
Any progress on getting it compatible with Bedrock?
I thought I was going crazy until I saw that last comment from Jun 20 :)

shamoon:
https://github.com/gethomepage/homepage/discussions/1179#discussioncomment-6647755

PR #1068, opened 2023-02-27
Add Quicklaunch feature to jump to a url

Aesop7:
## Proposed change

I wanted to be able to (quickly) open arbitrary URLs from Homepage.  This change adds some functionality to the quicklaunch feature so that users can input a url, ~~and then press SHIFT+ENTER to open the url.  The url will open in the window or tab corresponding to what the user has configured).  I've also added an indicator for this functionality so that a user knows when they are going to use the quicklaunch for a url.

Pressing enter while the shift button is pressed will open the url in the appropriate tab/window.~~

Edit: now
Screenshot 2023-02-27 at 8 08 31 PM

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -286,7 +286,9 @@
         "bookmark": "Bookmark",
         "service": "Service",
         "search": "Search",
-        "custom": "Custom"
+        "custom": "Custom",
+        "visit": "Visit",
+        "url": "URL"
     },
     "wmo": {
         "0-day": "Sunny",
--- a/src/components/quicklaunch.jsx
+++ b/src/components/quicklaunch.jsx
@@ -6,14 +6,16 @@ import ResolvedIcon from "./resolvedicon";
 
 import { SettingsContext } from "utils/contexts/settings";
 
-export default function QuickLaunch({servicesAndBookmarks, searchString, setSearchString, isOpen, close, searchDescriptions, searchProvider}) {
+export default function QuickLaunch({servicesAndBookmarks, searchString, setSearchString, isOpen, close, searchProvider}) {
   const { t } = useTranslation();
   const { settings } = useContext(SettingsContext);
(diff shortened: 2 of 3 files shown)


shamoon (changes requested):
Thank you for the PR. I could see the functionality being useful but I have some issues with the implementation (for now we can put code issues to the side). 

Principally, I think this is not easily discoverable and too different than other quicklaunch functions. I would propose instead of the shift key you change this to add an item (and you can make it the currently-selected one) to the list of options when the string is ‘detected’ as a url. This would be much more obvious and even require fewer keystrokes. I would also add a setting to disable this.

Let me know if that makes sense or if you’d like any help implementing this. Thanks again.

Aesop7:
I like it a lot.  Something like this right?
image

I just threw something together that matches based on whether or not the string has a "dot" in it .  Is there a better solution?

shamoon:
Yes, I would validate the string as a URL using [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL)

shamoon:
Im happy to take a crack at it if you'd like, or I can leave you to it =)

Aesop7:
That sounds good to me!  But first, there are a couple of use-cases with this that I was thinking about  (all of them are trying to mimic the modern way chrome/firefox handle URLs in their url/search bars)
- A user might type "http://mail.google.com"; and if they do, the "http" should be respected
- If a user types "mail.google.com", that should also "work" and hitting enter should bring them to "https://mail.google.com"

Do you think that's workable?

shamoon:
Take a look at the last change, would still need to add the setting

Aesop7:
:partying_face:  That'll do it!

Aesop7:
I can work on adding a setting tomorrow.  It's getting late for me.  I'm LOVING this functionality though

shamoon:
Cool 😎

I can probably finish it up later :)

Aesop7:
Oh one minor thing I just thought of: would it make sense to look for
“localhost:” as well?  Cheers!

shamoon (approved):

PR #1069, opened 2023-02-28
Refactor Healthchecks widget API host to be parametrized

rommelkott:
## Proposed change

I have refactored the healthchecks widget so users are able to use locally hosted instances of healthchecks. This was done by simply replacing the hardcoded host with a URL parameter. 

**This technically breaks the previous version of the widget**, as users missing the URL parameter in the config will cause the widget to not work.  What would you recommend?

Request
https://github.com/benphelps/homepage/pull/1023#issuecomment-1446052950

## Type of change

- [ ] New service widget
- [x] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/47

### DIFF
--- a/src/widgets/healthchecks/widget.js
+++ b/src/widgets/healthchecks/widget.js
@@ -1,7 +1,7 @@
 import credentialedProxyHandler from "utils/proxy/handlers/credentialed";
 
 const widget = {
-  api: "https://healthchecks.io/api/v2/{endpoint}/{uuid}",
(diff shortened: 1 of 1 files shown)


shamoon:
Yea there isn't an easy way to do this with some kind of fallback, I think better to do it soon. Will note in the release notes.

PR #1023, opened 2023-02-19
Add Health checks widget

rommelkott:
## Proposed change

First time open source contributor.

I implemented a widget for [healthchecks.io](https://healthchecks.io/). It shows the current status of a project. (online, offline, etc)

This closes https://github.com/benphelps/homepage/issues/965

image
image
image

### Widget configuration
    widget:
        type: healthchecks
        key: <YOUR_API_KEY>
        uuid: <YOUR_CHECK_UUID>

Let me know if you need any changes done. 👍

## Type of change

- [x] New service widget
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/39
- [x] If applicable, I have checked that all tests pass with e.g. `pnpm lint`.
- [x] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #916, opened 2023-02-02
Feature: Uptime kuma widget

shamoon:
Big thanks to @chain710 for kicking this off, using the status page api is a great solution.

```yaml
widget:
  type: uptimekuma
  url: http://uptime-kuma-host:port
  slug: statuspageslug
```

Screenshot 2023-02-02 at 1 08 59 AM
Screenshot 2023-02-02 at 12 51 11 AM

Closes #123

PR #1070, opened 2023-03-01
New Widget: Per Monitor Uptime Kuma Monitoring

LAX18:
## Proposed change
image
Adds a new uptime kuma widget that allows you to set specific monitors on a service by service basis. Similar to #1023 but with support for UptimeKuma

This is a continuation of #916, where this widget has a different approach of accessing a monitor's status without an API.

## Type of change

- [x] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/49

This is my first contribution to this project so if any additional information or work is needed, please let me know! I am also curious if it is possible in a service widget to add the 'healthy' banner (or the status dot from earlier versions of homepage) as a means to show status as opposed to a larger bar on the bottom.

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -469,6 +469,11 @@
         "incident": "Incident",
         "m": "m"
     },
+    "uptimekumamonitor": {
+        "up": "Up",
+        "down": "Down",
+        "maintenance": "Maintenance"
+    },
     "komga": {
         "libraries": "Libraries",
         "series": "Series",
--- a/src/widgets/components.js
+++ b/src/widgets/components.js
@@ -74,6 +74,7 @@ const components = {
   unifi: dynamic(() => import("./unifi/component")),
   unmanic: dynamic(() => import("./unmanic/component")),
   uptimekuma: dynamic(() => import("./uptimekuma/component")),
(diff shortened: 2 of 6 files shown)


shamoon:
Thank you for the PR. This is interesting but Im not sure we can accept it as-is. Some notes:

- It's one thing to get the overall stats from an uptimekuma instance (i.e. the current widget) but just checking if a site is up or down can already be done with [ping](https://gethomepage.dev/en/configs/services/#ping)
- The code here is parsing an svg which is why you were forced to create your own proxy. But in fact this data is already available and already being retrieved by the current uptimekuma endpoint via the actual API.
- To answer your question 'healthy' is used for docker integrations

Again, this:
Screenshot 2023-03-01 at 12 23 19 AM
can already be achieved with
```yaml
     - My Second Service:
          description: Homepage is the best
          href: https://wnyc.org
          ping: https://wnyc.org
```

Im going to close this for now but welcome other thoughts from you or anyone else of course. Thank you for the contribution.

PR #1071, opened 2023-03-01
Added Ghostfolio widget

math625f:
## Proposed change

This change will add a new [Ghostfolio](https://github.com/ghostfolio/ghostfolio) widget. 

I'm aware that 4 requests is a lot, and I'm hoping that perhaps there's a solution to this issue that I am not seeing. I've been unable to find documentation for the Ghostfolio API, and testing with combined endpoints (e.g. `range=max&range=ytd` or `range=max,ytd`) proved unsuccessful. Perhaps it'd be possible to calculate all four values from the `max` endpoint (using the `chart` data with the `value` field), but I have yet to explore this option - I wanted to hear your thought first.

Authentication for Ghostfolio is rather cumbersome. It involves using a "Secret Token" (given on user creation time), to obtain a (JWT) Bearer token that can then be used for further requests. Since this is not something already present as a proxy in the repository, currently the code assumes that the user obtains this bearer token by themselves (can be done using the browser, since it's just a get request). The downside of this, is that the token expires after 6 months. 

The "Secret Token" lasts forever, so if I were to implement a custom proxy for just this functionality, the user would never have to change their configuration after the initial setup.

Note that in the `component.jsx` file there are some commented out blocks. Initially I wanted to utilise the `common.percent` translation to display the percentages, but when using that, the percentages seem to be automatically rounded (to zero decimals). For this kind of information, I believe it is better to have at least two decimals, which is why I opted for this custom implementation. What are your thoughts on this issue?

## Type of change

- [X] New service widget
- [ ] Other (please explain)

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/51

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -535,5 +535,11 @@
         "targets_up": "Targets Up",
         "targets_down": "Targets Down",
         "targets_total": "Total Targets"
+    },
+    "ghostfolio": {
+        "gross_percent_today": "Today",
+        "gross_percent_ytd": "This year",
+        "gross_percent_1y": "One year",
+        "gross_percent_max": "All time"
     }
 }
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -538,7 +538,6 @@
     },
     "ghostfolio": {
         "gross_percent_today": "Today",
-        "gross_percent_ytd": "This year",
         "gross_percent_1y": "One year",
         "gross_percent_max": "All time"
     }
--- a/src/utils/proxy/handlers/credentialed.js
+++ b/src/utils/proxy/handlers/credentialed.js
@@ -47,6 +47,8 @@ export default async function credentialedProxyHandler(req, res, map) {
         headers.Authorization = `Bearer ${widget.key}`;
       } else if (widget.type === "pterodactyl") {
         headers.Authorization = `Bearer ${widget.key}`;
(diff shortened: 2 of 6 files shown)


shamoon:
Thanks for the PR, this worked out-of-the-box 👍

Please take a look at my changes, some notes:

- Only functional change here was to remove the YTD which reduces the API calls to 3. Hope that's OK with you but I see it as a compromise and yea API documentation is thin over there, no obvious way to get them at once.
- We don't allow auto-refreshing at the moment
- In general I simplified your code and tried to make it more DRY, for example extracting the code to get the percent as a function. (and btw yes, you can use decimals in i18next with `maximumFractionDigits`)
- Im ok with the JWT token rather than a custom proxy.

Of course welcome any thoughts / changes to the above.

math625f:
Thanks, your changes make a lot of sense.

This is my first time looking at your codebase, so I'm not surprised there are some things I have overlooked.

- This works for me, I removed the unused key from the translation file as well - I'm unsure if this is something that's generated automatically.
- I misunderstood the functionality I think. I thought the `refreshInterval` would allow me to "limit" the number of requests.
- Very nice, that makes a lot more sense.
- JWT it is; once every 6 months is not that bad anyways.

PR #1093, opened 2023-03-09
Fix: Use human readable units for storage in Immich widget

Zlendy:
## Proposed change
Immich updated its API endpoint to return raw bytes instead of human readable data. This PR adapts the existing code to take this into account.
| Before | After |
| --- | --- |
| image | image |

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:


shamoon:
Thanks for the PR, homepage (/ nextjs) already implements conversion like this, see the updated change and lmk if you see any issues.

Issue #264, opened 2022-09-24
[Feature Request] Jellyfin Widget to show number of movies, songs, indexing status


chall3ng3r:
It'd be useful if the Jellyfin widget displays number of movies and songs in library, indexing / scan status.

jagadam97:
https://github.com/benphelps/homepage/pull/1096

PR #1096, opened 2023-03-11
#264 : adding count for jellyfin media

jagadam97:
Added code for the above issue but was unable to test if its working as expected


shamoon:
Sorry just clarifying youre submitting a PR but haven't tested it?

jagadam97:
> Sorry just clarifying you're submitting a PR but haven't tested it?

Yes. I use code server to wrk from my company laptop as my personal laptop is unavalible for the moment and code server has wierd proxying to test live builds and it was not working properly... Will test it as soon as i can and update the same

PR #1112, opened 2023-03-15
Feature: Add Traefik IngressRoute support for Kubernetes

### LINKED ISSUE #906: [Feature Request] Traefik IngressRoute annotations readout
### Description

Please add the readout of the annotations of the Traefik IngressRoutes

### Other

_No response_

MrLemur:
## Proposed change

This adds support for Traefik IngressRoute CRDs as well as the traditional Kubernetes Ingress manifest.

A check for Traefik IngressRoute objects is carried out and if found, will then check for the `href` attribute set in annotations. If found, it will push the IngressRoute object into the existing `ingressList.items` array.

The reason for the `href` requirement is the matching rule that IngressRoute can be tricky to parse, and it seems better to rely on the user to set it. If it is not set, it will not be added to the `ingressList` for later processing.

A log error has been added for failing to find IngressRoute objects, but this could be only set via an additional Homepage setting.

Closes #906

Documentation PR: https://github.com/benphelps/homepage-docs/pull/58

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:


shamoon:
Interesting thanks. We’ll take a look when we have a moment

MrLemur:
> Interesting thanks. We’ll take a look when we have a moment

No worries, please let me know if you would want this to be locked behind a user setting.

PR #1121, opened 2023-03-20
Add support of NC-Token to Nextcloud widget

### LINKED ISSUE #1122: [Feature Request] Add support for Nextcloud's NC-Token
### Description

Would be great to not be forced to use user+password for Nextcloud widget, but go with NC-Token instead:

image

### Other

_No response_

shalak:
## Proposed change

Adds support for Nextcloud's NC-Token authentication. This is safer than user+pass, as the only purpose of this token is to access the serverinfo endpoint, AFAIK.
image

Closes #1122

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

https://github.com/benphelps/homepage-docs/pull/60

### DIFF
--- a/src/utils/proxy/handlers/credentialed.js
+++ b/src/utils/proxy/handlers/credentialed.js
@@ -49,11 +49,9 @@ export default async function credentialedProxyHandler(req, res, map) {
       } else if (widget.type === "miniflux") {
         headers["X-Auth-Token"] = `${widget.key}`;
       } else if (widget.type === "nextcloud") {
-        if ('key' in widget) {
-          logger.debug("Setting nextcloud to use NC-Token");
+        if (widget.key) {
           headers["NC-Token"] = `${widget.key}`;
         } else {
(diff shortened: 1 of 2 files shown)


src/utils/proxy/handlers/credentialed.js
```diff
@@ -48,6 +48,14 @@ export default async function credentialedProxyHandler(req, res, map) {
         headers.Authorization = `Token ${widget.key}`;
       } else if (widget.type === "miniflux") {
         headers["X-Auth-Token"] = `${widget.key}`;
+      } else if (widget.type === "nextcloud") {
```
shalak, line 51: It would be a good idea to refactor the whole approach, so we could configure the `credentialedProxyHandler` behavior from `widget.js` files, that way it would keep more of a SOLID approach. However this is outside of the scope of this PR

shamoon:
Cheers

Issue #1136, opened 2023-03-22
[Feature Request] Possibility for Local / Remote Scope


norsemanGrey:
### Description

This is a really awesome dashboard for my self-hosted services. Great work!

One thing I am missing, which would be great if you could have a look at, is the possibility to have a local and remote scope (if that is the correct terminology).

I have a lot of services which are only accessible locally on my LAN using a local address space. However, some services are also available externally via reverse proxy on a FQDN address. Would it be possible do define both local and external addresses for a service / widget and then show/hide the tiles based on where homepage is accessed from? So if I access homepage internally (e.g. using `http://homepage.at.home` ) only tiles with a local address/url (or both local and external) are shown, and if accessing homepage from the outside (e.g. using `https://homepage.external-domain.com`) only tiles with  an external address/url (or both local and external) are shown.

jramsgz:
I am also interested in this feature. In addition to what @norsemanGrey said, we should consider hiding or preventing access to system information such as CPU, RAM, and Docker stats since exposing that information to the internet could pose a security risk. The switch between LAN and WAN could be detected by checking if the IP accessing the service is within the user-defined LAN address space or not. If connected through a trusted reverse proxy, the homepage should look for the real client IP in the headers.

PR #1160, opened 2023-03-29
Initial scopes support

jramsgz:
## Proposed change

Here is a proposal (partially tracked by issue #1136) for implementing scopes. At present, it adds a way to identify whether a user is accessing from a local or remote scope, so that we can restrict or hide certain features such as system information, widgets, or services in the future. This proposal introduces two new configuration entries: "trustedproxies" and "localscope".
The "trustedproxies" entry contains an array of IP addresses or IP ranges. If a connection is made from any of these addresses, we assume that it is safe to extract the real user IP from the "x-real-ip" header.
The "localscope" entry contains an array of IP addresses or IP ranges. If a connection is made from any of these addresses, we consider the user to be in the local scope.

Example:
```
trustedproxies: [127.0.0.1, 192.168.1.0/24]
localscope: [192.168.1.0/24]
```

Please let me know if you have any feedback on this proposal so that we can move forward into supporting scopes.

Related to #1136 

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "dockerode": "^3.3.4",
     "follow-redirects": "^1.15.2",
     "i18next": "^21.9.2",
+    "ipaddr.js": "^2.0.1",
     "js-yaml": "^4.1.0",
     "json-rpc-2.0": "^1.4.1",
     "memory-cache": "^0.2.0",
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@headlessui/react": "^1.7.2",
     "@kubernetes/client-node": "^0.17.1",
+    "@supercharge/request-ip": "^1.2.0",
     "classnames": "^2.3.2",
     "compare-versions": "^5.0.1",
     "dockerode": "^3.3.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,6 +25,9 @@ dependencies:
   i18next:
     specifier: ^21.9.2
     version: 21.10.0
+  ipaddr.js:
+    specifier: ^2.0.1
+    version: 2.0.1
   js-yaml:
     specifier: ^4.1.0
     version: 4.1.0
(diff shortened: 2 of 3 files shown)


shamoon:
Hmm, we'll have to discuss. At the moment Im not sure if this is something we really want to incorporate. 

Without getting into an actual code review, one thing I have to wonder is whether there are existing [tried-and-tested or native] libraries for these things. 

Also:

- https://nodejs.org/dist/latest/docs/api/http.html#http_request_connection is deprecated.
- What about `request.headers['x-forwarded-for']`

jramsgz:
Thank you for your response. I think it would be valuable to wait and see what the community thinks before making a final decision. In the meantime, I'll continue to gather feedback and work on refining the proposal

I attempted to avoid introducing dependencies, however, upon researching existing options for retrieving the user's IP address, I discovered "supercharge/request-ip". We still need to check if we can trust headers though, for that, instead of performing CIDR comparison on our own, we could leverage another library for comparing IP ranges. If you would like me to, I can look into this further.

Regarding the `x-forwarded-for` header, it contains a list of IPs that forwarded the request, so to obtain the real user IP, we only need to consider the first IP on the list. As we only require the client's IP address, it seemed unnecessary to take that header into account. But now that you mention it, adding it might be helpful for some cases were x-real-ip is not present just like the library I mentioned earlier does.

Lastly, I will replace the `connection` with `socket` to fix the deprecation issue. Thank you for bringing this to my attention.

shamoon:
Yes, `supercharge/request-ip` looks good and as noted, it has more comprehensive logic than this implementation including checking all possible headers, getting the IP from `socket` etc. Seems like that will take care of a large portion of the current code.

> I think it would be valuable to wait and see what the community thinks before making a final decision.

On the one hand, yes of course always valuable to know. On the other hand, if FOSS has taught me anything it's that people who make FRs and PRs aren't always the same ones as those who actually maintain a project, so we always need to balance things for longevity and keeping the scope narrow helps with that. (And I know everyone thinks there idea "won't require much maintenance" but of course that doesn't always turn out to be the case 😉)

My worry in general with something like this is that it will get a bit messy (this util is one thing but actually integrating it into the app, etc.). Again, we'll discuss the feature in general at some point. Appreciate your interest in contributing either way.

shamoon:
@jramsgz So at this point I have to tell you I don't think we would support this PR. It's one thing for this PoC to do what it should but as I mentioned, when it comes to integrating this we feel things will get kinda out of control fast especially for what I think is sort of a niche feature. You would need checks everywhere, including all the frontend rendering code and every backend API call.

Ultimately this feels out of scope for homepage (pardon the pun). Thank you again for your effort and interest.

jramsgz:
@shamoon Thank you for considering my PR and explaining your concerns about its impact on homepage's maintainability. I understand that the proposed feature may not align with the project's scope and requirements.

Given the situation, I plan to maintain a fork with scopes support in my branch, which should be more than enough for my specific needs (here is my [docker image](https://github.com/jramsgz/homepage/pkgs/container/homepage)). However, I would be happy to collaborate with anyone who is interested in expanding and improving this feature in the future.

Once again, thank you for your feedback and for your efforts in maintaining homepage.

PR #1187, opened 2023-04-05
True fix for settings revert after container creation

### LINKED ISSUE #576: [Bug] Layout settings not picked up after update

shamoon:
## Proposed change

Perhaps just needed some time to meditate on this but I think this is a true solution here. Using similar logic from #963, this PR inserts a timestamp into the hash which gets set by docker-entrypoint.sh i.e. on container start. The only real downside of this that I can think of is that revalidate will happen even when the container is _restarted_ not just _re-created_, but actually... that seems totally acceptable =)

There's another (not really directly related) change in here just to make settings.yaml creation consistent with others

Fixes #576 , like for reals

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ COPY . .
 
 SHELL ["/bin/ash", "-xeo", "pipefail", "-c"]
 RUN npm run telemetry \
- && mkdir config && echo '---' > config/settings.yaml \
+ && mkdir config \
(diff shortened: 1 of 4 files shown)


shamoon:
Merging so people can test this in `main`

PR #1350, opened 2023-04-11
Added homeassistant widget

hen-ter:
## Proposed change

This change will add a new [Home Assistant](https://www.home-assistant.io/) widget.

There is some discussion about it and further information [here](https://github.com/benphelps/homepage/discussions/1246).

The widget is a bit unusual as it allows user defined information from Home Assistant to be displayed. It defaults however to displaying some basic information without custom configuration by the user.

There is a custom proxy handler but only one useWidgetAPI call is needed in component.jsx. The component is then dynamically created with the information returned.

*This is my first ever PR so I hope I didn't mess anything up.*
*This is also my first ever JS code so I more than welcome any changes/improvements/comments.*

## Type of change

- [x] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: [Home Assistant widget homepage-docs#72](https://github.com/benphelps/homepage-docs/pull/72)

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -571,5 +571,10 @@
         "books": "Books",
         "podcastsDuration": "Duration",
         "booksDuration": "Duration"
+    },
+    "homeassistant": {
+        "people_home": "People Home",
+        "lights_on": "Lights On",
+        "switches_on": "Switches On"
     }
 }
--- a/src/components/services/widget/container.jsx
+++ b/src/components/services/widget/container.jsx
@@ -14,7 +14,7 @@ export default function Container({ error = false, children, service }) {
     // fields: [ "resources.cpu", "resources.mem", "field"]
     // or even
     // fields: [ "resources.cpu", "widget_type.field" ]
-    visibleChildren = children.filter(child => fields.some(field => {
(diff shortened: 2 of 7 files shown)


shamoon:
Honestly, this is pretty great (and again gives me pangs about switching to HASS but thats my problem). I think there are some minor quibbles with the code but overall really nice work.

Screenshot 2023-04-12 at 12 26 11 AM

hen-ter:
Thanks, really appreciate that.
Who knows why I didn't just put the key prop in the element myself instead of doing the clumsy forEach and push to avoid the lint error...
I think the docs should mention the limit of 4 custom queries.

stevietv:
@shamoon - is it possible that custom states or templates don't work via docker labels and only via a yaml config?
I've tried many different ways of adding it and all result in an API error:
`API Error: Unexpected token 'I', "Internal S"... is not valid JSON`

here is how I've currently tried to add it as a label:
` - homepage.widget.custom.state=sensor.thermostat_current_temperature`

PR #1366, opened 2023-04-16
Add optional boxed styling and error component to information widgets

denispapec:
## Proposed change

This change adds a new optional setting for all available information widgets: `styleBoxed: true`, which allows us to have more stylistic view of the header by groups single widget under a common boxed style. This PR also includes a separate error component for information widgets that unifies handling of error output. 

Screenshot_20230416_003142
Screenshot_20230416_003053

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/75

### DIFF
--- a/src/components/widgets/datetime/datetime.jsx
+++ b/src/components/widgets/datetime/datetime.jsx
@@ -1,5 +1,6 @@
 import { useState, useEffect } from "react";
 import { useTranslation } from "next-i18next";
+import classNames from "classnames";
 
 const textSizes = {
   "4xl": "text-4xl",
@@ -17,7 +18,7 @@ export default function DateTime({ options }) {
   const { i18n } = useTranslation();
   const [date, setDate] = useState("");
   const dateLocale = locale ?? i18n.language;
-  
+
   useEffect(() => {
     const dateFormat = new Intl.DateTimeFormat(dateLocale, { ...format });
     const interval = setInterval(() => {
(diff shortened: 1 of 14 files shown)


shamoon:
Thanks for the PR.

I'm not sure how I feel about this aesthetically. Obviously that's subjective but I find the boxes just emphasizes all of the unequal gutters between columns and rows. If you want to keep the error change perhaps the rest of it should move to a discussion. I'm sure some folks will like it but some feedback before PR would be valuable

benphelps:
I'm not against adding a new style option for the widgets, but this highlights a problem that should be fixed first.

Header widgets need to be refactored to use wrapping components like the rest of the service widgets do.  Applying a style to the widgets shouldn't require changing every widget individually.  I think that should be taken care of first, so a change like this could be done in only a few files.

The option to enable this styling should live in the `settings.yaml` file where the other styling options are, not on each individual widget.

shamoon:
Please re-open if you'd like to address the above.

denispapec:
Hi, can you reopen the PR? It doesn't seem that I have rights?

I've pushed further changes to my branch and rebased the code to (almost) latest main branch, including the enhanced glances added in https://github.com/benphelps/homepage/pull/1534 in mean time.

I agree about refactoring the widgets btw, I did already error component at that time as I've had issue to have unified widget style with error even at that point. The option now lives in `settings.yaml`.

My latest changes include all those widget refactored to shared components (to everything what makes sense to have), except for Kubernetes, Longhorn, Unifi widget as they don't follow any shared style (except for the wrapping Container component). Also, as I don't use these 3 technologies, I've faked API response to test. For completeness, here is screenshot with all widgets, having different states:
Screenshot_20230605_230724


And a screenshot of an real life example:
Screenshot_20230605_231148


Thanks for this great homepage!

shamoon:
This can’t be reopened for whatever reason (perhaps your fork changed or whatever) so you can just open a new one

PR #1377, opened 2023-04-18
Add FreshRSS widget

henry40408:
## Proposed change

Closes #1340 (issue)

> Minimize the number of API calls

To retrieve the counts for subscriptions and unread items, it is necessary to make separate API calls for each.

By the way, does Homepage offer any means of storing the exchanged token for FreshRSS in order to minimize authentication API requests, given that FreshRSS requires exchanging tokens with username and password?

## Type of change

- [x] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/76

### DIFF
--- a/src/widgets/freshrss/proxy.js
+++ b/src/widgets/freshrss/proxy.js
@@ -27,7 +27,7 @@ async function login(widget, service) {
 
   try {
     const [, token] = data.toString().split("\n").find(line => line.startsWith("Auth=")).split("=")
-    cache.put(`${sessionTokenCacheKey}.${service}`, token, 30 * 60 * 1000); // 30m
+    cache.put(`${sessionTokenCacheKey}.${service}`, token);
     return { token };
   } catch (e) {
     logger.error("Unable to login to FreshRSS API: %s", e);


shamoon (changes requested):
Thanks for the PR.

There are some changes that need to be made however. Specifically your addition to the credentialed proxy which uses `fetch`:
1. This makes the backend calls no longer proxied, which isn't allowed (requests will come from the browser to the backend, thats only allowed if e.g. calling out to the internet)
2. This won't work in many situations e.g. Docker. If you use http://container:port as your URL (often the case) the client (browser) won't be able to resolve the hostname and the call will fail.
3. This will repeat auth on every call.

I don't know FreshRSS specifically but I would see if theres a way to auth that doesn't require the above. If it does then you will need to (unfortunately because we do like to avoid them) implement a custom proxy. As part of this the auth should be stored in cache. There are quite a few widgets that work similarly, an example sort of like this would be homebridge: https://github.com/benphelps/homepage/blob/main/src/widgets/homebridge/proxy.js which as you can see only calls login if necessary and stores auth in the cache.

Let us know how you get on.

henry40408:
@shamoon Based on the suggestion, I have implemented a custom proxy handler for FreshRSS in commit 651e06aa25dc7a6a0cb1411998053f56e25e42e0.

By the way, since the "login" function is asynchronous in the custom proxy handler for Homebridge, should it be called using the "await" keyword?

https://github.com/benphelps/homepage/blob/4794c8e2ed7cd0b4d7fb285bb8efe997f24dcad2/src/widgets/homebridge/proxy.js#L55

shamoon:
Ha great catch! Thanks I'll take a look when I have a moment

shamoon:
Awesome, this works great out-of-the-box.

Screenshot 2023-04-19 at 8 03 27 AM


Is the token really only valid for 30 minutes? I didnt dig too hard but didnt see where thats documented. NBD but I suppose could reduce the auth calls if its longer-lived.

shamoon:
> Since the salt is generated during installation and remains constant unless the user re-installs the FreshRSS instance, should I remove the TTL parameter from `cache.put` and cache the token until the Homepage server is restarted?

Yes I think so! 👍

henry40408:
The TTL parameter has been removed in 63e38a5fa780ab31f8fa782467a69924d312d5b2.

crispybegs:
> Awesome, this works great out-of-the-box.
> 
> Screenshot 2023-04-19 at 8 03 27 AM
> 
> Is the token really only valid for 30 minutes? I didnt dig too hard but didnt see where thats documented. NBD but I suppose could reduce the auth calls if its longer-lived.

this is great. are we able to use it yet? not sure what the field  / widget syntax is

PR #1383, opened 2023-04-19
Allow setting rpcUrl path for Transmission widget

mcclurec:
## Proposed change

[Transmission allows setting the RPC url path](https://github.com/transmission/transmission/blob/e68c72daa479f81d519fb83219aedc81d1516fb9/docs/Editing-Configuration-Files.md#L128) to something other than the default `/transmission/`. Changing this value can be necessary when doing some types of routing (e.g. `transmission.custom-domain.com`) to get the web front end to function properly.

The Transmission widget in Homepage hardcodes the value to the default and doesn't allow modifying it. This small refactor creates an `rpcUrl` optional prop on the Transmission widget that can be set to the same value as the `rpc-url` value in your Transmission `settings.json` file. If unset, the widget will use the default value of `/transmission/`.

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/81

### DIFF
--- a/src/widgets/transmission/proxy.js
+++ b/src/widgets/transmission/proxy.js
@@ -33,7 +33,9 @@ export default async function transmissionProxyHandler(req, res) {
     cache.put(`${headerCacheKey}.${service}`, headers);
   }
 
-  const url = new URL(formatApiCall(widgets[widget.type].api, { endpoint, ...widget }));
+  const api = `${widget.url}${widget.rpcUrl || widgets[widget.type].rpcUrl}rpc`;
+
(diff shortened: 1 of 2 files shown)


shamoon:
Sure, kinda specific but pretty harmless =)
Please do add the docs PR (you can target the v0.6.17 branch)

PR #1419, opened 2023-04-27
Fix: align services and widgets

juanmanuelbc:
## Proposed change

Fix misalignment of services and widgets...

Before: imatge

After: imatge

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/item.jsx
+++ b/src/components/services/item.jsx
@@ -34,7 +34,7 @@ export default function Item({ service }) {
       <div
         className={`${
           hasLink ? "cursor-pointer " : " "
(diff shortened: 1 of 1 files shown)


shamoon:
This doesn't work at all sizes (the height of blocks is dependent on the widget and the text in them...), but Im ok with it I suppose.

Screenshot 2023-04-27 at 2 20 52 AM

juanmanuelbc:
> This doesn't work at all sizes (the height of blocks is dependent on the widget and the text in them...), but Im ok with it I suppose.

I suppose the problem is in mixing services with and without description, isn't it???

PR #1433, opened 2023-04-30
solving docker multiple names

CiyLei:
Bug fix: Correct handling of docker multiple names

### DIFF
--- a/src/pages/api/docker/stats/[...service].js
+++ b/src/pages/api/docker/stats/[...service].js
@@ -30,7 +30,7 @@ export default async function handler(req, res) {
       });
     }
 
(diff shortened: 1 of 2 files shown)


shamoon:
what is the situation youre solving here?

CiyLei:
I have a container with multiple names
image

before,The first name contains /,I don't know how to fill in the **container** of **server.yaml**
image

after,I can fill out postgres
image

shamoon:
Yea, I guess with all that information why not just use `/joplin-server/pg`?

Anyway, this change seems fine thanks

CiyLei:
Because `/api/docker/status/joplin-server/pg/server` will not be parsed correctly

shamoon:
Ah, the slash in the name, anyway thanks

PR #1455, opened 2023-05-03
Replace fields in Nextcloud widget with file count and shared item count

GAntoine:
## Proposed change

I find that `CPU Load` and `Memory Usage` are pointless to query from Nextcloud, seeing as we have other ways to get that information from the system hosting it (resources, glances, etc.). So I've replaced them with `Files` and `Shared`.

Also adds a setting to the widget to choose the last active range (5 minutes, 1 hour, 24 hours).

Screenshot 2023-05-03 at 1 02 16 PM

## Type of change

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/88

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -549,7 +549,9 @@
         "cpuload": "Cpu Load",
         "memoryusage": "Memory Usage",
         "freespace": "Free Space",
-        "activeusers": "Active Users"
+        "activeusers": "Active Users",
+        "numfiles": "Files",
+        "numshares": "Shared Items"
     },
     "kopia": {
         "status": "Status",
--- a/src/components/services/widget/container.jsx
+++ b/src/components/services/widget/container.jsx
@@ -33,5 +33,5 @@ export default function Container({ error = false, children, service }) {
     }));
   }
 
-  return <div className="relative flex flex-row w-full">{visibleChildren}</div>;
+  return <div className="relative flex flex-row w-full">{visibleChildren.slice(0, 4)}</div>;
 }
--- a/src/components/services/widget/container.jsx
+++ b/src/components/services/widget/container.jsx
@@ -33,5 +33,5 @@ export default function Container({ error = false, children, service }) {
     }));
   }
 
-  return <div className="relative flex flex-row w-full">{visibleChildren.slice(0, 4)}</div>;
+  return <div className="relative flex flex-row w-full">{visibleChildren}</div>;
 }
--- a/src/widgets/nextcloud/component.jsx
+++ b/src/widgets/nextcloud/component.jsx
@@ -13,16 +13,20 @@ export default function Component({ service }) {
   if (nextcloudError) {
     return <Container service={service} error={nextcloudError} />;
   }
+  
+  // cpuload & memoryusage were deprecated, dont break existing installs & dont have > 4 blocks total
+  let deprecatedFieldsCount = widget.fields ? widget.fields.includes('cpuload') + widget.fields.includes('memoryusage') : 0;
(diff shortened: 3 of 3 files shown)


shamoon (changes requested):
Thanks for the PR.

One thing is you should only be updating en/common.json, all other files are automatically updated =)

The other issue with this PR is that it's a breaking change. I dont particularly care what stats nextcloud shows but anyone with `["cpuload", "memoryusage", "freespace", "activeusers"]` in their settings is going to get a surprise when they update

Seeing as there's basically no 'reason' for the change (eg its not like something is broken) you can either:

- Start a discussion about this change, make sure we get enough folks agreeing (I have no idea what that number is)
- Edit the widget to retain the old fields but support the new ones. It should still be a max of 4, however.

I think option 2 sounds better, personally.

Also, please remove the `activeUserPeriod` option, just use 24h (or whatever you prefer). Goal is keeping things simple.

GAntoine:
> Start a discussion about this change, make sure we get enough folks agreeing

Yeah that sounds good, I'll stat one in a bit and expand on the reason for these changes.

> Edit the widget to retain the old fields but support the new ones. It should still be a max of 4, however.

As in I could add a couple more fields, but they're disabled by default and users can enable them by replacing others, or there should only ever be 4 options in the `field` key?

> Also, please remove the activeUserPeriod option, just use 24h (or whatever you prefer).

I'll add that to the discussion as well and see what people prefer. Thanks for taking a look at this!

shamoon:
The max should be 4 no matter what, so in general we dont suggest having more than 4 options (I think it will just confuse users) but if that happens the widget needs to still limit it to 4.

For this one I would suggest making the cpu / mem fields 'optional' e.g. if user specified they will show, otherwise it will show these new ones. That makes it non-breaking for existing users but opt-out e.g. users who haven't specified will get your new-and-improved fields. LMK if that makes sense.

If you do that, no need to necessarily create a discussion, I'd be fine with merging like that.

GAntoine:
Ok I reset the PR history to just have the minimum changes needed. However I don't fully understand the "optional" fields. If I have no `fields` value for the widget, but there are more then 4 options in the component, it'll show all the fields.

Screenshot 2023-05-03 at 4 49 46 PM

I did add a line to container.jsx that only returns the first 4 children to render, which correctly limits the widget to 4 (original) fields.

Screenshot 2023-05-03 at 4 55 30 PM

Using `fields: ["freespace", "activeusers", "numfiles", "numshares"]`, or some other combination of values, you can get a custom list.

Screenshot 2023-05-03 at 4 50 24 PM

shamoon:
Please take a look and of course let me know if you notice any issues. I know its a little weird =)

Basically:

- Only show cpu / mem if specified in fields
- If cpu / mem specified in fields limit blocks to 4

I think this should handle various possibilities well, e.g.

No fields specified:
Screenshot 2023-05-03 at 2 31 40 PM

`["cpuload", "memoryusage", "freespace", "activeusers", "numfiles", "numshares"]`:
Screenshot 2023-05-03 at 2 32 11 PM

`["cpuload", "freespace", "activeusers", "numfiles"]`:
Screenshot 2023-05-03 at 2 32 58 PM

`["cpuload", "memoryusage", "numfiles"]`:
Screenshot 2023-05-03 at 2 33 38 PM

etc.

GAntoine:
Ahhhh ok that makes sense!! I've got ahead and refactored it a bit (and elaborated on the comments), but the behaviour stays the same.

`fields: ["cpuload", "freespace", "activeusers", "numfiles", "numshares"]`

Screenshot 2023-05-03 at 7 10 05 PM

shamoon:
Ha, ok great.

PR #1456, opened 2023-05-03
Set identical margin on all header widgets

GAntoine:
## Proposed change

Sets the same `margin-left` on all the header widgets so they align on narrow screens.

|Before|After|
|---|---|
|Screenshot 2023-05-03 at 1 09 08 PM|Screenshot 2023-05-03 at 1 10 33 PM|

## Type of change

## Checklist:

- [-] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: 

### DIFF
--- a/src/components/widgets/kubernetes/kubernetes.jsx
+++ b/src/components/widgets/kubernetes/kubernetes.jsx
@@ -45,7 +45,7 @@ export default function Widget({ options }) {
 
   if (!data) {
     return (
-      <div className="flex flex-col max-w:full sm:basis-auto self-center grow-0 flex-wrap">
+      <div className="flex flex-col max-w:full sm:basis-auto self-center grow-0 flex-wrap ml-4">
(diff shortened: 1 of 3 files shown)


shamoon:
Thanks for the PR but this just substitutes one problem for another IMHO. Now the resources widget doesn't left-align, which I dont like (I realize glances doesn't currently, which is not ideal either).

A solution here needs to implement something more like collapsing gutters, i.e. space between but not around. I still find this tricky to get right in CSS, but if you can figure that out please let us know and we can re-open this and we'd be happy to merge!

GAntoine:
> Now the resources widget doesn't left-align, which I don't like 

I see, didn't notice that when I was testing this. If I do find something that works it'll be on a different branch, so this PR can stay closed IMO.

shamoon:
Up to you what branch, thanks =)

PR #1485, opened 2023-05-10
Add support for dashboard-icons SVG file format

flor0:
## Proposed change

On my personal homepage I usually use dashboard-icons. Up until now I've only used PNGs but since they introduced some changes some of my icons are only available in SVG format for now. I think since dashboard-icons supports SVGs the user should have the choice what file format they want to use for their icons.

It's completely backwards compatible and the change only takes effect if the user specifies an icon with the .svg ending as compared to .png, so for switching file formats on dashboard-icons the user only has to change the file ending in the config yaml file. All other cases where the user uses SVGs are unaffected.

## Type of change

Added SVG support for dashboard-icons.

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/resolvedicon.jsx
+++ b/src/components/resolvedicon.jsx
@@ -57,7 +57,27 @@ export default function ResolvedIcon({ icon, width = 32, height = 32, alt = "log
     );
   }
 
+ 
   // fallback to dashboard-icons
+  if (icon.endsWith(".svg")) {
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks. GH actions is broken at the moment so the checks may fail but can re-run once fixed.

PR #1519, opened 2023-05-19
QNAP plugin, fix key expiry issue

theshaun:
## Proposed change

Hi All, I found an issue where the proxy was not correctly triggering a login event once the existing key had expired. I'm not sure if something changed between versions of QNAP, or I missed this during my testing.

I now check to see if the authPassed has gone false and then re log in to let the API routes work again.

Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:


theshaun:
Realised I should have cleaned my branch up a bit better before submitting, but its only pulling the change through which is good

shamoon:
Thanks, I'll mostly have to take your word on this since I cant test directly. I did add another check for status = 200, basically need to make sure api call succeeds after re-auth

theshaun:
Thanks @shamoon have had it running since the update (watchtower auto updates homepage for me) and its holding pretty steady so I think this is all good.

Thanks for adding that != 200 check, that is good for validation after it tries to retrieve a new token and it fails for whatever reason.

Cheers again!

PR #1537, opened 2023-05-23
Next pvr service widget

karl0ss:
## Proposed change
It doesn't look like anyone has requested this, but I use it alot, and thought people might also want it.

NextPVR service widget to show Upcoming and Recent recording counts.

image

I have updated the docs with a PR  [here ](https://github.com/benphelps/homepage-docs/pull/105)-

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "react-dom": "^18.2.0",
     "react-i18next": "^11.18.6",
     "react-icons": "^4.4.0",
+    "salted-md5": "^4.0.5",
     "shvl": "^3.0.0",
     "swr": "^1.3.0",
     "systeminformation": "^5.17.12",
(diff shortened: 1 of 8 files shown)


shamoon:
Thank you but we only accept service widget PRs based on feature requests with interest (currently arbitrarily designated as 5 "up-votes") from the community...

Feel free to open FR for this

See https://gethomepage.dev/en/more/development/#service-widget-guidelines

That info is also in the PR template (which looks like you didn’t use)

karl0ss:
Ok shamoon, 

Where is the community for voting on wanted widgets? As I don't appear to find one? (I did see the PR guidelines, hence saying I know no one had requested it (from what I could see))

Thanks

shamoon:
Not sure if you found it after this comment but posting a discussion in the feature requests category allows users to click the up-arrow

PR #1560, opened 2023-06-02
Feature/sonarr radarr queue list

Schoggi0815:
## Proposed change

I have added a simple queue-list to the Sonarr and Radarr widgets, where one can see the current queue.
image
image
image

Currently this feature is toggled together with the sonarr.queue / radarr.queue field, since I wasn't sure on how you create new translations for new fields, but of course it could be changed to have it's own field.
I'm aware that this would add another row to those widgets, but I would like to have this feature, since the queue can't otherwise be seen by people who don't have full access to Sonarr and Radarr and since it's togglable I don't think it would be an issue.

The APIs of the widgets had to be expanded, to also get the required queue information and title information, but I made sure, no information, that isn't necessary, gets to the client, instead it's filtered on the server.

Closes # (no issue)

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/widget/block-list.jsx
+++ b/src/components/services/widget/block-list.jsx
@@ -1,0 +1,31 @@
+import { useTranslation } from "next-i18next";
+import { useCallback, useState } from 'react';
+import classNames from "classnames";
+
+import ResolvedIcon from '../../resolvedicon';
+
+
+export default function BlockList({ label, children, childHeight }) {
+  const { t } = useTranslation();
+  const [isOpen, setOpen] = useState(false);
+
+  const changeState = useCallback(() => setOpen(!isOpen), [isOpen, setOpen]);
+
+  return (
+    <div
+      className={classNames(
+        "bg-theme-200/50 dark:bg-theme-900/20 rounded m-1 w-full p-1",
+        children === undefined ? "animate-pulse" : ""
+      )}>
+      <button type="button" onClick={changeState} className="w-full flex-1 flex flex-col items-center justify-center text-center">
+        <div className="font-bold text-xs uppercase">{t(label)}</div>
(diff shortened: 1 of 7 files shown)


shamoon (changes requested):
Thanks for the PR, before we can do a full code review (there are some issues), this PR should be changed to:
 
 1. Have styling consistent with eg the Tautulli now playing widget 
 2. This needs to be opt-in not opt-out, it should definitely not be linked to the queue or any other field but should be it’s own option. I’m not sure what you mean about adding new translations but we can help figure it out if needed.

Schoggi0815:
Thanks for the quick response.

I have reworked the styling and made the feature opt-in.

shamoon:
Thanks, there were some issues. Please take a look at my changes and let me know if you have any concerns

Screenshot 2023-06-05 at 10 11 32 PM

1. I am not happy with the dropdown thing, it's so much dead space, I find it finicky and kinda un-homepage-y. Again, consistency with other similar components is just to remove it.
2. The play / pause thing was very confusing because it looks like a button, the downloading text is enough I think.
3. For whatever reason, my install had an item in the queue that didnt return proper episode data, I updated to handle this.
4. If two items in the radar are the same movie (ID) you get a react error. I've fixed this (in both) by tacking on `sizeLeft`, which should almost guarantee uniqueness.

Screenshot 2023-06-05 at 8 18 54 PM
Screenshot 2023-06-05 at 9 32 51 PM
Screenshot 2023-06-05 at 8 38 30 PM

Schoggi0815:
Thanks for the review, I've taken a look at your changes and don't have any concerns.
I wasn't sure about the dropdown, since the sonarr queue can get quite long, but considering that the current design is now less than half the size compared to my original, it should be fine.
Everything else looks perfectly fine to me.

PR #1561, opened 2023-06-03
Translated ru/common.json

Dimidroll06:
## Proposed change

Translated untranslated locales into Russian

Closes # (issue)

## Type of change

- [ ] Other (please explain)


shamoon:
Thank you but all translation should be handled via weblate, see https://hosted.weblate.org/engage/homepage/

PR #1571, opened 2023-06-05
Fix: use service labels instead of container for service discovery with Docker Swarm

### LINKED ISSUE #752: Use service labels and average value in docker Swarm mode
### Description

When we are using Docker SwarmMode with possible multiple replica for each services, we need to be linked to the service in place of container instance. 
In this case we need to define the label at the service level (service/deploy/labels)  to obtain one effective link to the service  and the widget data must use the average value of all replica in the service. 

This is a little bit more complicated  but more  effective.

### Other

_No response_

rkokkelk:
## Proposed change

The current implementation of service detection focuses on containers. However when using Docker Swarm this limits automatic service detection to the containers running on the same node as homepage (or docker-proxy). Detecting other services in the docker swarm is with the current implementation not possible.

This commit fixes this by getting the appropriate labels from services when Docker swarm is configured in the config file. This ensures that the appropriate labels are gathered from the service definition instead of the container definiation, thus allowing for automatic service detection for the entire Docker swarm. Please note that in order for this to function the homepage (or dockerproxy) should be running on a manager node. Only the manager node is able to gather all the relevant service labels.

Closes #752, #970, #1255, #1045

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

  * https://github.com/benphelps/homepage-docs/pull/107

### DIFF
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -62,13 +62,10 @@ export async function servicesFromDocker() {
 
   const serviceServers = await Promise.all(
     Object.keys(servers).map(async (serverName) => {
-      const isSwarm = servers[serverName].swarm ?? false;
-
       try {
-        const listProperties = { all: true };
-        const labelProperty = (isSwarm) ? 'Spec.Labels' : 'Labels';
-        const nameProperty = (isSwarm) ? 'Spec.Name' : 'Name[0]';
+        const isSwarm = !!servers[serverName].swarm;
         const docker = new Docker(getDockerArguments(serverName).conn);
+        const listProperties = { all: true };
         const containers = await ((isSwarm) ? docker.listServices(listProperties) : docker.listContainers(listProperties));
 
         // bad docker connections can result in a <Buffer ...> object?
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks for the PR, my only question is whether there is a possibility this will be backwards-incompatible for folks? I don’t use swarm in production, I have just spun up instances for testing homepage but it seems there are a couple different ways things might be setup. Point is if someone has this setup at the container level might this change potentially cause that to stop working?

rkokkelk:
It is a bit yes and no. Basically, when you set up Docker swarm on a single node and use container labels, automatic service detection with this PR will no longer work. However, this setup is inherently no different that running docker on a single system without Swarm. In this case removing the `swarm: true` from your config will return the automatic service detection.

Docker Swarm is meant for managing a cluster of Docker environments, so more than one single system. So when you use Docker swarm, as it is meant to be, container labels will never work for detecting all services because this information is simply not known by the Docker swarm manager. In that case service labels are required for it to the automatic service detection to work correctly function.

That why in homepage the status check now correctly checks the service status and not the container status. 

TLDR:
* For users which use docker swarm (1 node) and container labels, they should remove `swarm: true` from config. Technically they do not really use docker swarm. 
* For users which use docker swarm (2+ nodes) and container labels, they should change this to service labels, (in this setup automatic service detection currently would not work properly a 100%)
* For users which use docker swarm (2+ nodes) and service labels, it will now work.

rkokkelk:
Looks good to me!

PR #1579, opened 2023-06-07
Fixes for portainer and kubernetes and error component

denispapec:
## Proposed change

Fixes for portainer when the environment is offline, and kubernetes widgets when it tries to pass error message.

A change in https://github.com/benphelps/homepage/pull/1437 broke the error output when services are behind the reversed proxy as proxy will return buffer data instead, so original code was working as intended.

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/widget/error.jsx
+++ b/src/components/services/widget/error.jsx
@@ -9,10 +9,12 @@ function displayData(data) {
   return (data.type === 'Buffer') ? Buffer.from(data).toString() : JSON.stringify(data, 4);
 }
 
-export default function Error({ error: err }) {
+export default function Error({ error }) {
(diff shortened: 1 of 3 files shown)


shamoon:
What is portainer's api output when the environment is offline?

denispapec:
`data` in `useWidgetApi` is:
```json
{
  "message": "Unable to find an environment with the specified identifier inside the database",
  "details": "object not found inside the database"
}
```
I get that this will catch everything that is not list of containers, but looking at the Portainer documentation, it seems that any other response that is not list of containers is an object with message of something going wrong.

PR #1626, opened 2023-06-21
Feature: collapsible layout sections

ionyx0:
## Proposed change
This allows for collapsible `ServiceGroup` objects in the UI.  I have introduced an optional configuration value in the `settings.yaml` file to control this.  `disableCollapse: true` (defaults to `false`) will disable the component that collapses the different `ServiceGroup` sections.

This can also be included in `BookmarksGroup` at a later date.  The configuration option can also be included at the individual group level at a later date.

Closes #1210

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/group.jsx
+++ b/src/components/services/group.jsx
@@ -1,13 +1,11 @@
 import classNames from "classnames";
-import { Transition } from '@headlessui/react'
-import { useState } from 'react'
+import { Disclosure, Transition } from '@headlessui/react';
+import { MdKeyboardArrowDown } from "react-icons/md";
 
 import List from "components/services/list";
 import ResolvedIcon from "components/resolvedicon";
 
 export default function ServicesGroup({ group, services, layout, fiveColumns, disableCollapse}) {
-
-  const [isShowing, setIsShowing] = useState(true)
 
   return (
     <div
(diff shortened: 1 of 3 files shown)


shamoon (changes requested):
Im not sure about adding an entire UI library for a collapse feature. Probably a question for @benphelps , but I think this would be straightforward to implement natively and would prefer that to bringing in an entire (not small) package.

ionyx0:
> Im not sure about adding an entire UI library for a collapse feature. Probably a question for @benphelps , but I think this would be straightforward to implement natively and would prefer that to bringing in an entire (not small) package.

Agreed.  I'm quite new to next.js, however.  I'll be looking to refactor.  Any suggestions are welcome.

ionyx0:
> > Im not sure about adding an entire UI library for a collapse feature. Probably a question for @benphelps , but I think this would be straightforward to implement natively and would prefer that to bringing in an entire (not small) package.
> 
> Agreed. I'm quite new to next.js, however. I'll be looking to refactor. Any suggestions are welcome.

Updated to use a package that already exists in the project.

shamoon:
Thanks, this still needs some work.

- In general, I dont recommend disabling eslint rules as a way to fix warnings, they generally are there for a reason, please see the change to fix that.
- There needs to be some more user interaction to make this more discoverable, I think a hover to show a collapse arrow or something like that
- This currently breaks icons in the group title

shamoon:
Ok, happier with this:

https://github.com/benphelps/homepage/assets/4887959/ac47e398-a9cf-46c9-b7ff-df87d4071f73

Only question is whether to have this default to enabled or disabled

ionyx0:
I am much happier with the changes you made @shamoon!  Thank you for contributing.  I am trying to get more familiar with React and its concepts since I live and work in Java every day and don't get a chance to do so.

PR #1627, opened 2023-06-22
fix(kubernetes): add a boolean to disable traefik

dotdiego:
Hello as seen in this discussion https://github.com/benphelps/homepage/discussions/1584 when using Kubernetes without Traefik we have a constant API Error showing.

I tried to implement a boolean in kubernetes.yaml configuration to stop looking for traefik services.

You need to add the `traefik: false` option to your kubernetes.yaml and it will stop looking for traefik ingressroutes.

### DIFF
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -1,4 +1,4 @@
-import { promises as fs } from "fs";
+import { promises as fs, readFileSync } from "fs";
 import path from "path";
 
 import yaml from "js-yaml";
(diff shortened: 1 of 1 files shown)


src/utils/config/service-helpers.js
```diff
@@ -173,10 +159,37 @@ export async function servicesFromKubernetes() {
...
+    const crd = kc.makeApiClient(CustomObjectsApi);
+
+    const ingressList = await networking.listIngressForAllNamespaces(null, null, null, null)
+      .then((response) => response.body)
+      .catch((error) => {
+        logger.error("Error getting ingresses: %d %s %s", error.statusCode, error.body, error.response);
+        return null;
+      });
+
+    if (config.traefik) {
```
shamoon, line 191: Presumably this will disable traefik by default, breaking all installations. You'd need to test for equality with `false`

src/utils/config/service-helpers.js
```diff
@@ -173,10 +159,37 @@ export async function servicesFromKubernetes() {
...
 
-    if (traefikIngressList && traefikIngressList.items.length > 0) {
-      const traefikServices = traefikIngressList.items
-      .filter((ingress) => ingress.metadata.annotations && ingress.metadata.annotations[`${ANNOTATION_BASE}/href`])
+  try {
+
+    const configFile = path.join(process.cwd(), "config", "kubernetes.yaml");
+    const rawConfigData = readFileSync(configFile, "utf8");
+    const configData = substituteEnvironmentVars(rawConfigData);
+    const config = yaml.load(configData);
```
shamoon, line 175: This is redundant, thats what `getKubeConfig` does

shamoon (changes requested):
I cant test this since I dont run Kubernetes, but I think there are some changes needed. It would be good if someone who does run k8s can, to be sure

shamoon:
Closing as stale.

PR #1712, opened 2023-07-21
Adding Azure PullRequests

Nizzan943:
## Proposed change
Add new service widget - Azure Pull Requests. 
use to monitor pull requests status in specific repository.

image

Closes #1693 

## Type of change

- [X] New service widget
- [ ] Other (please explain)

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: [127](https://github.com/benphelps/homepage-docs/pull/127)


shamoon:
This really should be combined with the other one, I think. I appreciate they're not exactly the same data...

Nizzan943:
Its all two different widget that does not live together.. the one can (and probably sometimes will) live without the other..

shamoon:
The other widget only has 2 blocks, this one 3, combine the three here or drop one and you can make one "Azure DevOps" widget or whatever. Since they literally use the same API, they should be the same widget, we're not going to maintain two different ones.

Nizzan943:
> The other widget only has 2 blocks, this one 3, combine the three here or drop one and you can make one "Azure DevOps" widget or whatever. Since they literally use the same API, they should be the same widget, we're not going to maintain two different ones.

I disagree on this, each widget requires different access in organization portal, fetching different api and different fields (pipelines requires branch name as optional and pipelineId, pull requests requires repositoryId and userEmail ) and have different meaning. and the most crucial topic is there is no logic to place both on the same card. they have all different meaning.
I think combine them together will end in big `if else` that each one will hold a whole logic inside.
Is that really worth it?

Nizzan943:
@shamoon I was able to manage it by adding another replace of `{}` in `api-helpers.js
to:
```
export function formatApiCall(url, args) {
  const find = /\{.*?\}/g;
  const replace = (match) => {
    const key = match.replace(/\{|\}/g, "");
    return args[key] || "";
  };

  **return url.replace(/\/+$/, "").replace(find, replace).replace(find,replace);**
}
```
please tell me what you think about it, and if everything seems fine to you, we can abandon this PR and create new one that merge all to [azureDevOps](https://github.com/Nizzan943/homepage/tree/merge-to-azureDevOps/src/widgets/azureDevOps).

shamoon:
Yea something like that is what I had in mind. I don’t think I’ll be able to today but feel free to open a new PR with that and we’ll take a look asap

PR #1715, opened 2023-07-23
Add Azure DevOps

Nizzan943:
## Proposed change
add new features to Azure Pipelines and merged together as Azure DevOps.

use to monitor pull requests status in specific repository.

image

Closes #1693 

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: [127](https://github.com/benphelps/homepage-docs/pull/127)


Nizzan943:
The only con in this way is we fetching widget 2 api requests and not 1.
Because of useWidgetAPI hook should be before ```if(azureType)```. looking for advice

shamoon (changes requested):
This looks good for the most part but I think we need to shorten "My Approved Not Completed" text, "Incomplete"? "Approved"?

Also where does the 10 reviewers thing come from?

Nizzan943:
> This looks good for the most part but I think we need to shorten "My Approved Not Completed" text, "Incomplete"? "Approved"?
> 
> Also where does the 10 reviewers thing come from?

Done.

I know only 3 peoples that hit up-vote, we are all using homepage and love with the product.

shamoon (changes requested):
So a couple things:

1. Even though they are kinda two widgets you can merge some things, for example the error display should be one block.
2. Currently this doesnt allow showing both types of the widget. Do you want to consider supporting that? As I mentioned in the other PR if you can keep it under 4 total blocks then you could conceivably have both at the same time.
3. If not then there is a way to actually limit the API calls, you would change the second parameter to `useWidgetAPI` i.e. `count endpoint = (type === "Pipeline") ? "pipeline" : "pr"` and then call it once.

But again, I think it would be cooler to have them both work at the same time, users can disable any that they dont want with the `fields` option

Nizzan943:
Hey,
I dropped `BuildId` Block.
I tried to make both functionalities work together, but unfortunately, it seems to be not convenient.
If the user wants to have both functionalities, it works fine because the widget has all the relevant fields to fetch both API requests.
but let's say if user wants only a pipeline, so he wants to use this configuration:
```
          type: azuredevops
          fields: ["result", "status"]
          organization: some-organization
          project: some-project
          userEmail:  some@email.com
          branchName: some-branch-name
          definitionId: some-id
          key: keykeykey
```
The issue is that we are missing branchName field (from PR functionality) which leads to the widget fetching this endpoint:
`..../_apis/git/repositories//pullrequests` and gets 404.
the opposite way is not an issue, because `builds/Build` endpoint will just give back the last pipeline that ran, the user can filter it with `fields` parameter.

Regardless of what I wrote up here, I have weird behavior after merging widgets that make widgets crazy, nothing seems to be broken. When this situation happens to me in the video, I used all fields for each widget to prevent 404 from PR endpoint.
so I used, but with real data:
```
          type: azuredevops
          fields: ["result", "status"]
          organization: some-organization
          project: some-project
          userEmail:  some@email.com
          branchName: some-branch-name
          definitionId: some-id
          repositoryId: some-repository          
          key: keykeykey
```
https://github.com/benphelps/homepage/assets/59150911/0fa82d30-7fa2-412f-b100-6905fdaf32c8

shamoon:
If you push the changes you have I can try to help out tomorrow

Nizzan943:
> If you push the changes you have I can try to help out tomorrow

yeah sorry I did it in 4am last night :D
pushed

shamoon:
Ok take a look at the last commit, it should handle this stuff a bit better.

If `branchName` & `userEmail` aren't supplied we can assume user doesnt want that so just hide the fields, ignore any api errors for PR. Arguably we could have a similar logic the other way around, we can still add that but I think as you mentioned it won't error and they can just hide the field.

Play with it a bit and see if you note any issues, we'll squash the commits and resolve the merge errors once its ready.

Nizzan943:
I did and it seems good to me.
I accidently told you the `branchName` is the issue but it was repositoryId, sorry it was 4AM :)

2 things:
1. I dont like the idea we are fetching api request everytime twice when it not needed, how it can be solved?
2. using fields for unknown reason crashing widget.
image
image
Am I use it wrong?
both widgets in the picture has same settings but `Repo 1` have also ```fields: ["totalPrs", "myPrs","approved"]```

shamoon:
should be sorted, lmk

Screenshot 2023-07-26 at 8 33 58 PM
Screenshot 2023-07-26 at 8 34 06 PM

shamoon (approved):

PR #1724, opened 2023-07-25
Support custom colors for icons

Mega-Volti:
## Proposed change

Enables appending a color code (e.g. "#123456") to all mdi and si icons, in order to change their color to a per-icon custom one.
If no color code is appended, defaults to current behaviour (depending on theme setting).

Example: mdi-folder-#123456

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/resolvedicon.jsx
+++ b/src/components/resolvedicon.jsx
@@ -34,33 +34,27 @@ export default function ResolvedIcon({ icon, width = 32, height = 32, alt = "log
 
   // check mdi- or si- prefixed icons
   const prefix = icon.split("-")[0];
-  const suffix = icon.split("-")[icon.split("-").length - 1];
-  
+
   if (prefix in iconSetURLs) {
     // check whether icon ends with color code
-    if (suffix.test(`[#][a-fA-F0-9][a-fA-F0-9][a-fA-F0-9][a-fA-F0-9][a-fA-F0-9][a-fA-F0-9]`)) {
+    let iconName;
+    let iconColor;
+    const suffix = icon.split("-")[icon.split("-").length - 1];
 
-      const iconColor = `${suffix}`;
-      const iconName = icon.replace(`${prefix}-`, "").replace(".svg", "").replace(`-${suffix}`, "");
-      const iconSource = `${iconSetURLs[prefix]}${iconName}.svg`;
-      return (
-        <div
-          style={{
-            width,
-            height,
-            maxWidth: '100%',
-            maxHeight: '100%',
-            background: `${iconColor}`,
-            mask: `url(${iconSource}) no-repeat center / contain`,
-            WebkitMask: `url(${iconSource}) no-repeat center / contain`,
-          }}
-        />
-      );
-    };
-    
-    // default to theme setting if no custom icon color code is provided
-    const iconName = icon.replace(`${prefix}-`, "").replace(".svg", "");
+    if (!(suffix.match(`[#][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]`) == null)) {
+      // set custom color code
+      iconName = icon.replace(`${prefix}-`, "").replace(".svg", "").replace(`-${suffix}`, "");
+      iconColor = `${suffix}`;
+    } else {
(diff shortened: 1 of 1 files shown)


src/components/resolvedicon.jsx
```diff
@@ -33,30 +33,53 @@ export default function ResolvedIcon({ icon, width = 32, height = 32, alt = "log
...
+            maxHeight: '100%',
+            background: `${iconColor}`,
+            mask: `url(${iconSource}) no-repeat center / contain`,
+            WebkitMask: `url(${iconSource}) no-repeat center / contain`,
+          }}
+        />
+      );
+    };
 
+    if (suffix.match(`[#][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]`) == null) {
```
benphelps, line 62: Should be an else statement here, instead of checking the suffix twice.

Mega-Volti: I had that first but then some check complained and it wouldn't compile properly when building the Dockerfile. This "ugly" check was the only way I got the Dockerfile to build, sadly.

benphelps: Ah right, don't wrap it in an if statement at all then, the if before it guards by returning when it matches.

Mega-Volti: Ah, right, I couldn't figure out why it didn't like the return-else, but I think I understand now. Fixed with the latest commit, and building the Dockerfile works now, thanks!

src/components/resolvedicon.jsx
```diff
@@ -33,13 +33,34 @@ export default function ResolvedIcon({ icon, width = 32, height = 32, alt = "log
...
-  const prefix = icon.split("-")[0]
-
+  const prefix = icon.split("-")[0];
+  const suffix = icon.split("-")[icon.split("-").length - 1];
+  
   if (prefix in iconSetURLs) {
-    // get icon source
+    // check whether icon ends with color code
+    if (!(suffix.match(`[#][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]`) == null)) {
+
```
shamoon, line 42:
```suggestion
    if (suffix.test(`[#][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]`)) {

```

use `test` to check if a regex matches

shamoon (changes requested):
Instead of one big if / else please change this to set the appropriate const's and the output the icon, i.e. DRY

src/components/resolvedicon.jsx
Mega-Volti: Done, and I also included upper case letters since the regex was case sensitive so far.

Mega-Volti:
> Instead of one big if / else please change this to set the appropriate const's and the output the icon, i.e. DRY

Updated. Although I had to switch from const to let. With const block scope, I had to repeat. Eliminating the repetition means I have to declare the variables outside of the if statement, hence no block scoped const, so I used let. I'm new to js, currently I don't see a better way, I'd be happy for pointers if I'm missing something.

Regarding the regex test: I updated my comment earlier and reverted the commit, the test function doesn't work on strings.

shamoon:
Please take a look at the last commit:

- Yea sorry that `test` snippet was written on my phone, it was incorrect as you noted
- Use regex literals
- Default vars to the current behavior and only change them if custom color detected

Let me know if you have any concerns or notice any issues

Mega-Volti:
Looks great, that matcher is indeed better, thanks!
No concerns or major issues from my side. I think I would have gone with an if..else statement to write the iconName and iconColor only once, instead of writing it and possibly overwiritng it again with a different value. But that's personal preference I guess. As long as it works, I'm very happy with it!

shamoon (approved):

hex-developer:
Hello @benphelps,
Is there a chance you will implement this feature for other svg icons? I want to use some custom icons (https://tabler-icons.io) and they show up black, which doesn't work very well with my dark theme! Alternatively you could also add built-in support for tabler icons. I might submit a PR at sone point.

PR #1729, opened 2023-07-26
Enable custom refresh rate for resource widget

Mega-Volti:
## Proposed change

Adds the option to set a custom refresh rate for the resource widget. It reads the "refresh" setting in the widget options and, if it is a number, sets the refresh rate to this number. In case it is not a number, it defaults to the previous behavior (default refresh rate of 1500 ms).

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/widgets/resources/cpu.jsx
+++ b/src/components/widgets/resources/cpu.jsx
@@ -5,11 +5,11 @@ import { useTranslation } from "next-i18next";
 import Resource from "../widget/resource";
 import Error from "../widget/error";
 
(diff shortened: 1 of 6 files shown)


shamoon (changes requested):
You can simplify these by giving refresh a default value (which tbh is kinda pointless since they will always have it set, but still). Also there should be a minimum enforced value in the parent, maybe 1000ms

shamoon (approved):

PR #1788, opened 2023-08-08
Fix(AzureDevOps): PR approved count and Pipeline running status

Nizzan943:
## Proposed change

- Fix an issue when fetching pipeline request with `$top=1`, running pipelines not return in response.
- Fix an issue where `Approved with Suggestions` status is applied to PR but not count as `Approved` in filtering.
Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/azuredevops/component.jsx
+++ b/src/widgets/azuredevops/component.jsx
@@ -56,7 +56,7 @@ export default function Component({ service }) {
         value={t("common.number", {
           value: prData.value
             ?.filter((item) => item.createdBy.uniqueName.toLowerCase() === userEmail.toLowerCase())
-            .filter((item) => item.reviewers.some((reviewer) => reviewer.vote === 10 || reviewer.vote === 5)).length
+            .filter((item) => item.reviewers.some((reviewer) => [5,10].includes(reviewer.vote))).length
         })}
       />}
 


src/widgets/azuredevops/component.jsx
```diff
@@ -56,7 +56,7 @@ export default function Component({ service }) {
         value={t("common.number", {
           value: prData.value
             ?.filter((item) => item.createdBy.uniqueName.toLowerCase() === userEmail.toLowerCase())
-            .filter((item) => item.reviewers.some((reviewer) => reviewer.vote === 10)).length,
+            .filter((item) => item.reviewers.some((reviewer) => reviewer.vote === 10 || reviewer.vote === 5)).length
```
shamoon, line 59: This is weird because I assume youre intentionally saying vote = 6/7/8/9 should not count? But effectively this will show approved for vote = 5 even if its not (when its not approved with suggestions). Without looking at the API at all, is there no way to detect the 'with suggestions' part?

Nizzan943: 6/7/8/9 is not a valid vote id.

`10 - approved 
5 - approved with suggestions 
0 - no vote 
-5 - waiting for author 
-10 - rejected `

source: https://learn.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/get-pull-requests?view=azure-devops-rest-7.0&tabs=HTTP#identityrefwithvote

shamoon: Ah great, I thought it was a vote count. Makes much more sense 😀

shamoon, line 59:
```suggestion
            .filter((item) => item.reviewers.some((reviewer) => [5,10].includes(reviewer.vote))).length
```

PR #1804, opened 2023-08-11
Forward cookies from request

j-waters:
## Proposed change

Take any cookies that are present on the request to endpoints such as `api/ping` and `api/services` and add them to the cookie jar. This means that services that are behind authentication such as authelia can be pinged and used in widgets.

The way this is done probably isn't overly correct - but it works for my usage. I thought I'd make an MR anyway to see if there was any interest in this feature.

Closes #1797

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/utils/proxy/cookie-jar.js
+++ b/src/utils/proxy/cookie-jar.js
@@ -52,7 +52,7 @@ export function importCookieHeader(url, cookieHeader) {
             // Otherwise we add a new cookie
             cookieJar.setCookieSync(new Cookie({
                 key, value
-            }), url.toString(), { ignoreError: false });
+            }), url.toString(), { ignoreError: true });
         }
     }
 }
--- a/src/utils/proxy/cookie-jar.js
+++ b/src/utils/proxy/cookie-jar.js
@@ -48,7 +48,9 @@ export function importCookieHeader(url, cookieHeader) {
         let existingCookie;
         try {
             existingCookie = cookieJar.getCookiesSync(url).find(existing => existing.key === key);
-        } catch {}
+        } catch (e) {
+            console.debug(`Failed to get cookies for ${url}: ${e}`)
+        }
 
         if (existingCookie) {
             existingCookie.value = value;
--- a/src/utils/proxy/cookie-jar.js
+++ b/src/utils/proxy/cookie-jar.js
@@ -37,3 +37,12 @@ export function addCookieToJar(url, headers) {
     cookieJar.setCookieSync(cookies[i], url.toString(), { ignoreError: true });
   }
 }
+
+export function importCookieHeader(url, cookieHeader) {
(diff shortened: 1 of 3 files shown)


src/utils/proxy/cookie-jar.js
```diff
@@ -37,3 +37,22 @@ export function addCookieToJar(url, headers) {
     cookieJar.setCookieSync(cookies[i], url.toString(), { ignoreError: true });
   }
 }
+
+export function importCookieHeader(url, cookieHeader) {
```
shamoon, line 41: Shouldn't this be just `importCookie` or `importOrCreateCookie`? (header has nothing to do with it)

src/utils/proxy/cookie-jar.js
```diff
@@ -37,3 +37,22 @@ export function addCookieToJar(url, headers) {
...
+    const cookies = cookieHeader.split(';')
+    for (let i = 0; i < cookies.length; i += 1) {
+        const [key, value] = cookies[i].trim().split('=')
+
+        // If there's an existing cookie with a matching key for this url,
+        // we want to update it
+        const existingCookie = cookieJar.getCookiesSync(url).find(existing => existing.key === key)
```
shamoon, line 48: If url isn't reachable this fails very noisily. Perhaps wrap in a try / catch

shamoon (commented):
Can you explain a bit more about how youre using this specifically? When are the authelia headers getting set initially?

j-waters:
> Can you explain a bit more about how youre using this specifically? When are the authelia headers getting set initially?

So I have a bunch of services on a subdomain, all behind authelia. Requests made to `homepage.domain.com` have an `authelia_session=<TOKEN>` cookie set. That same cookie is consumed by other services on the same domain.

The idea here is that when a request comes in to homepage, we take the cookies from the cookie header, add them to the cookie jar, and then they can be used in requests to the actual services.

I've marked the PR as draft because in production for me it works with pings but not with widgets.

src/utils/proxy/cookie-jar.js
j-waters: This function takes a [cookie header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) rather than an individual cookie so to me it makes sense but I'm happy to change

shamoon:
I still havent tested this out much, but it looks OK to me. Any reason its still a draft?

shamoon:
@j-waters if youre still interested in this let us know, otherwise we'll close it out

j-waters:
Sorry yeah I'm still interested, been using this branch in my home setup for about a month now and haven't found any issues

shamoon:
I'd like to add this to the docs somewhere, not sure where. Hmm...

PR #1807, opened 2023-08-12
OpenMediaVault widget

userXinos:
## Proposed change

Widget for integration with OpenMediaVault (https://github.com/benphelps/homepage/discussions/1204)

изображение

```yaml
- OpenMediaVault:
    - Services:
        widget:
          type: openmediavault
          url: http://
          username: admin
          password: pass
          method: services.getStatus
    - S.M.A.R.T.:
        widget:
          type: openmediavault
          url: http://
          username: admin
          password: pass
          method: smart.getListBg
    - Downloader Plugin:
        widget:
          type: openmediavault
          url: http://
          username: admin
          password: pass
          method: downloader.getDownloadList
```

OMV RPC docs: https://docs.openmediavault.org/en/6.x/development/tools/omv_rpc.html
Real docs: [Source Code on GH](https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault/usr/share/openmediavault/engined/rpc)

## Type of change

- [x] New service widget
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Other (please explain)

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added a corresponding PR to the documentation:[https://github.com/benphelps/homepage-docs/pull/140](https://github.com/benphelps/homepage-docs/pull/140) 
- [X] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/en/more/development/#service-widget-guidelines).
- [X] If applicable, I have checked that all tests pass with e.g. `pnpm lint`.
- [X] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #1817, opened 2023-08-16
OpenMediaVault widget: Part 2

userXinos:
## Proposed change

Sequel for #1807 

изображение

```yaml
# services.yaml

- OpenMediaVault:
#...
    - ZFS Hits/Misses:
        widget:
          type: openmediavault
          url: http://
          username: admin
          password: yes
          method: zfs.getStats
```

```yaml
# widgets.yaml

- openmediavault:
    url: http://
    username: admin
    password: yes
    updates: true
    expanded: true
```

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

- [ ] If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: https://github.com/benphelps/homepage-docs/pull/141


userXinos:
[http.zip](https://github.com/benphelps/homepage/files/12358605/http.zip)

shamoon:
Since there havent been requests specifically for either of these:

Feels like updates in an info widget is weird, I think of those for persistently-relevant data. I wonder if would be better as another service widget option.

ZFS charts one is cool because charts are cool but I wonder how meaningful it is, again there was no discussion on these.

Overall Im a little concerned because it feels like we already do / would have even more of a large amount of OMV-related code to maintain, imbalanced with most other services we support, for something that I suspect is not used by the vast majority of [our] users based off votes etc

Welcome your thoughts

userXinos:
The update widget will then look like this, is it really necessary in this form?  As an info widget, I thought it would be more accurate.
Screenshot_20230817_133105

ZFS, these charts are also in TrueNas, it seems there is some demand, why not port them

I'm not sure that the implementation of endpoints will be expensive, because so far this is a copy-paste to a large extent. 

If there is some kind of off-limit on the number of endpoints, then ok, I just implemented what I'm interested in and in this part, what's interesting is secondary.  I thought to keep the first (then this) PR open, so as not to be accepted and used as a draft.

shamoon:
There may be demand for ZFS data on a truenas dashboard which is a dedicated space for that but not necessarily the same in homepage... Homepage does not aim to be an "everything" command center, more a very nice dashboard with some useful info.

My biggest concern is code maintenance, etc. This is a lot of code for just one ecosystem used by a minority of users (I think).

My suggestion would be to move the updates info into an (option) of the service widget and drop the graph widget.

userXinos:
Oh, this ?
изображение

src/widgets/openmediavault/proxy.js
```diff
@@ -11,6 +13,7 @@ const BG_POLL_PERIOD = 500;
 
 const logger = createLogger(PROXY_NAME);
 
+// noinspection DuplicatedCode
```
shamoon, line 16:
```suggestion
```

src/widgets/openmediavault/methods/smart_get_list.jsx
```diff
@@ -1,3 +1,5 @@
+// noinspection JSUnresolvedVariable
+
```
shamoon, line 2:
```suggestion
```

src/widgets/openmediavault/methods/downloader_get_downloadlist.jsx
```diff
@@ -1,3 +1,5 @@
+// noinspection JSUnresolvedVariable
+
```
shamoon, line 2:
```suggestion
```

src/widgets/openmediavault/methods/apt_enumerateUpgraded.jsx
```diff
@@ -0,0 +1,37 @@
+// noinspection JSUnresolvedVariable
+
```
shamoon, line 2:
```suggestion
```

src/widgets/openmediavault/methods/downloader_get_downloadlist.jsx
```diff
@@ -10,10 +12,11 @@ const downloadReduce = (acc, e) => {
 };
 
 const items = [
-  { label: "openmediavault.downloading", getNumber: (data) => (!data ? null : data.reduce(downloadReduce, 0)) },
-  { label: "openmediavault.total", getNumber: (data) => (!data ? null : data?.length) },
+  { label: "openmediavault.downloading", getNumber: (data) => data.reduce(downloadReduce, 0) },
+  { label: "openmediavault.total", getNumber: (data) => data?.length },
 ];
 
+// noinspection DuplicatedCode
```
shamoon, line 19:
```suggestion
```

src/widgets/openmediavault/methods/services_get_status.jsx
```diff
@@ -16,11 +16,12 @@ const notRunningReduce = (acc, e) => {
...
 const items = [
-  { label: "openmediavault.running", getNumber: (data) => (!data ? null : data.reduce(isRunningReduce, 0)) },
-  { label: "openmediavault.stopped", getNumber: (data) => (!data ? null : data.reduce(notRunningReduce, 0)) },
-  { label: "openmediavault.total", getNumber: (data) => (!data ? null : data?.length) },
+  { label: "openmediavault.running", getNumber: (data) => data.reduce(isRunningReduce, 0) },
+  { label: "openmediavault.stopped", getNumber: (data) => data.reduce(notRunningReduce, 0) },
+  { label: "openmediavault.total", getNumber: (data) => data?.length },
 ];
 
+// noinspection DuplicatedCode
```
shamoon, line 24:
```suggestion
```

src/widgets/openmediavault/methods/apt_enumerateUpgraded.jsx
```diff
@@ -0,0 +1,37 @@
...
+import Block from "components/services/widget/block";
+
+const items = [
+  {
+    label: "openmediavault.updatesAvailable",
+    getValue: (data, t) => (data.length > 0 ? t("openmediavault.yes") : t("openmediavault.no")),
+  },
+];
+
+// noinspection DuplicatedCode
```
shamoon, line 16:
```suggestion
```

shamoon (changes requested):
Yea thats great, thanks.

Small thing is to remove these IDE comments. 

If youre ready you can mark this as such and Im happy to test it out / merge

shamoon:
@userXinos Im not clear how much demand there is for these options, of course lets please re-address if there is more interest from the discussion(s).

Thanks again

userXinos:
oh, alright

PR #1829, opened 2023-08-20
Calibre web widget

GAntoine:
## Proposed change

This PR adds support for a calibre-web widget. Since the calibre-web opds endpoint returns XML, I've had to write a custom proxy to parse the XML into JSON. While calibre-web does offer a JSON API, it's only accessible with a valid session cookie, and only meant for consumption from the browser app.

Documentation PR: https://github.com/benphelps/homepage-docs/pull/143

Closes #239 

Screenshot 2023-08-20 at 10 59 37 AM

## Type of change

- [x] New service widget
- [ ] Other (please explain)

## Checklist:


shamoon:
Thanks. Their api doesn’t include any kind of summary endpoint? As opposed to making multiple separate calls

Also, why exactly can’t we use this? https://github.com/calibreapp/cli

shamoon:
Ha sorry, wrong calibre! :)

GAntoine:
> Their api doesn’t include any kind of summary endpoint?

I don't believe so; their internal API is JSON-based but is only available to session-based auth (based on my reading of the code `Basic` headers aren't supported). The [OPDS](https://en.wikipedia.org/wiki/Open_Publication_Distribution_System) endpoints do accept `Basic` headers, but are XML based and limited in scope.

One thing I forgot the mention: OPSD responses are usually paginated (hence the `00` in each endpoint), but I don't have enough books to hit the pagination limit, and I don't see any explicit limit in their docs. If some users report hitting the pagination limit I can add support for supplemental fetching-the-next-page requests.

shamoon (approved):
Thanks, great. I initially got an error I suppose because my calibre install has no series (havent used it in years but it was cool that calibreweb just hooked up to it magically). I added optional chaining (`?`) for each of them just in case.

Screenshot 2023-08-21 at 9 19 32 AM

Screenshot 2023-08-21 at 9 17 45 AM

PR #1834, opened 2023-08-20
Round openweather temperature

Skkay:
## Proposed change

I don't think it's useful to have two-decimal precision for temperature, so here's a PR for rounding off the value to the nearest integer using `Math.round()`

Before:

After:

Closes #1833

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:


shamoon:
Thanks but:

- This is kinda just your opinion. If there is significant agreement on your discussion thread we can readdress
- This is not how you round for next-i18next, see https://www.i18next.com/translation-function/formatting#number

PR #1844, opened 2023-08-21
Add user and restart options to docker-compose

joaoportela:
Add user:group 1000:1000 to the docker compose, to match host user. Restart container unless stopped.

## Proposed change

Change the default user. When mounting files from the host, it's better to match the host user:group. Usually that is 1000:1000, which makes it a good default.

When deploying this container, the user usually want to keep the container running persistently, hence the restart unless stopped.

## Type of change

- [ ] New service widget
- [X] Other (please explain)

Change example docker-compose in README.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ services:
     volumes:
       - /path/to/config:/app/config # Make sure your local config directory exists
       - /var/run/docker.sock:/var/run/docker.sock:ro # (optional) For docker integrations
-    user: 1000:1000 # default user id
+    # user: 1000:1000 optional, not compatibile with direct socket see https://gethomepage.dev/en/configs/docker/#using-socket-directly
     restart: unless-stopped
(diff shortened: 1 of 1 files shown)


joaoportela:
I don't know if this affects the `/var/run/docker.sock:ro` mount, because I'm not using it.

shamoon:
I dont think this belongs in our compose, it is not always the case and I think we're asking for a lot of issues from people having problems

This will also break mounting the docker socket

I suppose if you feel strongly about the restart option we can include that

joaoportela:
> This will also break mounting the docker socket

I guess that's a deal breaker then... 😅 

I'll open one with the restart. I think it's pretty standard usage.

shamoon:
Actually, you could just comment out the `user` line with a (brief) explanation. That way users can think about it

PR #1875, opened 2023-08-28
Making both Containo and Io work regardless if they are both present or not

atropos112:
## Proposed change

Two parts here:
1. Do not pollute logs if list cluster object on traefik returns 404, they resources likely are simply not there and lack of their existence would bring host of other problems that would be far more "loud" than an error log in one pod.
2. Do not use fall back logic using try  catch, if both exist in cluster unexpected behaviour can occur.

P.S. Although I am a developer by day I rarely ever touch javascript so there is a chance i did something stupid, please double check my changes.

Closes https://github.com/benphelps/homepage/issues/1874

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -161,7 +161,7 @@ export async function servicesFromKubernetes() {
     const traefikIngressListContaino = await crd.listClusterCustomObject("traefik.containo.us", "v1alpha1", "ingressroutes")
       .then((response) => response.body)
       .catch(async (error) => {
-        if (error.statusCode != 404) {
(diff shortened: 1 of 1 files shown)


shamoon:
This project is definitely more docker-centric than k8s, seems k8s is the one that requires much more maintenance but I dont run k8s.

Change looks fine to me, thanks.

atropos112:
I get that, k8s is usually more complex and bespoke setups don't help.

Having said that, I think this is the best dashbord solution for homelab kubernetes setup, and I have looked for ages. Thank you for working on this project it is very much appreciated.

PR #1911, opened 2023-09-04
Enhancement: CustomAPI widget data formats

kayatf:
## Proposed change
I noticed that the customapi widget does not allow for data types boolean and 8-bit values (if you want to display them as a percentage). Therefore I added boolean format as "state" which either displays "Active" or "Inactive" and corresponding translation and percentBinary which converts 8-bit values ( ≤255) to a 0-100% range.

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -306,6 +306,10 @@
         "7days": "7 Days",
         "30days": "30 Days"
     },
+    "customapi": {
+      "active": "Active",
+      "inactive": "Inactive"
+    },
     "gotify": {
         "apps": "Applications",
         "clients": "Clients",
--- a/src/widgets/customapi/component.jsx
+++ b/src/widgets/customapi/component.jsx
@@ -1,4 +1,4 @@
-import { useTranslation } from "next-i18next";
+import {useTranslation} from "next-i18next";
 
 import Container from "components/services/widget/container";
 import Block from "components/services/widget/block";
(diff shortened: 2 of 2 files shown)


shamoon (changes requested):
Thanks. A few things:

- Please revert the whitespace changes, it just muddies the PR
- The `percentBinary` feels kinda random. Are there a lot of APIs that return an 8-bit number thats actually supposed to represent a percent? Its not that its so complicated to include here, but it feels a little bloated...
- As for `state`, first of all your regex will fail if an API returns e.g. 0/1 or even `True` so please use some kind of falsy test instead. Second, restricting this to "active" or "inactive" doesnt make sense to me as we're trying to make things as flexible as possible and a user may want a 'state' to represent "raining / not" or "temp is above 30deg / not" or _whatever_. Im not positive what the best solution would be, you could format the label like `label:valueiftrue:valueiffalse` or something like that, but again, should be more flexible I think

Let me know if you have any concerns / questions

benphelps:
> Thanks. A few things:
> 
> * Please revert the whitespace changes, it just muddies the PR
> * The `percentBinary` feels kinda random. Are there a lot of APIs that return an 8-bit number thats actually supposed to represent a percent? Its not that its so complicated to include here, but it feels a little bloated...
> * As for `state`, first of all your regex will fail if an API returns e.g. 0/1 or even `True` so please use some kind of falsy test instead. Second, restricting this to "active" or "inactive" doesnt make sense to me as we're trying to make things as flexible as possible and a user may want a 'state' to represent "raining / not" or "temp is above 30deg / not" or _whatever_. Im not positive what the best solution would be, you could format the label like `label:valueiftrue:valueiffalse` or something like that, but again, should be more flexible I think
> 
> Let me know if you have any concerns / questions

Perhaps a more generic approach, the full example:

```yaml
widget:
    type: customapi
    url: http://custom.api.host.or.ip:port/path/to/exact/api/endpoint
    mappings:
      - field: key # needs to be YAML string or object
        label: Field 1
        format:
          type: number
          options:
            style: percent
            precision: 2
```

And some more of just the format section, for a percent you'd like formated:

```yaml
format:
  number:
    style: percent
    precision: 2
```

or perhaps you want to format a currency:

```yaml
format:
  number:
    style: currency
    currency: USD
    maximumFractionDigits: 2
    roundingIncrement: 5
```

Here we mix both options, which could be very cool:

```yaml
format:
  map:
    from: [0, 255]
    to: [0, 100]
format:
  number:
    style: percent
    precision: 2
```

You could also have this option, to map values to other values:

```yaml
format:
  map:
    values:
      0: 'Off'
      1: 'On'
```

rauenzi:
How do you feel about the concept I laid out in this discussion https://github.com/benphelps/homepage/discussions/1870 to allow for template strings? It would allow users of the CustomAPI widget to make use of the built in formats like `bytes` or `bbytes` as well as the [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) by doing it directly in the template (e.g. `{{value, number(style: unit; unit: millisecond)}}`).

The only issue with using these directly rather than in JavaScript is there's no potential for scaling. Like if an API reports in MiB it won't work with `bytes` formatter without scaling (as I described in the discussion), or if an API reports a very large `millisecond` there's no way to scale that to minutes or hours or etc.

shamoon:
Thanks for this PR but Im going to close it in favor of #1958 which can do the same with more flexibility.

As mentioned above, there are other ideas to explore here.

Thanks again

PR #1918, opened 2023-09-06
link fixes

Shagon94:
## Proposed change

The change fixes the top right side of the service when using docker/ping. Previously the div would not have a URL so it wasn't clickable.

It also fixes the property name typo `crossorigin` => `crossOrigin`.

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/services/item.jsx
+++ b/src/components/services/item.jsx
@@ -76,7 +76,12 @@ export default function Item({ service, group }) {
             </div>
           )}
 
-          <div className="absolute top-0 right-0 w-1/2 flex flex-row justify-end gap-2 mr-2 z-30">
+          <a
+            href={service.href}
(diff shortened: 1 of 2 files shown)


shamoon:
Thanks but:

- Rather than creating another div we can just disable pointer events
- the attribute is [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)

Superseded by #1920

shamoon:
Update, youre right about the crossorigin, React complains about it even though it's not wrong. My bad. I added you to [the commit](https://github.com/benphelps/homepage/commit/ae3634e9d4791b08e288870d72da2c3d1356dd63)

PR #1950, opened 2023-09-09
Custom JS and CSS

TheRolfFR:
## Context

Hello @benphelps,
I have been experiencing with homepage, and the idea looks very promising.
I wanted some custom CSS, and even though the intent was not successful at first #904, I really wanted to give it a shot.

## Proposed changes
First, I started to add classes to the different information widgets. It is not easy as the structure itself involves modifying every single component. It's indeed very time consuming and __I have never used Next.js__ (I used many frameworks, but not this one).
However I think we could figure out something more general to make it faster. **Some help would be very appreciated**.

The use [CSS modules](https://nextjs.org/docs/app/building-your-application/styling/css-modules)  and [@apply directive](https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply) would also help reduce visual clutter and abusive use of the classNames module.

As you will see, it includes many new classes, generally starting with ``information-widget-`` to make it clear it's an information widget. A little long, but it can be changed as desired.
I also added several ids for the information widgets container, the right container part and the elements at the bottom of the page.

To stylize, I added 2 files into the skeleton folder (maybe they could be optional, what do you think?).
Then I created and API endpoint to read from the config folder any file basically.
I used SWR to make a new component that refreshes dynamically the file content changes (useful when experimenting styles).
I included new style and script html tags into the index page (I don't know where to place, it was sometimes not rendered)

## Service IDs 
I added the possibility to add an ID to the service HTML element, by setting the ``id`` attribute of the described service.
You can then use it to style or add JS if you want.

## Result possible
I use it to move the resource widget label, make the logo container grow, and adjust the spacing and alignment in general:

*Before*
image
*After*
image

## Document PR when solutions established
As this framework and ideas very subject to change, I will make a PR 

## Issues and discussions related
Please modify this but at least #139 

## Type of change

- [no] New service widget
- [no] Bug fix (non-breaking change which fixes an issue)
- [no] Other (please explain)

## Checklist:

- [x]  If adding a service widget or a change that requires it, I have added a corresponding PR to the [documentation](https://github.com/benphelps/homepage-docs) here: 

- [no] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/en/more/development/#service-widget-guidelines).
- [n/a] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,17 +1,5 @@
 {
     "files.exclude": {
-        "**/.git": true,
-        "**/.svn": true,
-        "**/.hg": true,
-        "**/CVS": true,
-        "**/.DS_Store": true,
-        "**/Thumbs.db": true,
-        "**/__pycache__": true,
-        "**/.idea": true,
-        "**/.classpath": true,
-        "**/.project": true,
-        "**/.settings": true,
-        "**/.factorypath": true,
         "**/.next": true,
         "**/node_modules": true
     }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,17 @@
 {
     "files.exclude": {
+        "**/.git": true,
+        "**/.svn": true,
+        "**/.hg": true,
+        "**/CVS": true,
+        "**/.DS_Store": true,
+        "**/Thumbs.db": true,
+        "**/__pycache__": true,
+        "**/.idea": true,
+        "**/.classpath": true,
+        "**/.project": true,
+        "**/.settings": true,
+        "**/.factorypath": true,
         "**/.next": true,
         "**/node_modules": true
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "js-yaml": "^4.1.0",
         "json-rpc-2.0": "^1.4.1",
         "memory-cache": "^0.2.0",
+        "mime": "^3.0.0",
         "minecraft-ping-js": "^1.0.2",
         "next": "^12.3.1",
         "next-i18next": "^12.0.1",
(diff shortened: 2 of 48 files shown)


shamoon (changes requested):
This is an interesting approach, thanks for the PR. Im guessing youve tested this in a docker image and it works? I wont be able to test until later but looking at it, the only downside of the approach I can imagine would be a delay in applying the custom styles but perhaps thats minor / not even noticeable.

I dont love all of the custom classes, it feels a bit anti-tailwind but obviously I see the purpose. And you havent added them for service widgets?

The requested changes are all minor things, regardless the PR is well done but Im curious what phelps things.

TheRolfFR:
> I dont love all of the custom classes (also inconsistent use of id vs class e.g. `information-widget-glances` should probably be an id), it feels a bit anti-tailwind but obviously I see the purpose. And you havent added them for service widgets?

I added a paragraph in the issue description and personnally I use multiple Glances instances

>  I can imagine would be a delay in applying the custom styles but perhaps thats minor / not even noticeable.

Default refresh time for the file content component is 1500ms, but it's an attribute that can be changed as desired

shamoon:
Yea I realized the multiples thing for classes vs ids, duh thanks

shamoon (approved):
Again, this is really clever and I think people will love it. Thanks for the PR

shamoon:
Sorry I did make one actual change which is instead of refreshing and wasting network calls every 1.5s I added these files to the hash which means:

- homepage will auto-detect changes
- the [empty] files will get created automatically

Let me know if you have any concerns with that

TheRolfFR:
Yes, sure, please do. You know the tech and codebase better.
I didn't want to touch others API routes because I didn't want to break anything ^^
Thanks for the doc by the way!

I added preloading to the pages to get them earlier, tell me what you think

shamoon:
Cool yea preload a good idea

Issue #1998, opened 2023-09-16
[Bug] Kubernetes service discovery via annotations faulty in 0.6.30


janpfischer:
### Description

Hey, 

after upgrading all my Kubernetes service annotation widget's and entries stopped working. Basically my homepage is empty and not filling with the services which have homepage annotations. 

I have no services which are not done with annotations.

Before upgrading from 0.6.29 everything was fine. 

### Steps to reproduce

1. Have services with homepage annotations in Kubernetes
2. Upgrade to 0.6.30
3. No services displayed on homepage

### homepage version

0.6.30

### Installation method

Other (please describe above)

### Configuration

```yaml
Example annotations:

gethomepage.dev/enabled: "true"
gethomepage.dev/description: A E-Book library
gethomepage.dev/group: Media
gethomepage.dev/icon: calibre-web.png
gethomepage.dev/name: Calibre Web
```

### Container Logs
```
[2023-09-16T12:44:23.313Z] error: <service-helpers> TypeError: Cannot read properties of undefined (reading 'length')
    at servicesFromKubernetes (/app/.next/server/chunks/7586.js:267:56)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async servicesResponse (/app/.next/server/chunks/5876.js:121:127)
    at async handler (/app/.next/server/pages/api/services.js:84:12)
    at async Object.apiResolver (/app/node_modules/.pnpm/next@12.3.4_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/api-utils/node.js:366:9)
    at async NextNodeServer.runApi (/app/node_modules/.pnpm/next@12.3.4_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/next-server.js:481:9)
    at async Object.fn (/app/node_modules/.pnpm/next@12.3.4_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/next-server.js:741:37)
    at async Router.execute (/app/node_modules/.pnpm/next@12.3.4_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/router.js:252:36)
```
### Browser Logs

none

### Troubleshooting

n/a

### Other

k3s v1.27.4+k3s1

### Before submitting, I have made sure to

- [X] Check [the documentation](https://gethomepage.dev/)
- [X] Follow [the troubleshooting guide](https://gethomepage.dev/en/more/troubleshooting/) (please include output above if applicable).
[...]

PR #2003, opened 2023-09-16
Kubernetes Check for Custom Resource Definitions

janpfischer:
## Proposed change

Within #1998 we discussed that the RBAC for the Kubernetes service account currently needs the following permission to not run into errors and spam the log. 
```
- apiGroups:
      - traefik.io
      - traefik.containo.us
    resources:
      - ingressroutes
    verbs:
      - get
      - list
```

This is needed even if you have no custom resource definitions of traefik. If you have this permission not in place the kubernetes-client in "service-helpers.js" will run into an error because you can not access the CRD. You have to add the permission to silence the alert but because you do not have the CRDs installed the next API call will fail too. Until now this was simply silenced with an statusCode != 404 if-clause. @benphelps  extended this today to exclude 403 errors which are generated if you do not have the above RBAC in place. 

With this change it is first checked if the CRDs are present. If they are not no further call is been made. If they are present they will be called. In case of an error every error will then be logged, because we do not have to exclude 404 or 403 here anymore. 

This change is some kind breaking because for this CRD check you will now need an new RBAC permission: 
```
  - apiGroups:
      - apiextensions.k8s.io
    resources:
      - customresourcedefinitions/status
    verbs:
      - get
```
If you do not have this in place a meaningful error message will be logged. 

If this change is considered to be merged I will put a MR to the unofficial helm chart to extent this RBAC permission and maybe exclude the traefik ones. 

This function can be used for more than the current traefik CRDs if there will be added more CRDs to the logic of service-helpers.js in the future. 

Happy to discuss this change!

## Type of change

- [ ] New service widget
- [x] Other (please explain)

## Checklist:

### DIFF
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -219,9 +219,9 @@ export async function servicesFromKubernetes() {
         return [];
       });
 
-    const traefikIngressList = [...traefikIngressListContaino, ...traefikIngressListIo];
+    const traefikIngressList = [...traefikIngressListContaino?.items ?? [], ...traefikIngressListIo?.items ?? []];
 
-    if (traefikIngressList?.items?.length > 0) {
+    if (traefikIngressList.length > 0) {
       const traefikServices = traefikIngressList.items.filter(
         (ingress) => ingress.metadata.annotations && ingress.metadata.annotations[`${ANNOTATION_BASE}/href`]
       );
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -4,7 +4,7 @@ import path from "path";
 import yaml from "js-yaml";
 import Docker from "dockerode";
 import * as shvl from "shvl";
-import { CustomObjectsApi, NetworkingV1Api } from "@kubernetes/client-node";
+import { CustomObjectsApi, NetworkingV1Api, ApiextensionsV1Api } from "@kubernetes/client-node";
 
 import createLogger from "utils/logger";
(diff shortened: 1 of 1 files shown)


shamoon:
I suppose this is fine but do you want to discuss the change at the helm chart first? If I understand correctly, if that were not merged this wouldn’t really work, no? 

And, as it is currently does this mean every k8s user will be greeted with that new error message? I realize now there’s an actual solution (instead of just ignoring errors) but just making sure

janpfischer:
I believe it would not break anything more than it is broken at the moment already because if currently the CRDs can't be found (does not matter if 403, 404 or something else) the return value is already an [empty array](https://github.com/benphelps/homepage/pull/2003/files#diff-eba25b09f3c7f5f245149ee19bf3e3d35bc8d82e0ee4eccf32130aea46e0600aR219) which means that the traefik resources are not evaluated for the homepage generation. 

What this will do if the RBAC is not added to the helm chart a new [log message](https://github.com/benphelps/homepage/pull/2003/files#diff-eba25b09f3c7f5f245149ee19bf3e3d35bc8d82e0ee4eccf32130aea46e0600aR152) will appear, but the misleading ones with traefik will not be shown because of the if-clauses (like [here](https://github.com/benphelps/homepage/pull/2003/files#diff-eba25b09f3c7f5f245149ee19bf3e3d35bc8d82e0ee4eccf32130aea46e0600aR194).

I've opened a MR in the helm chart repo -> https://github.com/jameswynn/helm-charts/pull/15

Issue #2056, opened 2023-09-24
[Bug] TypeError: Cannot read properties of undefined (reading 'drive')


rayansailani:
### Description

When starting the application in local with `pnpm dev` it was seen that while rendering the homepage, I encountered an error 
TypeError: Cannot read properties of undefined (reading 'drive'). |
![image](https://github.com/benphelps/homepage/assets/54993789/a91b8a19-a37b-44a9-89b1-17dae1bab109)

### Steps to reproduce

git clone 
pnpm install
pnpm dev

### homepage version

v0.6.35

### Installation method

Source

### Configuration

```yaml
n/a
```

### Container Logs

TypeError: Cannot read properties of undefined (reading 'drive')

### Browser Logs

TypeError: Cannot read properties of undefined (reading 'drive')

### Troubleshooting

TypeError: Cannot read properties of undefined (reading 'drive')

### Other

n/a

### Before submitting, I have made sure to

- [X] Check [the documentation](https://gethomepage.dev/)
- [X] Follow [the troubleshooting guide](https://gethomepage.dev/en/more/troubleshooting/) (please include output above if applicable).
- [X] Search [existing issues](https://github.com/benphelps/homepage/search?q=&type=issues) and [discussions](https://github.com/benphelps/homepage/search?q=&type=discussions).

rayansailani:
I would like to take this up as it seems like a simple empty object check that could be fixed by using lodash.

rayansailani:
https://github.com/benphelps/homepage/pull/2057 - have PR for this issue.

shamoon:
This code hasn't changed in ages and it's not something reported a lot, rather than adding random checks, I'd like to figure out why it's happening first. If you've incorrectly specified a drive or the si dependency is returning empty data or what.

And re your second PR, what is the even-more specific situation where it would return data.drive but drive is actually empty??

PR #2057, opened 2023-09-24
fix-#2056:Added null checking using lodash

rayansailani:
## Proposed change
Added null/undefined/empty object check using lodash library to prevent error in #2056 

Closes (#2055)

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "i18next": "^21.9.2",
     "js-yaml": "^4.1.0",
     "json-rpc-2.0": "^1.4.1",
+    "lodash": "^4.17.21",
     "memory-cache": "^0.2.0",
     "minecraft-ping-js": "^1.0.2",
     "next": "^12.3.1",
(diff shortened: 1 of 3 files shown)


shamoon:
Thanks but we're not going to add an entire dependency for a 1 line fix. It also wasn't added correctly to npm lock file.

Im not clear under what circumstances this situation happens, but vanilla javascript will do just fine.

Thanks anyway

rayansailani:
Okay, can I look into fixing it with Vanilla JS ?

shamoon:
I took care of it mate https://github.com/benphelps/homepage/commit/6483b10ebabe6803be2b8bd5fe036b7d1983614e

sorry I’m impatient

Since you’re running from source can you tell why this would even happen? I can’t reproduce it

rayansailani:
That's great! I was looking into why this is happening at a deeper level. But I feel checking it like this would be better 
`if (!data || Object.keys(data).length === 0)` reason being `drive.available` could also throw an error at some point of time.

PR #2068, opened 2023-09-26
Feature: Added detect octoprint offline printer

batuhan0sanli:
## Proposed change

I found a way to detect when the 3D printer is not connected, allowing us to get rid of that annoying red box.

Closes #1841
See #834 

**Before:**
Ekran Resmi 2023-09-26 12 39 49

**After:**
Ekran Resmi 2023-09-26 13 06 25

**Solution:**
The `job` endpoint returns 200 even when the printer is not connected. It also provides information that the printer is Offline. Here's an example `job` endpoint's response when the printer is not connected:

```
{
    "job": {
        "averagePrintTime": null,
        "estimatedPrintTime": null,
        "filament": null,
        "file": {
            "date": null,
            "display": null,
            "name": null,
            "origin": null,
            "path": null,
            "size": null
        },
        "lastPrintTime": null,
        "user": null
    },
    "progress": {
        "completion": null,
        "filepos": null,
        "printTime": null,
        "printTimeLeft": null,
        "printTimeLeftOrigin": null
    },
    "state": "Offline"
}
```

## Type of change

- [ ] New service widget
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/octoprint/component.jsx
+++ b/src/widgets/octoprint/component.jsx
@@ -8,9 +8,9 @@ export default function Component({ service }) {
   const { data: printerStats, error: printerStatsError } = useWidgetAPI(widget, "printer_stats");
   const { data: jobStats, error: jobStatsError } = useWidgetAPI(widget, "job_stats");
 
-  if (printerStatsError && jobStats && jobStats.state==="Offline") {
+  if (printerStatsError && jobStats) {
     return <Container service={service}>
-        <Block label="octoprint.printer_state" value="Not Operational" />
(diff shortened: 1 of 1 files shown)


batuhan0sanli:
I have dynamically updated value of the Block. Octoprint has indicated that the specified endpoint provides detailed information for the "state" value, such as `Error`, `Offline`, `Offline after error`, `Opening serial connection`, etc. I found to be potentially useful. See: https://docs.octoprint.org/en/master/api/job.html#job-information-response

PR #2134, opened 2023-10-05
Fix: Service without a name breaks homepage when using quicklaunch

denispapec:
## Proposed change

As per title, following configuration breaks homepage 

```yaml
- Calendar:
    - Calendar:
        name:
        widget:
          type: calendar
```

with following error:
> Unhandled Runtime Error
> TypeError: can't access property "toLowerCase", r.name is null

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/quicklaunch.jsx
+++ b/src/components/quicklaunch.jsx
@@ -84,7 +84,7 @@ export default function QuickLaunch({servicesAndBookmarks, searchString, setSear
     if (searchString.length === 0) setResults([]);
     else {
       let newResults = servicesAndBookmarks.filter(r => {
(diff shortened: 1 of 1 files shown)


shamoon:
I mean, Is this something we're particularly keen to support in general? And I assume '' works

denispapec:
It is valid YAML. 
" works. I've tried to use it as empty and then noticed that, haha :see_no_evil:

shamoon:
I mean yea it's valid YAML, just not sure we should be recommending this anyway. You can hide headers etc. already

If theres a lot of push for this we can readdress. Obviously its a simple change, more of a philosophical thing here 😂

PR #2135, opened 2023-10-05
Fix: Quicklaunch duplicating services in the list

denispapec:
## Proposed change

When using quicklaunch, and deleting some letters and writing again; it starts to duplicate the same services in the list (as the UL object reuses the same key).

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/quicklaunch.jsx
+++ b/src/components/quicklaunch.jsx
@@ -152,6 +152,11 @@ export default function QuickLaunch({servicesAndBookmarks, searchString, setSear
     return <span>{parts.map((part, i) => part.toLowerCase() === searchString.toLowerCase() ? <span key={`${searchString}_${i}`} className="bg-theme-300/10">{part}</span> : part)}</span>;
   }
 
+  function getRandomKeyId() {
+    // as these are dynamically generated elements, we can generate random keys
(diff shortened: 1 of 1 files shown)


shamoon:
https://github.com/gethomepage/homepage/assets/4887959/0d0ff60e-01ba-4fd0-a3d1-7a36aeaf6405

Again, I cant reproduce this. Also doesnt make sense to me the key should be on the ul but rather the li's, which already have keys

denispapec:
> Again, I cant reproduce this.
Ah, another clue is that service needs to be added multiple times on homepage, for example: multiple PiHole widgets, multiple Portainer widgets, multiple Watchtower widgets....

> Also doesnt make sense to me the key should be on the ul but rather the li's, which already have keys
Yeah, that's what I went to fix at first, but adding key to UL seems to resolve this issue.

shamoon:
I think that only happens if they have the same name, which is explicitly something we discourage.


https://github.com/gethomepage/homepage/assets/4887959/8802f4c6-c526-41ea-b4b6-cd8af9f9fb15

Btw, using random numbers for react keys is definitely not best practice

Let me know if Im missing anything here. Thanks anyway.

Issue #2182, opened 2023-10-13
[Bug] Status running / metrics not working with traefik autodiscovery on k8s


hugosxm:
### Description

Hi, 

I use homepage with the kubernetes autodiscovery for ingressroutes ( the traefik one, not the vanilla ingress ) 

It works as my service appear on homepage BUT i do not have the green "running" little square to the top right of the card and if I click on it it show OFFLINE.

See the screenshot 
![homepagebug](https://github.com/gethomepage/homepage/assets/23743455/dc2e0545-61a4-48d2-8720-8d507eb93338)

I hope this is not me missing something obvious...

### Steps to reproduce

Create ingress route with annotations like this : 

  ```
annotations:
    gethomepage.dev/href: "https://tools.xxx.com"
    gethomepage.dev/ping: "https://tools.xxx.com"
    gethomepage.dev/enabled: "true"
    gethomepage.dev/description: Various tools for it
    gethomepage.dev/group: Prod services
    gethomepage.dev/icon: https://tools.xxx.com/favicon-32x32.png
    gethomepage.dev/name: It tools

    # I tried with and without this, multiple values but no success
    gethomepage.dev/podSelector: "app.kubernetes.io/name: ittools-app=ittools-app"
```

### homepage version

v0.7.3

### Installation method

Docker

### Configuration

```yaml
Nothing special in my kubernetes.yaml

  kubernetes.yaml: |
    mode: cluster
```

### Container Logs

Nothing in the container log 

```
[2023-10-12T17:58:50.884Z] warn: API response for /api/services/proxy?type=prometheus&group=K8S+services&service=Prometheus&endpoint=targets exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit
[2023-10-12T18:02:12.887Z] warn: API response for /api/services/proxy?type=prometheus&group=K8S+services&service=Prometheus&endpoint=targets exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit
[2023-10-12T18:02:16.073Z] warn: API response for /api/services/proxy?type=prometheus&group=K8S+services&service=Prometheus&endpoint=targets exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit
[...]

PR #2198, opened 2023-10-16
Enhancement: Support k8s `app` IngressRoute annotation

shamoon:
## Proposed change

See #2182

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [x] Other (please explain): enhancement

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added corresponding documentation changes.
- [x] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).

### DIFF
--- a/docs/configs/kubernetes.md
+++ b/docs/configs/kubernetes.md
@@ -110,6 +110,7 @@ metadata:
         gethomepage.dev/description: Media Server
         gethomepage.dev/group: Media
         gethomepage.dev/icon: emby.png
+        gethomepage.dev/app: emby-app # optional, may be needed if app.kubernetes.io/name != ingress metadata.name
         gethomepage.dev/name: Emby
         gethomepage.dev/widget.type: "emby"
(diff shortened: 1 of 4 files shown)


hugosxm:
I see that the docker image is built with the pull request content in it, can I access that particular docker image to test ? If yes with which url ?

shamoon:
I dont think so since its not pushed to a registry (we should set that up tho), so I think you'd have to build locally

shamoon:
@hugosxm you can try out in `main`, its a pretty safe change

hugosxm:
@shamoon I can confirm to you, just tested, it works exactly as expected !

I tested both case, without the app label, it works if the name of the ingress is the same ( as before the PR ), with the app label and different ingress name, works too as i wanted :)

Thanks a lot for this little PR !

I'll keep using nightly in the meantime ^^

PR #2215, opened 2023-10-19
Feature: true ping, rename old ping to siteMonitor

shamoon:
## Proposed change

I've always been a little annoyed (at myself) that ping isn't a real ping and as noted in the linked request, at least some people want it. So I'm proposing (or at least considering) that we:

- Make ping true ping, using a tiny wrapper module for system `ping`. AFAIK almost all systems will have this available. Options may differ across platforms but we're using the absolute basics.
- Rename the current ping functionality to 'site monitor' (`siteMonitor`) since thats more actually what it is.

This is _kind of_ breaking:

- Current users will now use true `ping`, that said, the implementation here does still accept the URL, so it should work but results will be different than before.

This isn't like, the most important thing to me in the world, but I think it's 'worth' the hassle. But open to other thoughts of course.

Closes #2141

Screenshot 2023-10-19 at 2 36 13 PM

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If adding a service widget or a change that requires it, I have added corresponding documentation changes.
- [x] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).

### DIFF
--- a/docs/configs/services.md
+++ b/docs/configs/services.md
@@ -101,29 +101,49 @@ To use a local icon, first create a Docker mount to `/app/public/icons` and then
 
 ## Ping
 
-Services may have an optional `ping` property that allows you to monitor the availability of an endpoint you chose and have the response time displayed. You do not need to set your ping URL equal to your href URL.
-
-!!! note
-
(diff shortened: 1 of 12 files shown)


shamoon:
Whoops, wrong commit

benphelps (approved):
I like this change 👍🏻 

Pings are a little weird in the devcontainer but work just fine locally and from the Docker image.

PR #2254, opened 2023-10-29
Update nextcloud.md

dhrions:
Add precisions on how get the key or a specific way to get dedicated username and password.

- [ ] New service widget
- [X] Documentation only
- [ ] Other (please explain)


shamoon:
Thanks anyway but the faux domains aren’t all that helpful nor are things like “bottom of the page”. In general we try to keep these instructions simple as occasionally things change in the relevant app.

PR #2261, opened 2023-10-30
Feature: iFrame widget

Reiss-Cashmore:
## Proposed change

This adds a basic iframe component with customisability for height and some other basic iframe parameters. This widget does deviate from the design guidelines. However I think by the very nature of the Widget it is justifiable.

I am not entirely sure my changes within service-helpers.js are really inline with the projects standards and would appreciate feedback. The component communication flow and organisation of HoCs is....not like anything I've ever come across before and is diffuclt to decipher so if I have not done something in the style of the project please point it out and I'll update.

There is unfortunately no decent way of providing fallback content if the iframe fails to load so right now you just get the default browser response.

image

Closes # 
Resolves
https://github.com/gethomepage/homepage/discussions/1227

## Type of change

- [X] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [X] If adding a service widget or a change that requires it, I have added corresponding documentation changes.
- [X] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).
- [X] I have checked that all code style checks pass using pre-commit hooks and linting checks with `pnpm lint` (see development guidelines).


justaCasualCoder (approved):
Looks Awesome!

Reiss-Cashmore:
Might consider adding some zoom and transform functionality like this afterwards too. Better allow for positioning etc

https://www.w3docs.com/snippets/css/how-to-scale-the-content-of-iframe-element.html

docs/widgets/services/iframe.md
shamoon, line 1: The docs need to be more similar to our other docs, none of them are like this.

src/widgets/iframe/components/block.jsx
shamoon, line 1: No need for a separate component if you use it exactly 1 time

src/widgets/iframe/component.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import Container from "components/services/widget/container";
+
+export default function Component({ service }) {
+  const [refreshTimer, setRefreshTimer] = useState(0);
+
+  const { widget } = service;
+
+  useEffect(() => {
+    if (widget?.refreshInterval) {
```
shamoon, line 13: This needs to be constrained to a minimum of 1s

src/widgets/iframe/component.jsx
```diff
@@ -0,0 +1,52 @@
...
+      const refreshInterval = setInterval(() => {
+        setRefreshTimer(refreshTimer + 1);
+      }, widget?.refreshInterval);
+      return () => clearInterval(refreshInterval);
+    }
+    return undefined;
```
shamoon, line 19: useEffect doesnt need to return anything.

src/widgets/iframe/component.jsx
```diff
@@ -0,0 +1,52 @@
...
+  }, [refreshTimer, widget?.refreshInterval]);
+
+  const scrollingDisableStyle = widget.scrolling
```
shamoon, line 22: This cant work, widget.scrolling isn't set anywhere

shamoon (changes requested):
Thanks for the work. I know it's been requested, but holy cow it's ugly. Are there any use cases where it isn't? Realize that's not your fault, but in general it's not really in line with homepage's POV to support this, we're more about them APIs.

In the meantime, I've noted some code issues, up to you to take care of them now or wait for a final decision about the PR overall.

src/widgets/iframe/component.jsx
Reiss-Cashmore: > useEffect doesnt need to return anything.

Causes a lint error, happy to take your advice on whether thats ok

src/widgets/iframe/components/block.jsx
Reiss-Cashmore: See what you think of my refactor, I've kept the <Block> Component function but moved it inside the Widget file. I think keeping the main widget with the` <Container><Block>~~~Snip` semantic html helps to show that the intention that I'm trying to be consistent with the other widgets Container -> Block grouping for but that I needed a variant with a children passthrough

docs/widgets/services/iframe.md
Reiss-Cashmore: I'll be honest, a lot of the docs confuse me.

image

I have absolutely no idea what "Allowed Fields: ["Etc", "Etc"]" is trying to describe here? I genuinely tried to find other components to get an idea of what this is referring to for a good hour or two last night. I still don't understand it.

I can rewrite my descriptions in to the yaml block at the bottom like this:

https://gethomepage.dev/v0.7.4/widgets/services/calendar/

The custom API Docs are more descriptive

https://gethomepage.dev/v0.7.4/widgets/services/customapi/

But again this whole section means very little to me right now,

image


So you'll have to be quite prescriptive with how you would like me to refactor the docs but I am struggling to intuitively get them inline with whats already around at the moment

shamoon: Allowed fields won’t apply here, it refers to homepage’s ability to filter out blocks (fields)

in your case id include a small description and a link to MDN, then in the example yaml just small comments to explain each parameter.

as is it’s a lot of regurgitation of MDN docs since after all they’re just html props

Reiss-Cashmore:
Some rounding on the iframe element might help a little? Past that I can only think of maybe an optional subtle drop shadow perhaps? Let me know if you have some specific ideas

shamoon:
Yea those are cool. 

I think it’ll be OK, just want to confer with my associate. We’ve always been pretty strict about styling, but ultimately yea, not much to do about it here. Anyway, I suppose beauty is in the eye of the [self-hosted] beholder...

shamoon:
Yes, had that concern and good point, it's probably worth noting that these requests won't be proxied (yes it's obvious if you think about it but...). I do think the built-in browser protections for iframes mitigate the problem somewhat

Reiss-Cashmore:
Updated the docs

P.s what do you think about the way I did the Sizes config?

image

Is doing an object map like that in YAML okay? I couldn't find any Widgets that were configurable with a complex object like this. Is this how you guys would do it?

shamoon:
Hmm, re the sizes thing, Im not sure how intuitive it is, like its not obvious if a user needs to supply that entire map, and overall I think its just too much. I agree fine-grained control is nice but it's at the cost of simplicity. My inclination would be to use a single value, especially since even with this control it's just never going to perfectly align, I think people will have to accept that.

shamoon:
Also, please use pre-commit as noted in the guidelines https://gethomepage.dev/v0.7.4/more/development/#code-formatting-with-pre-commit-hooks

Reiss-Cashmore:
> Hmm, re the sizes thing, Im not sure how intuitive it is, like its not obvious if a user needs to supply that entire map, and overall I think its just too much. I agree fine-grained control is nice but it's at the cost of simplicity. My inclination would be to use a single value, especially since even with this control it's just never going to perfectly align, I think people will have to accept that.

You know what, I think you're right.

I have an idea that is even simpler

Reiss-Cashmore:
> Hmm, re the sizes thing, Im not sure how intuitive it is, like its not obvious if a user needs to supply that entire map, and overall I think its just too much. I agree fine-grained control is nice but it's at the cost of simplicity. My inclination would be to use a single value, especially since even with this control it's just never going to perfectly align, I think people will have to accept that.

Let me know what you think of the refactored approach. Simplified to just allow user to specify tailwind classes on the iframe element

shamoon:
Yea this is better, pared it down even more—K.I.S.!

shamoon:
Well, here we are. LMK if you have any concerns.

Reiss-Cashmore:
Looks good to me, I was erring on the side of making w-full and rounded optional just cause you never know what people might need and destyling can be annoying. This will work for my use cases so if you are happy so am I!

shamoon (approved):

adm2k:
@Reiss-Cashmore 

Greetings, Since there currently is no no widget for NUT UPS monitor I am trying to use the iFrame and get it looking close to the others. I have been trying to solve this on my own without success. How can I change the background color of the iFrame widget? I am trying to get to show in the "lighter" slate color - see the UptimeKuma in the photo below. I belive the code is 2A3445.

image


Since you are the creator of the iFrame widget I was hoping you could answer. Any help is greatly appreciated!

Reiss-Cashmore:
> @Reiss-Cashmore 
> 
> 
> 
> Greetings, Since there currently is no no widget for NUT UPS monitor I am trying to use the iFrame and get it looking close to the others. I have been trying to solve this on my own without success. How can I change the background color of the iFrame widget? I am trying to get to show in the "lighter" slate color - see the UptimeKuma in the photo below. I belive the code is 2A3445.
> 
> 
> 
> image
> 
> 
> 
> 
> 
> Since you are the creator of the iFrame widget I was hoping you could answer. Any help is greatly appreciated!

I'm out at the moment so can't give a full response to this and the other thread I've been tagged in. You want to select the css on the parent container of the iframe. That's what has the theme background in. Which looking at the template, won't be very simple to do. It has generic theme styles applied like bg-theme-200/50 on it which need overriding. Selecting parent CSS from a child is fragile. The iframe itself shouldn't have any background applied as far as I remember. 

I did want to make the classes applied more explicit but I don't think that's the style they 

FYI @shamoon this is the exact sort of reason why I broke out the w-full and rounding classes and made them explicitly overridable. Cause you just never know what people might want to do and destyling this after the fact will lead to people creating really fragile setups likely to break with updates. Should I bring that back and include the container elements classes as an option too?

adm2k:
Thanks for the response! I am hoping a remedy can be implemented as I suspect it may be awhile before we see a service widget for NUT UPS tools (https://networkupstools.org/). As of now I am just referencing an internal webpage that NUT provides and then using the iFrame widget to display the data.  I had to do a bunch of tweaking of the html page to get the four main pieces of data. I am attempting to keep the "look and feel" of  other widgets, so getting the iFrame widget to use the background color of A23445 (slate) should solve the issue for me. I can tweak the tables in the html page and adjust the cell colors to match other widgets.

All this could be avoided if we could get a NUT widget :)  but that is beyond my capabilities. I know the NUT UPS tool is used by many individual and organizations to monitor their UPS backup batteries etc.

Thank you!!!

shamoon:
It’s not reasonable to expect that we can satisfy every users’ specific needs, they should be balanced with overall ease-of-use, stability, maintenance etc.

I don’t think there’s anything particularly “fragile” about using CSS to override any styles. Frankly this is kind of an argument against iframe itself, it’s “messier” because we don’t have much control. I’m not actually suggesting we remove it or anything like that but I don’t see any issue leaving the onus for further customization to the user themselves.

also, in general I think folks should really create separate discussions for these separate questions. No one except us maintainers and the PR authors (and anyone subscribed to the PR who likely is not interested in these questions) see these random comments on closed pull requests so asking here means they don’t benefit from any possible wider community input.

Reiss-Cashmore:
> It’s not reasonable to expect that we can satisfy every users’ specific needs, they should be balanced with overall ease-of-use, stability, maintenance etc.

I am in agreement with you 100%! Personally I don't think anything suggested here goes against those principles. It's a very minimal change code wise. Giving the user the ability to _control_ all the classes, _optionally_, if they choose to. Thus giving them the tools to cater to their own needs.  Anyone who doesn't need to, won't need to do any extra config or even worry about the existence of class options.

Maintenance and reliability wise, it is a very minimal penalty. Yes it's a couple of extra lines, granted. But the actual functionality here is literally just passing values along a chain or using a default is nothing provided. 

The fragile part, was in reference to the kind of CSS this person will have to write, in order to do something simple like remove the background of the Container. As far as I can tell the below is the only way to do something as simple as remove the Block components background

FYI @adm2k this is probably the piece of CSS you want for now. Replacing "MyServiceName" with the value you set as "id" in your Service config as shown here https://gethomepage.dev/v0.8.0/configs/custom-css-js/
```
#MyServiceNameID > div > div:nth-child(2) > div {
  background: none;
}
```
I'm guessing this is pretty obvious why having users write XPath based CSS selectors is not great.
Just incase it isn't that obvious. This is doubly fragile, if the user ever updates their service name / id, this will need updating in their custom styles. Which at worst is annoying. But if a maintainer ever changes the template of the iFrame widget introducing a new div or changes one of the element tags in the chain this will break a users custom style pretty easily.

Alternatively an easy compromise would just be to add id's to the various child elements in the component and pass through the service name into the id so that they are uniquely selectable. This seems to be within the existing conventions of the project, however, I believe by far this is much worse for the "ease-of-use" principle you outlined, versus making all of the styles for both the iFrame and it's container overridable.

I think opening up this configuration and moving the classes to all be overridable would go a long way to making sure the project isn't pestered by future issues about styling the iFrame widget. I don't want the iFrame to become one of those features the maintainers despise because it results in the most "noise" .

shamoon:
I mean, Im not 100% certain what he's trying to style here but the 'darker' container around the iframe itself has a 'service-block' class, so it would literally just be
```
#MyServiceNameID .service-block {
    background: none;
}
```

Reiss-Cashmore:
> I mean, Im not 100% certain what he's trying to style here but the 'darker' container around the iframe itself has a 'service-block' class, so it would literally just be
> 
> ```
> #MyServiceNameID .service-block {
>     background: none;
> }
> ```
> 
> Screenshot 2023-11-18 at 6 49 04 PM

Ahh yep you are right that would be easier. Albeit specificity might start coming into play if they take this too far. I do think some of what I was saying still stands, I made this dicussion so we can move away from the PRs

https://github.com/gethomepage/homepage/discussions/2342

PR #2354, opened 2023-11-20
Adjust german translations and add inlang for a better inline translation overview

jannesblobel:
## Proposed change

I adjusted some German translations cause they looked like typical auto-translations and even weren't correct. Additionally, I added the inlang extension to see the missing translations and keys inline. 

Editor
A live instance of the editor (which I used for the contribution) can be previewed the following link:
https://inlang.com/editor/github.com/jannesblobel/homepage

and unfortunately, there are still 73 Translations are missing
[inlang status badge](https://inlang.com/editor/github.com/jannesblobel/homepage?ref=badge)
## Type of change

## Checklist:


shamoon:
Thanks for this but all translation is managed through crowdin, see https://crowdin.com/project/gethomepage

which I also think obviates the need for the inlang extension (which is cool btw).

I hope you’re willing to contribute over there, thanks

jannesblobel:
Thanks for your quick reply, 
Hmm, I see the point. The reason why I don't use Crowdin, I have to create a new account, get spam emails, and can't contribute/adjust inline :/

CleanShot 2023-11-20 at 17 36 12@2x

shamoon:
Well I guess that's why it's your call if you'd like to participate or not.

PR #2376, opened 2023-11-25
Feature: Implement iCal integration for calendar/agenda, improve calendar/agenda styling

denispapec:
## Proposed change

This PR implements iCal integration for calendar/agenda. 
Having iCal events in calendar/agenda, new configuration setting was added to show time in a day for events happening today.
During the implementation some styling improvements were made to bring styling consistency in line with other widgets, and event line was simplified and is now shared between calendar and agenda, having the same functionality.

Closes #1294 
Closes #2263 

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [x] Other (please explain): New calendar/agenda integration

## Checklist:


shamoon (commented):
Nice work, in general this works well. To be honest, Im a little worried about the potential for numerous little FRs given how particular calendar stuff can be, but lets hope not and we'll aim to keep this dead-simple.

I like the refactoring and getting rid of the context provider (I removed the actual file)

My only significant-ish issue was with the changes to the generic proxy handler, that felt kind of 'shoehorned'. I also didnt like that there was a calendar 'widget' and ical 'widget that seemed a bit redundant. Instead I created a generic proxy for the calendar component and linked ical to that. It keeps the code tidy and the integration stuff out of generic proxy handler, which may also prove helpful going forward.

Other stuff was minor, such as adding a little error if no events are found (presumably an error, like if you use google.com as your ical URL).

As always, let me know your thoughts.

Screenshot 2023-11-24 at 11 32 26 PM

src/widgets/calendar/integrations/ical.jsx
```diff
@@ -10,13 +10,21 @@ export default function Integration({ config, params, setEvents, hideErrors }) {
...
   });
 
+  let parsedIcal;
+
+  if (!icalError && icalData && !icalData.error) {
+    parsedIcal = parseString(icalData.data);
+    if (parsedIcal.events.length === 0) {
+      icalData.error = { message: "No events found" };
+    }
+  }
```
denispapec, line 20: Seems like this change causes continuous re-rendering. Reverting this change back solves this issue. Is there maybe a different way of sorting this one?

denispapec:
In the beginning I did create new proxy handler, but seeing how minimal changes are needed for generic handler, I've decided to extend it,  haha :sweat_smile: . It does make more sense having a separate handler :+1: .

I can see few more simple changes that can be added: 
- translation for `No events found`
- add ical name (`config.name`) into the error message, so in the error response it can say: `API error: ical (google): No events found.` 

Let me know if you want me to make changes for these two suggestions.

shamoon:
Thanks, good catch, didnt meant to put that outside the useEffect hook =)

Good suggestions too, added those. Let me know if you have any other concerns / thoughts.

denispapec:
Nice, thanks.
LGTM

shamoon (approved):
Thanks again

PR #2401, opened 2023-11-29
Change: changedetection.io widget count diff if not viewed

xconverge:
## Proposed change

I would consider this non-breaking, and non-contentious, but I am open to other critiques/options

I think this makes the widget match the functionality of changedetection and what is shown on the changedetection site a bit better...

Prior to this change, the widget could display a count of diffs detected that did not match "new unseen" diffs so when you navigated to the changedetection site, they had already been seen and were not bolded. The state would not "reset" until the following check happened.

More background at https://github.com/dgtlmoon/changedetection.io/issues/1998

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/latest/more/development/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/latest/more/development/#code-linting).

### DIFF
--- a/src/widgets/changedetectionio/component.jsx
+++ b/src/widgets/changedetectionio/component.jsx
@@ -28,7 +28,7 @@ export default function Component({ service }) {
   let diffsDetected = 0;
 
   Object.keys(data).forEach((key) => {
-    if (data[key].last_changed > 0 && data[key].last_checked === data[key].last_changed) {
+    if (data[key].last_changed > 0 && data[key].last_checked === data[key].last_changed && !data[key].viewed) {
       diffsDetected += 1;
     }
   });


xconverge:
@dgtlmoon FYI

shamoon:
I'll be honest, I dont care. Ha, no I just mean I dont use this and cant say I feel strongly. Since theres no way to poll the users who do, I think it's fine.

The only question I have, reviewing it a bit (but if Im honest not spending too much time understanding the details), is to make sure that the new `viewed` is being used correctly. You said in the linked thread "if it is set", but this PR doesnt actually check that currently. `!data[key].viewed` means that if `viewed` is false _or undefined_ it won't count that diff, is that correct? As opposed to eg `data[key].viewed === false` which would still count the diff if `viewed` isnt actually set (again, not sure if thats a possibility now)

xconverge:
Close but you have it backwards

If viewed === true, we dont count the diff

for all other scenarios (it is explicitly false, or the user is using an older version of changedetection where it is undefined), the behavior is unchanged and we would count the diff

shamoon:
In that case, great

shamoon (approved):

xconverge:
@dgtlmoon Can you confirm if something else changed with the API changes you made recently?

I have a response from the API that looks like this, and the diff is still bold/unviewed in the changedetection application

```
  "13efee39-b02e-4530-ac5f-700d5d4b3da3":
    {
      "last_changed": 1702397751,
      "last_checked": 1702419350,
      "last_error": false,
      "title": "OPNSense packages",
      "url": "https://pkg.opnsense.org/FreeBSD:13:amd64/23.7/latest/All/",
      "viewed": false,
    },
```

so the code to check for changes here doesnt quite make sense to me and how it worked before, unless the functionality changed?

```
    if (data[key].last_changed > 0 && data[key].last_checked === data[key].last_changed && !data[key].viewed) {
      diffsDetected += 1;
    }
```

last_changed and last_checked are not equal, so I think I need to make another code change here to remove that check?

dgtlmoon:
@xconverge thats correct, you should only need to count `viewed=false` as per your request earlier

> Close but you have it backwards
> 
> If viewed === true, we dont count the diff
> 
> for all other scenarios (it is explicitly false, or the user is using an older version of changedetection where it is undefined), the behavior is unchanged and we would count the diff

PR #2429, opened 2023-12-04
Enhancement: improve fritzbox proxy perfomance

evorg:
- only request the endpoints required by the field configuration
- note in the documentation on the use of http instead of https, as it is significantly faster (~1.2 seconds => ~90 milliseconds)

## Proposed change

Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/fritzbox.md
+++ b/docs/widgets/services/fritzbox.md
@@ -11,7 +11,7 @@ Home Network > Network > Network Settings > Access Settings in the Home Network
 [x] Transmit status information over UPnP
 ```
 
-You don't need to provide any credentials. Since that, you should consider using `http` instead of `https` as the requests are significantly faster.
+Credentials are not needed and, as such, you may want to consider using `http` instead of `https` as those requests are significantly faster.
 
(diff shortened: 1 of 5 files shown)


shamoon (changes requested):
Again, I have no way to test this, but let's assume you have =)

Two small things:

1. Please revert the German translation change here, it should done via crowdin
2. regarding the https change, I believe fritzbox is a local router, in which case I could see this being fine but why force the change? Why not let users chose in the way it currently does?

docs/widgets/services/fritzbox.md
```diff
@@ -11,12 +11,12 @@ Home Network > Network > Network Settings > Access Settings in the Home Network
 [x] Transmit status information over UPnP
 ```
 
-You don't need to provide any credentials.
+You don't need to provide any credentials. Since that, you should consider using `http` instead of `https` as the requests are significantly faster.
```
shamoon, line 14:
```suggestion
Credentials are not needed and, as such, you may want to consider using `http` instead of `https` as those requests are significantly faster.
```

shamoon:
Yea I think the comment in the docs is a good idea, lets users know why to consider it. I just made a suggestion about the way its written =)

shamoon (approved):

PR #2284, opened 2023-11-08
Fix: handle immich v1.85.0 breaking change

shamoon:
## Proposed change

Closes #2282 

Screenshot 2023-11-07 at 11 58 10 PM

## Type of change

- [ ] New service widget
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] If adding a service widget or a change that requires it, I have added corresponding documentation changes.
- [x] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).
- [x] I have checked that all code style checks pass using pre-commit hooks and linting checks with `pnpm lint` (see development guidelines).
- [x] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #2498, opened 2023-12-19
Added Option for Grafana Alertmanager

myztillx:
## Proposed change

Added alertsmanager to grafana widget since the api used by `alerts` is deprecated (https://grafana.com/docs/grafana/latest/developers/http_api/alerting/). This change would put the number of blocks over 4, but only `alerts` or `alertmanager` will work, not both.

This is also not a feature request that had more than 5 upvotes, but does address a broken integration if using the new grafana alerts manager.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/grafana/component.jsx
+++ b/src/widgets/grafana/component.jsx
@@ -37,10 +37,7 @@ export default function Component({ service }) {
         label="grafana.alertstriggered"
         value={t("common.number", { value: alertsData.filter((a) => a.state === "alerting").length })}
       />
-      <Block
-        label="grafana.alertmanager"
-        value={t("common.number", { value: alertmanagerData.length })}
-      />
+      <Block label="grafana.alertmanager" value={t("common.number", { value: alertmanagerData.length })} />
     </Container>
   );
 }
--- a/src/widgets/grafana/component.jsx
+++ b/src/widgets/grafana/component.jsx
@@ -10,18 +10,20 @@ export default function Component({ service }) {
   const { widget } = service;
   const { data: statsData, error: statsError } = useWidgetAPI(widget, "stats");
   const { data: alertsData, error: alertsError } = useWidgetAPI(widget, "alerts");
+  const { data: alertmanagerData, error: alertmanagerError } = useWidgetAPI(widget, "alertmanager");
 
-  if (statsError || alertsError) {
+  if (statsError || alertsError || alertmanagerError) {
     return <Container service={service} error={statsError ?? alertsError} />;
   }
 
-  if (!statsData || !alertsData) {
+  if (!statsData || !alertsData || !alertmanagerData) {
     return (
       <Container service={service}>
         <Block label="grafana.dashboards" />
         <Block label="grafana.datasources" />
         <Block label="grafana.totalalerts" />
         <Block label="grafana.alertstriggered" />
+        <Block label="grafana.alertmanager" />
       </Container>
     );
   }
@@ -34,6 +36,10 @@ export default function Component({ service }) {
       <Block
         label="grafana.alertstriggered"
         value={t("common.number", { value: alertsData.filter((a) => a.state === "alerting").length })}
+      />
+      <Block
+        label="grafana.alertmanager"
+        value={t("common.number", { value: alertmanagerData.length })}
       />
     </Container>
   );
--- a/src/widgets/grafana/component.jsx
+++ b/src/widgets/grafana/component.jsx
@@ -14,7 +14,7 @@ export default function Component({ service }) {
 
   let alertsInt = 0;
 
-  if ((alertsError && alertsError.message === "HTTP Error") || !alertsData || alertsData.length === 0) {
+  if (alertsError || !alertsData || alertsData.length === 0) {
     if (alertmanagerData) {
       alertsInt = alertmanagerData.length;
     }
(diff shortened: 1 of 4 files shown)


myztillx:
My understanding is that while the api will return a 200 response, it will just return an empty array. At least that is what happens whenever I have tried to get the `alerts` to work. So, it is correct to say that both will "work" in that regard, but useful data will not be returned. 

It also appears that starting with grafana 9, alerting is handed via the alertmanager. The use of the alerts api must be opted into. (https://grafana.com/docs/grafana/latest/alerting/set-up/migrating-alerts/). I believe a user who has gone through all these steps will know which one they are using (alerts or alertmanager).

Removal will occur in grafana 11 ([source](https://grafana.com/docs/grafana/latest/developers/http_api/alerting/))

I am open to alternatives if there is a way to have an either/or option. I did not investigate the code a lot to see if this was an option.

shamoon:
Yea so what I want to make sure is if it’s returning status 200 and an array (even if it’s empty) that the user still sees 4 blocks, working as expected.

similarly for users on older versions the new api call will fail, won’t it? So that would appear as an error for them.

Basically want to avoid this being functionally a breaking change (I realize it is in grafana but ideally we handle here gracefully)

I have to see how annoying it is to spin up a grafana instance to test, I don’t use it any more

myztillx:
Understandable, I was going with the idea that the user would know which fields they were using and would chose those options. I was not thinking of when no fields have been chosen it would show all 5. 

I want to avoid making this a "breaking" change as you say. The best way I can think of would be to do both queries and if the `alerts` query returns an empty array or error (non 200 response or otherwise), then try the `alertmanager` query. This way, it is hidden from the user at this time and should not result in breaking anything. I will test this out shortly if this seems feasible to you?

shamoon:
Yes I think something like that would make sense. It’ll be slightly more complicated logic than i wish to account for all the cases, but I think worth it in terms of making things seamless for the homepage user. We’ve had similar breaking changes in an API like this in the past, for example https://github.com/gethomepage/homepage/pull/2284/files but in that case the API gave us version info to use, not sure if we have anything like that here or it’ll be sorta trial-and-error like you said.

LMK when you’re happy with it and I can take a look

myztillx:
So after version 11, we could use that. Version 11 could be checked for and to use the `alertmanager` in that case. But with the old api being opt-in for now, I do not think there is a way to detect which alerts the user is using (at least not based on version).

src/widgets/grafana/component.jsx
```diff
@@ -10,12 +10,23 @@ export default function Component({ service }) {
...
+  } else {
+    alertsInt = alertsData.filter((a) => a.state === "alerting").length;
+  }
+
+  if (statsError || (alertsError && alertmanagerError)) {
     return <Container service={service} error={statsError ?? alertsError} />;
   }
 
-  if (!statsData || !alertsData) {
+  if (!statsData || (!alertsData && !alertmanagerData)) {
```
myztillx, line 29: Need to check these as an and condition to prevent breaking when api is finally deprecated.

src/widgets/grafana/component.jsx
```diff
@@ -10,12 +10,23 @@ export default function Component({ service }) {
   const { widget } = service;
   const { data: statsData, error: statsError } = useWidgetAPI(widget, "stats");
   const { data: alertsData, error: alertsError } = useWidgetAPI(widget, "alerts");
+  const { data: alertmanagerData, error: alertmanagerError } = useWidgetAPI(widget, "alertmanager");
 
-  if (statsError || alertsError) {
+  let alertsInt = 0;
+
+  if ((alertsError && alertsError.message === "HTTP Error") || !alertsData || alertsData.length === 0) {
```
myztillx, line 17: How I see the flow working is checking for alertsError and the message "HTTP Error". This should occur when the api is finally removed in grafana 11. The second checks to see if any data was returned. And finally check if the returned data has length.

shamoon: Hmm, I would probably use the status code not a message, though in either case Im not certain what they will be, maybe we can figure that out.

But in general I think this would be OK. I guess the alternative is https://grafana.com/docs/grafana/latest/developers/http_api/other/#returns-health-information-about-grafana but that wont save us from the extra API call (in fact would add one, of course).

Have you had a chance to test this?

myztillx:
I was trying to, but it is fairly nested in the error message and I wasn't
sure how reliable it was to always be returned in that format.

I have tried this patch on my machine and it works for me, but I have the
newer alert manager setup and not the old alerts. Was hoping someone with
alerts setup would be able to test as well.

src/widgets/grafana/component.jsx
shamoon, line 17:
```suggestion
  if (alertsError || !alertsData || alertsData.length === 0) {
```

It should just be enough to check for the error. If the entire Grafana instance is down then everything should error out, so this wont be a problem and the later conditional will catch it

shamoon (changes requested):
Screenshot 2023-12-20 at 9 24 02 PM
Screenshot 2023-12-20 at 9 11 04 PM

Tested this with v8 and v9 and seems to be OK. See the one suggestion, LMK thoughts of course

shamoon (approved):

PR #2580, opened 2024-01-05
Enhancement: support multiple checks for healthchecks widget

strboul:
## Summary

I use Healthchecks in my homelab and manage multiple checks. While I am grateful for the initial development work outlined [here](https://github.com/gethomepage/homepage/pull/1023), I believe the initial approach could not have been the most effective. The reason being that having a widget for a single check often falls too short. I agree with the sentiments expressed in other [comments](https://github.com/gethomepage/homepage/pull/1023#issuecomment-1586332373) and the associated [feature request](https://github.com/gethomepage/homepage/discussions/2411). The widget would be vastly improved if it could display the total number of "ups" and "downs."

Other changes:

- I also got [confused](https://github.com/gethomepage/homepage/discussions/2566) about the terms "Online"/"Offline" that don't exist in the original application. I have opted to rename them as "Up"/"Down," aligning with the original terminology.

- I replaced the API endpoint from `v2` to `v3` as it's the most recent, see [here](https://healthchecks.io/docs/api/).

image

## Type of change

- [ ] New service widget
- [x] New feature (**breaking** change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/healthchecks.md
+++ b/docs/widgets/services/healthchecks.md
@@ -3,16 +3,21 @@ title: Health checks
 description: Health checks Widget Configuration
 ---
 
+Specify a single check by including the `uuid` field or show the total 'up' and 'down' for all
+checks by leaving off the `uuid` field.
+
 To use the Health Checks widget, you first need to generate an API key.
 
 1. In your project, go to project Settings on the navigation bar.
 2. Click on API key (read-only) and then click _Create_.
 3. Copy the API key that is generated for you.
 
+Allowed fields: `["status", "last_ping"]` for single checks, `["up", "down"]` for total stats.
+
 ```yaml
 widget:
   type: healthchecks
   url: http://healthchecks.host.or.ip:port
   key: <YOUR_API_KEY>
-  uuid: <YOUR_CHECK_UUID> # optional, if not present group statistics is shown
+  uuid: <CHECK_UUID> # optional, if not included total statistics for all checks is shown
 ```
--- a/docs/widgets/services/healthchecks.md
+++ b/docs/widgets/services/healthchecks.md
@@ -3,18 +3,16 @@ title: Health checks
 description: Health checks Widget Configuration
 ---
 
-To use the Health Checks widget, you first need to generate an API key. To do this, follow these steps:
+To use the Health Checks widget, you first need to generate an API key.
 
-1. Go to Settings in your check dashboard.
+1. In your project, go to project Settings on the navigation bar.
 2. Click on API key (read-only) and then click _Create_.
 3. Copy the API key that is generated for you.
-
-Allowed fields: `["status", "last_ping"]`.
 
 ```yaml
 widget:
   type: healthchecks
   url: http://healthchecks.host.or.ip:port
   key: <YOUR_API_KEY>
-  uuid: <YOUR_CHECK_UUID>
+  uuid: <YOUR_CHECK_UUID> # optional, if not present group statistics is shown
 ```
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -491,9 +491,9 @@
     },
     "healthchecks": {
         "new": "New",
-        "up": "Online",
+        "up": "Up",
         "grace": "In Grace Period",
-        "down": "Offline",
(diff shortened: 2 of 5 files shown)


shamoon:
I dont actually use healthchecks.io but I assume some of our users do and Im not sure how they'll feel about this change sorta breaking their widget. In general, they tend to complain when that happens, but maybe that would be outweighed by the "pros" of this. I have no idea.

Did you think about making this change opt-in? Or even opt-out?

ituri:
Would love to see this merged. This is exactly what’s missing in the existing widget.

Jayknightfr:
That's how the widget should have looked like from the start. The goal of the widget is to monitor our checks and get a visual cue of which are down. This PR makes the widget usefull 👍 
Also the "online/offline" stuff is confusing in the original widget.

strboul:
Hi @shamoon :wave:  I think this is the right implementation of the service (also happy to see the crowd resonates with this sentiment :green_heart: ). I initially thought of keeping the current version as optional but I realized that it's a bad idea to carry something wrong forward so I suggest introducing this breaking change.

Jayknightfr:
>  I initially thought of keeping the current version as optional but I realized that it's a bad idea to carry something wrong forward so I suggest introducing this breaking change.

@strboul It is indeed a breaking change that may be too much. As a general rule keeping backward compatibility should be kept, and as much as i find the current version useless, and yours spot on, it may be better to make your version as an option. This way updating the widget won't break for current users.

Another way of doing it would be to leave the current display, and add your two blocks "ups" and "downs", with the option to hide "offline" and "last ping". Though on a side note, i'd keep the last ping and have it display the time of the latest ping.

This way you're not breaking the widget for current users, but enhancing it 👍

shamoon (changes requested):
Comments from two individuals who already commented on the FR is not exactly convincing “sentiment” from the community. Presumably people who don’t want this to change were not subscribed to that thread and aren’t aware of this. And since we’re talking about tiny numbers there’s already one other voice so 50% of the total “supporters” :)

But really, the more I thought about it the more I think this just doesn’t need to be breaking. Using the presence (or not) of uuid I think you should be able to make this work fine in both instances.

If you’re not able to do that, please let us know.

strboul:
Fair enough! I have implemented it in that way. Not sure about the best implementation though, but it's non breaking now. Please review it

shamoon:
Now we’re talking 😎

Thanks for making the changes, I’ll take a look as soon as I can

shamoon (approved):
Excellent, progressive-enhancement FTW

Thanks

Screenshot 2024-01-07 at 8 49 16 AM
Screenshot 2024-01-07 at 8 49 33 AM

Churator:
I'm getting this error when I remove the UUID from the widget:
API Error: Invalid data
URL: https://healthchecks.io/api/v2/checks/

Config:
    - HealthChecks:
        href: https://healthchecks.io
        icon: healthchecks
        widget:
          type: healthchecks
          url: https://healthchecks.io
          key: <ommited>

Fully response:
{"checks": [{"name": "LanSVC", "slug": "lansvc", "tags": "", "desc": "", "grace": 60, "n_pings": 27636, "status": "up", "started": false, "last_ping": "2024-01-21T13:56:01+00:00", "next_ping": "2024-01-21T13:58:00+00:00", "manual_resume": false, "methods": "", "subject": "", "subject_fail": "", "start_kw": "", "success_kw": "", "failure_kw": "", "filter_subject": false, "filter_body": false, "unique_key": "<Removed>", "schedule": "*/2  * * * *", "tz": "Asia/Jerusalem"}, {"name": "Uptime Kuma", "slug": "uptime-kuma", "tags": "", "desc": "", "grace": 120, "n_pings": 492274, "status": "up", "started": false, "last_ping": "2024-01-21T13:56:04+00:00", "next_ping": "2024-01-21T13:58:04+00:00", "manual_resume": false, "methods": "", "subject": "", "subject_fail": "", "start_kw": "", "success_kw": "", "failure_kw": "", "filter_subject": false, "filter_body": false, "unique_key": "<Removed>", "timeout": 120}]}

PR #2668, opened 2024-01-17
Fix: Calendar timezone performance improvements

denispapec:
## Proposed change

This PR addresses slower performance when using timezone in calendar: 
- doubles the performance when timezone is specified
- slower timezone performance should have no impact when timezone is not specified. 

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/calendar.md
+++ b/docs/widgets/services/calendar.md
@@ -16,6 +16,7 @@ widget:
   view: monthly # optional - possible values monthly, agenda
   maxEvents: 10 # optional - defaults to 10
   showTime: true # optional - show time for event happening today - defaults to false
+  timezone: America/Los_Angeles # optional and only when timezone is not detected properly (slightly slower performance) - force timezone for ical events (if it's the same - no change, if missing or different in ical - will be converted to this timezone)
   integrations: # optional
     - type: sonarr # active widget type that is currently enabled on homepage - possible values: radarr, sonarr, lidarr, readarr, ical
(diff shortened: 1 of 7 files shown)


shamoon:
Thanks. What do you mean by "doubles the performance"? Basically this changes it so that if no timezone is specified dont run this "expensive" code, but specifying timezone is still very slow?

Do you have any iCal links with which to test?

denispapec:
> Thanks. What do you mean by "doubles the performance"? Basically this changes it so that if no timezone is specified dont run this "expensive" code, but specifying timezone is still very slow?
> 
> Do you have any iCal links with which to test?

Baseline speed is current behavior on `main`. It seems to update the calendar quite a bit faster now since I've moved logic where it changes timezone. Basically it's still slower than having no timezone definition, but calendar is now quicker and responsive enough in my case when timezone is defined. Sorry, don't have measurements to show precisely.

Found this example with that can replicate issue: https://ics.fixtur.es/v2/league/premier-league.ics

shamoon (approved):
Thanks, yea seems like most of the slowdown is https://github.com/denispapec/homepage/blob/2bd60080db9416b768abda8814a97f848b14e368/src/widgets/calendar/integrations/ical.jsx#L47 but we're so reliant on luxon it seems impractical to rip it out now.

Thanks for the link, works great in testing and yea, no penalty without specifying timezone

PR #2675, opened 2024-01-18
Fix: remove date time load delay

LukeHagar:
Howdy!

This small change should allow the date time to pop up on the initial page load and not after a one-second delay.

I'm just now getting into homepage in my homelab, and I am very interested in using and improving it any way I can.

### DIFF
--- a/src/components/widgets/datetime/datetime.jsx
+++ b/src/components/widgets/datetime/datetime.jsx
@@ -23,6 +23,7 @@ export default function DateTime({ options }) {
 
   useEffect(() => {
     const dateFormat = new Intl.DateTimeFormat(dateLocale, { ...format });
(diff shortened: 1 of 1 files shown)


shamoon:
Honestly, I dont know why but I have to assume there was a reason for this, I assume its a runtime thing. @benphelps do you remember why?

benphelps:
> Honestly, I dont know why but I have to assume there was a reason for this, I assume its a runtime thing. @benphelps do you remember why?

I don't remember exactly, but it's probably due to react hydration, or something along those lines with the compiled static output.

If it gives no compile or console errors and actually functions as expected, then I don't see why not.

shamoon (approved):
Great, thanks

PR #2775, opened 2024-01-27
Feature: search suggestions for search and quick launch

Flo2410:
## Proposed change

This PR adds search suggestions to the quick launch menu. 
Currently, only the top four suggestions are shown, but it could be made adjustable in the config. I just wasn't sure where to put the setting. To the quick launch setting or the search widget?

### Screenshot
image

### Config
I added an option to the `quicklaunch` setting called `hideSearchSuggestions` which hides the suggestions. 
Furthermore, I added the option `suggestionUrl` to the custom search widget. 
This allows the user to set a custom URL for suggestions, like this:

```yaml
- search:
    provider: custom
    url: https://www.ecosia.org/search?q=
    suggestionUrl: https://ac.ecosia.org/autocomplete?type=list&q=
    target: _blank
```

The body of the response from the API configured with  `suggestionUrl` should have the following format:

```json
[
  "home",
  [
    "home depot",
    "home depot near me",
    "home equity loan",
    "homeworkify",
    "homedepot.com",
    "homebase login",
    "home depot credit card",
    "home goods"
  ]
]
```

> The first entry of the array contains the search query, the second one is an array of the suggestions.
> In the example above, the search query was **home**.

The above example is also included in the change to the docs.

I found suggestion URLs for all implemented search providers and *Ecosia* which I used as an example for the custom provider.

Relates to #1225 and #1517

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:


shamoon (changes requested):
Thanks. I can take a look at the PR later but this should be opt-in not opt-out, otherwise we’re going to get a lot of complaints about this showing up out of nowhere.

in that case presumably the variable should be renamed to `show…`

Also FYI our guidelines now require 10 up-votes for new features but I realize the docs don’t show that for the latest version yet

shamoon:
Sorry I just noticed that the feature guidelines were in fact published before the PR was opened. In the interest of applying the rules equally (we get complaints otherwise) I’m going to close this but I’ll be more than happy to re-address when the requirement is met. I hope your understand, thank you for the PR

shamoon:
Sorry for the craziness, this closes two FRs so let’s call it > 10 :)

src/components/quicklaunch.jsx
```diff
@@ -109,9 +113,46 @@ export default function QuickLaunch({
...
+                // If there is an error, just ignore it. There just will be no search suggestions.
+              });
+          }
+
+          // Show search suggestions from their state. This will show the "old" suggestions until they are updated.
+          if (searchSuggestions[1]) {
+            newResults = newResults.concat(
+              searchSuggestions[1].map((suggestion) => ({
+                href: searchProvider.url + encodeURIComponent(suggestion),
+                name: suggestion,
```
shamoon, line 150:
```suggestion
                name: `${searchProvider.name ?? t("quicklaunch.custom")} ${t("quicklaunch.search")}: ${suggestion}`,
```

Im curious what you think about this one. I think it might be better to have the name of the search in the line to make it more explicit, it's obviously visually a bit more cluttered but I think it helps when e.g. theres also a service in the list.

I think actually even better would be to add a new quick launch "type", something like "searchSuggestion" and then that could be slightly visually differentiated like smaller or gray text or something like that

Again, curious your thoughts

Screenshot 2024-01-27 at 5 02 04 PM

src/components/widgets/search/search.jsx
```diff
@@ -12,31 +12,37 @@ export const searchProviders = {
...
   brave: {
     name: "Brave",
     url: "https://search.brave.com/search?q=",
+    suggestionUrl: "https://search.brave.com/api/suggest?&rich=false&q=",
     icon: SiBrave,
   },
   custom: {
     name: "Custom",
     url: false,
+    suggestionUrl: false,
```
shamoon, line 45:
```suggestion
```

src/pages/index.jsx
```diff
@@ -212,7 +212,11 @@ function Home({ initialSettings }) {
       searchProvider = getStoredProvider() ?? searchProviders[searchWidget.options.provider[0]];
     } else if (searchWidget.options?.provider === "custom") {
       searchProvider = {
-        url: searchWidget.options.url,
+        ...searchProviders.custom,
+        ...{
+          url: searchWidget.options.url,
+          suggestionUrl: searchWidget.options.suggestionUrl,
+        },
```
shamoon, line 219:
```suggestion
        name: "Custom",
        url: searchWidget.options.url,
        suggestionUrl: searchWidget.options.suggestionUrl,
```

shamoon (changes requested):
It's also a bit odd that this works only for quick launch but comes from search. I think it should be implemented for both so they mirror each other. We'll need to make sure the UI for the search one is good.

src/components/quicklaunch.jsx
Flo2410: That's a good question and probably depends on personal preferences. 

I don't like how it looks with the `Google Search:` in front of the suggestions. I find it difficult to quickly see the suggestion, always having to scan the line for the colon/start of the suggestion.

I have two ideas myself down below.

image

## Idea 1

Showing the icon of the search provider would be an easy way of indicating that it is a search and which provider is used. Ideally, the icon would be the same as the one used in the search bar. For that to work, the `ResolvedIcon` component would need to allow `react-icons` icons to be passed in as an icon. Alternatively, the search bar could use the current version of `ResolvedIcon` and the `searchProviders` object would store the SI icon string instead of the component from `react-icons`. I did not test these yet, for the images below I just hard-coded the string for the `icon` property. 

The result would look like this:

image

## Idea 2

My second idea would include your suggestion of creating a new "type" for the search suggestions. 
With that, we could show the suggestion part of the search query with a lower opacity or a darker color (for the example below, I used `opacity-50`).

I must say, I really like how it looks. It's super easy to see the suggestions, and it's clear which provider they use. 

image

src/pages/index.jsx
Flo2410: I'm curious about that one. Why wouldn't you copy the custom search provider from the `searchProviders` object when it is done for all others? 
Furthermore, shouldn't the `name` property be `undefined` for the custom provider?
The `custom` name is resolved here (as I understand it): https://github.com/gethomepage/homepage/blob/d20cdbb9ab1432e69ce68c9b4255aef9eac52ff4/src/components/quicklaunch.jsx#L112

Moreover, if we are going to show the icon, the icon for the custom provider would need to be set here as well. Or we get it from the `searchProviders`

So if I didn't overthink something, it should look like this:
```js
searchProvider = {
  url: searchWidget.options.url,
  suggestionUrl: searchWidget.options.suggestionUrl,
  icon: searchProviders.custom.icon // only if we show the icon of the prvider as suggested above
};
```

Flo2410:
Regrading the search in the search bar: 

I made a quick implementation using the `Combobox` from `headlessui`. 
It obviously still needs some work, but I would like some guidance regarding the design of the dropdown. 

image

Furthermore, I would suggest moving the settings for the search provider, including the `showSearchSuggestions` setting, into a [provider](https://gethomepage.dev/main/configs/settings/#providers) in the future. This probably would also enable the search function in the quick launch to work without having the search widget enabled. But this most likely falls out of the scope of this PR.

src/pages/index.jsx
shamoon: Yea all I was trying to say here was that `{ ...{ url: ... } }` is just a weird way of writing `{ url: ... }`, I didnt really double-check the specific properties are what we want

src/components/quicklaunch.jsx
shamoon: Yea, idea 2 is exactly the kind of thing I had in mind, agree it looks great.

I'm not sure about the icon but Im also sorta tempted to leave it for another PR, keep this one focused?

shamoon (commented):
Thanks for changes.

Yea the search widget auto-complete is a good start. As far as the design, it'd be easier if you push those changes and I can play with it, if youre OK with that.

And yea, maybe it seems overkill but users generally want more control especially as this is sending data (and getting tracked, of course) so I would include a setting for the widget (showSuggestions) in addition to the quick launch one. And I think something even better would be what we've done similarly elsewhere which is to let the search widget one act as the 'default' for quick launch (in other words if you set search widget `showSuggestions` to true, the quick launch will do the same unless you explicitly set it to false). If you only want them in quick launch set it just in quick launch. That feels 'clean' and means a lot of users will only need to set it in one place (if you want suggestions in search widget theres a good chance you do in QL).

docs/widgets/info/search.md
```diff
@@ -28,4 +29,25 @@ multiple providers is also supported via a dropdown (excluding custom):
     provider: [brave, google, duckduckgo]
 ```
 
+The response body for the URL provided with the `suggestionUrl` option should looke like this:
```
shamoon, line 32:
```suggestion
The response body for the URL provided with the `suggestionUrl` option should look like this:
```

Flo2410:
I pushed the draft for the suggestions in the search bar.
You can activate them by adding `showSearchSuggestions: true` to the search widget config.

src/pages/index.jsx
Flo2410: I've fixed it in 5f0924d9328cda79d118a630ff8916e372c62964

src/components/quicklaunch.jsx
Flo2410: I added it without the image in ed827837aa86a892fd7603e6d0dc8187138f58c2

It looks like this:
image

shamoon:
Cool thanks! I think something happened with the custom provider for search widget, I'll take a look at all of it as soon as I can find a minute

src/components/quicklaunch.jsx
```diff
@@ -17,18 +16,11 @@ export default function QuickLaunch({
...
-    ? settings.quicklaunch
-    : {
-        searchDescriptions: false,
-        hideVisitURL: false,
-      };
-  const showSearchSuggestions =
-    settings?.quicklaunch?.showSearchSuggestions ?? searchWidget.options.showSearchSuggestions ?? false;
+  const { searchDescriptions = false, hideVisitURL = false } = settings?.quicklaunch ?? {};
+  const showSearchSuggestions = !!(
+    settings?.quicklaunch?.showSearchSuggestions ?? searchProvider.showSearchSuggestions
```
Flo2410, line 22: This does not work as expected, as `searchProvider.showSearchSuggestions` is only set for the `custom` search provider. This is because for the custom provider the `searchProvider` gets set to the options, but the predefined ones get set from the `searchProviders` object in `search.jsx`.

src/pages/index.jsx
```diff
@@ -211,10 +211,7 @@ function Home({ initialSettings }) {
       // if search provider is a list, try to retrieve from localstorage, fall back to the first
       searchProvider = getStoredProvider() ?? searchProviders[searchWidget.options.provider[0]];
     } else if (searchWidget.options?.provider === "custom") {
-      searchProvider = {
-        url: searchWidget.options.url,
-        suggestionUrl: searchWidget.options.suggestionUrl,
-      };
+      searchProvider = searchWidget.options;
```
Flo2410, line 214: I am not sure what to do here exactly, but currently the `searchProvider` object looks different for the `custom' and predefined providers.

As the problem is above, the predefined one is missing the `showSearchSuggestions` option and the custom one has no `name` (the quick launch is relying on that for showing the `t("quicklaunch.custom")`) https://github.com/gethomepage/homepage/blob/ef257565344f6c1db818609b94e01eec28993b31/src/components/quicklaunch.jsx#L118

custom provider:
```json
"index": 1,
"provider": "custom",
"showSearchSuggestions": true,
"suggestionUrl": "https://ac.ecosia.org/autocomplete?type=list&q=",
"target":  "_blank",
"url":  "https://www.ecosia.org/search?q="
```

predefined provider:
```json
"icon": SiGoogle,
"name": "Google",
"suggestionUrl": "https://www.google.com/complete/search?client=chrome&q=",
"url": "https://www.google.com/search?q="
```

Flo2410 (commented):
The search provider setting is still broken... 
I'm not sure what a good solution would be.

Flo2410:
I also found a bug with the greying out of the suggestions.
I will fix this and would also implement it for the search bar suggestions.

image

src/components/quicklaunch.jsx
shamoon: Oh good point, I'll take a look

shamoon:
Great catch about the non-custom providers, I think should be sorted but lmk

src/components/quicklaunch.jsx
```diff
@@ -219,7 +273,15 @@ export default function QuickLaunch({
                           </div>
                         )}
                         <div className="flex flex-col md:flex-row text-left items-baseline mr-4 pointer-events-none">
-                          <span className="mr-4">{r.name}</span>
+                          {r.type !== "searchSuggestion" && <span className="mr-4">{r.name}</span>}
+                          {r.type === "searchSuggestion" && (
+                            <>
+                              <span className="whitespace-pre">{searchString}</span>
+                              <span className="mr-4 whitespace-pre opacity-50">
+                                {r.name.substring(searchString.length)}
```
shamoon, line 281:
```suggestion
                              <span className="whitespace-pre">{r.name.indexOf(searchString) === 0 ? searchString : ''}</span>
                              <span className="mr-4 whitespace-pre opacity-50">
                                {r.name.indexOf(searchString) === 0 ? r.name.substring(searchString.length) : r.name}
```

I think something simple like this would do it, but realize you may be working on it

Screenshot 2024-01-30 at 11 18 09 AM
Screenshot 2024-01-30 at 11 18 19 AM

shamoon: Just playing with them a bit aren't the suggestions always going to be at the beginning of the string if its in the suggestion at all?

Flo2410: No, unfortunately not.

 See the `http://gmx.at`

Flo2410: Okay, I gave up on trying my idea...

I would probably implement your first idea because with your second one, if you spell the search wrong, everything full opacity. I think it looks better all with reduced opacity in that case, because nothing is actually matching the input.

image

image

shamoon: Ok yea Im good with that.

Flo2410:
I think we are good here.
The only thing that bothers me is that the Google API doesn't return UTF-8 strings.

Flo2410:
I couldn't live with it. Looks like Google only returns UTF-8 if the User Agent includes `Mozilla/5.0`.

src/utils/proxy/cached-fetch.js
shamoon, line 1: Oh. What does this actually fix? I'd prefer we not touch this, just in the tiny possibility of unintended consequences and all.

Flo2410: This sets the `User-Agent`  to a value recognized by the Google search suggestion API so that it returns the response UTF-8 encoded. The default user agent for node.js is `undici` or as of some 21.x.x version `node`.
I suspect that the change would only improve compatibility, but I can understand your perspective. So, maybe it could be investigated and changed in a future PR.

With the `User-Agent` header:
image

Without the header:
image

Also, the API gets spammed with this because *�* doesn't match *ü*

Flo2410:
While writing the response above, I found some more bugs:

The first one is the Z-Index of the quick launch not being high enough:

image

The second one are unexpected line breaks because of the `span` elements for the suggestions.

image

The window width was 620px for the second screenshot.

shamoon:
Thanks great catches! Should be fixed and I reverted the UA thing. Agree we can address in the future if needed, you ok with that?

Flo2410:
I found another bug, I didn't URI encode the query again in the API route. This made it so that special characters were not correctly sent to the search suggestions API. 

But apart from that, I think we are fine.

The thing with the Umlaut for Google doesn't bother me too much (I use DuckDuckGo anyway ;) ) 

While testing the Umlaut thing, I found another bug, but I will open an issue for that one.

shamoon:
> How do I report a Bug?

You can open a discussion about it.

shamoon (approved):
Great work

PR #2867, opened 2024-02-08
Fix: search opens when losing focus, prevent unnecessary search API calls

Flo2410:
## Proposed change

This PR fixes two bugs with the search suggestion.

- The search box opened the search URL when loosing focus.
- The suggestion API kept getting called without the search changing.

Closes #2866 

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/widgets/search/search.jsx
+++ b/src/components/widgets/search/search.jsx
@@ -167,6 +167,7 @@ export default function Search({ options }) {
               autoComplete="off"
               // eslint-disable-next-line jsx-a11y/no-autofocus
               autoFocus={options.focus}
+              onBlur={(e) => e.preventDefault()}
             />
             <Listbox
               as="div"
--- a/src/components/widgets/search/search.jsx
+++ b/src/components/widgets/search/search.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, Fragment } from "react";
+import { useState, useEffect, useCallback, Fragment, useRef } from "react";
 import { useTranslation } from "next-i18next";
 import { FiSearch } from "react-icons/fi";
 import { SiDuckduckgo, SiMicrosoftbing, SiGoogle, SiBaidu, SiBrave } from "react-icons/si";
@@ -71,6 +71,9 @@ export function getStoredProvider() {
 export default function Search({ options }) {
   const { t } = useTranslation();
 
+  const searchProviderButton = useRef();
+  const comboboxOptions = useRef();
+
   const availableProviderIds = getAvailableProviderIds(options);
 
   const [query, setQuery] = useState("");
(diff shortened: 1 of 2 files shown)


shamoon (changes requested):
Thanks for finding / working on these.

-  I still see the focus bug if you tab out of the search widget
- I cant recreate the other one, not sure exactly what conditions are needed. Anyway, that change seems fine.

shamoon:
Oh, perhaps the other PR will fix the tab thing.

shamoon:
Gotcha. Please put the tab fix here, the other one is a feature and to be fair I think it should meet the usual requirements. Sorry for the confusion.

Flo2410:
I've added the fix for the Tab issue.

src/components/widgets/search/search.jsx
```diff
@@ -119,6 +119,12 @@ export default function Search({ options }) {
     };
   }, [selectedProvider, options, query, searchSuggestions]);
 
+  const handleSearchKeyDown = (event) => {
+    if (event.key === "Tab") {
+      event.preventDefault();
```
shamoon, line 124: Im on my phone (so cant test right this second) but wont this also prevent actually tabbing out of the field?

shamoon (changes requested):

src/components/widgets/search/search.jsx
Flo2410: Yes, it does.  I'll try to fix that.

FYI: The reason the *tab* key submits the search is the [headlessui combobox](https://headlessui.com/react/combobox#keyboard-interaction)

shamoon: Yea I figured it came from that, might be annoying to figure out a workaround if they dont provide an easy way

Flo2410: Fixed it. Now the *Tab* is on disabled if the suggestion dropdown is open (it can be closed with *escape*).

shamoon: Hmm, but this has the same issue, no? If the user wants to tab out of the field i.e. they have multiple search engines and they want to tab over to that dropdown, they'd have to hit esc first?

Maybe its the best we can do, I find the opening on Tab a weird thing but its listed under their accessibility notes, I'm not aware of that interaction...

Flo2410: Maybe there should be a custom implementation of the combobox for the search widget. 

But for now, the search provider button gets manually focused if the dropdown is open.

Flo2410: Also, just remembered that the `useRef` hook exists. Therefore, I changed the `document.getElementById` calls to a ref with the `useRef` hook.

shamoon: This feels like the wrong direction, we're just creating more and more specific cases. For example you cant tab backwards (shift tab) out of the field.

Perhaps just not using onChange at all and using our own handlers? We only want the search to open on click or enter key, after all...

I'll find some time to play with it and see, indeed maybe we cant do any better.

shamoon: It's a little odd, but take a look at the last commit and let me know what you think. Still open to another solution, of course

Flo2410:
Nice, that's much better. Thx.

shamoon (approved):
Great, thanks again for hunting these down

PR #2934, opened 2024-02-18
Enhancement: allow api auth with moonraker service

stevendejongnl:
## Proposed change

Changed the genericProxyHandler to use the credentialedProxyHandler for moonraker instances that uses authorization by API keys.
With the genericProxyHandler you get a 401 unauthorized error on those instances.

## Type of change

### DIFF
--- a/docs/widgets/services/moonraker.md
+++ b/docs/widgets/services/moonraker.md
@@ -12,3 +12,13 @@ widget:
   type: moonraker
   url: http://moonraker.host.or.ip:port
 ```
+
+
+If your moonraker instance has an active authorization and your homepage ip isn't whitelisted you need to add your api key ([Authorization Documentation](https://moonraker.readthedocs.io/en/latest/web_api/#authorization)).
+
+```yaml
+widget:
+  type: moonraker
+  url: http://moonraker.host.or.ip:port
+  key: api_keymoonraker
+```
--- a/docs/widgets/services/moonraker.md
+++ b/docs/widgets/services/moonraker.md
@@ -13,7 +13,6 @@ widget:
   url: http://moonraker.host.or.ip:port
 ```
 
-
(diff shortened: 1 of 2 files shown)


shamoon:
Please provide an explanation and or / link to a discussion, even if the change seems obvious. It doesn’t seem any other moonraker users have asked about auth

stevendejongnl:
> Please provide an explanation and or / link to a discussion, even if the change seems obvious. It doesn’t seem any other moonraker users have asked about auth

Sorry, I submitted to quickly.
Updated the description

shamoon:
Thanks. So moonraker api uses 

https://github.com/gethomepage/homepage/blob/36585a2c44acda10bea1c7a1efe0ef78956b213d/src/utils/proxy/handlers/credentialed.js#L69

?

stevendejongnl:
> Thanks. So moonraker api uses
> 
> https://github.com/gethomepage/homepage/blob/36585a2c44acda10bea1c7a1efe0ef78956b213d/src/utils/proxy/handlers/credentialed.js#L69
> 
> ?

Thats correct:
https://moonraker.readthedocs.io/en/latest/web_api/#authorization

stevendejongnl:
> Also, thing to check here is that it won’t break non-auth instances…
> 
> (also assuming it never allows basic auth, which this would also break)



> Also, thing to check here is that it won’t break non-auth instances…
> 
> (also assuming it never allows basic auth, which this would also break)

Good point, I will spin up another instance without authentication to check this.

shamoon:
Great thanks. Yea as you can guess I don’t have an instance to test.

Thanks for the docs link, yea in this case please update the homepage docs to note the optional `key` parameter

stevendejongnl:
> Great thanks. Yea as you can guess I don’t have an instance to test.
> 
> Thanks for the docs link, yea in this case please update the homepage docs to note the optional `key` parameter

I can confirm that an instance without authentication and no key set in the configuration does work:
image
image

shamoon (approved):
Thanks for confirming that.

I don't really think this is a bug but since it literally adds 0 code thats cool. Do try to mind the PR requirements (checkboxes you removed) in the future (here and in general).

PR #2964, opened 2024-02-21
BUG - Disk sizes for the glances and resources widgets are reporting incorrect sizes

brikim:
The sizes are reported in bytes and need the binary: flag set to true so that the values are divided by 1024 and not 1000. The memory widgets were already doing this but the disk widgets missed this option.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/components/widgets/glances/glances.jsx
+++ b/src/components/widgets/glances/glances.jsx
@@ -132,9 +132,9 @@ export default function Widget({ options }) {
         <Resource
           key={`disk_${disk.mnt_point ?? disk.device_name}`}
           icon={FiHardDrive}
-          value={t("common.bytes", { value: disk.free })}
+          value={t("common.bytes", { value: disk.free, binary: true })}
           label={t("glances.free")}
-          expandedValue={t("common.bytes", { value: disk.size })}
+          expandedValue={t("common.bytes", { value: disk.size, binary: true })}
           expandedLabel={t("glances.total")}
           percentage={disk.percent}
           expanded={options.expanded}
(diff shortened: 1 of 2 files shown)


shamoon:
I do not think this is accurate?

My drive has 255217975296 bytes available. Homepage reports 255GB
https://duckduckgo.com/?q=255217975296+bytes+in+GB&t=h_&ia=web
With your PR it reports 238GiB, which I dont think we want for disk drives
https://duckduckgo.com/?t=h_&q=255217975296+bytes+in+GiB&ia=web

Maybe you can provide more context...

brikim:
The glances homescreen any my synology nas reports the values that I get when doing the 1024 division. The way the widgets currently work it just divides by 1000 which isn't the correct way to convert bytes to Gigabytes or Terabytes. View the glances FS service widget to see that it is using binary bytes bbytes instead of the bytes conversion.

shamoon:
1. People freak out when we change units for no reason.
2. This PR was for resources and glances, which are you saying is wrong? I only looked at resources but I can check glances. But we dont need to match what glances shows as long as what we show is correct. Thats what matters.

shamoon:
With respect to the image above: https://duckduckgo.com/?va=e&t=hj&q=890+GB+in+GiB&ia=web

So if this is all about changing what units are displayed in the info widgets, we will politely decline, at least not without evidence that others are unhappy with this, see point 1 above. Argument could certainly be made to bring the glances service widget in line with this. But the point is neither is actually wrong.

shamoon:
@brikim see #2966, I think thats the best plan here, let the user decide. Appreciate all your efforts.

brikim:
@shamoon Sounds good! Thanks for doing this.

PR #2986, opened 2024-02-23
Feature: ESPHome widget

RoboMagus:
## Proposed change

image

This adds a service widget for ESPHome, as the CustomAPI widget does not allow for API response data processing as required to get the sum of devices per state.

Closes #2070

## Type of change

- [x] New service widget
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If applicable, I have added corresponding documentation changes.
- [x] If applicable, I have reviewed the [feature](https://gethomepage.dev/latest/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).
- [x] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/latest/more/development/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/latest/more/development/#code-linting).
- [x] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #3106, opened 2024-03-11
Add refresh interval to EspHome

RoboMagus:
## Proposed change

This adds a `refreshInterval` to the EspHome widget (As the device states that are being monitored are expected to change from time to time).

This was something I missed in #2986.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/esphome.md
+++ b/docs/widgets/services/esphome.md
@@ -13,4 +13,5 @@ Allowed fields: `["total", "online", "offline", "unknown"]`.
 widget:
   type: esphome
   url: http://esphome.host.or.ip:port
+  refreshInterval: 30000 # optional - in milliseconds, defaults to 30s. Minimal allowed: 10000.
 ```
--- a/src/widgets/esphome/component.jsx
+++ b/src/widgets/esphome/component.jsx
@@ -4,11 +4,17 @@ import Block from "components/services/widget/block";
 import Container from "components/services/widget/container";
 import useWidgetAPI from "utils/proxy/use-widget-api";
 
+const defaultInterval = 30000;
+const minRefreshInterval = 10000;
+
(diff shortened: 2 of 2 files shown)


RoboMagus:
@shamoon, as stated;
> This was something I missed in https://github.com/gethomepage/homepage/pull/2986.

I was initially under the impression that all widgets would regularly refresh their data. Based on what I had seen after broader deployment is that even when ESPHome devices had been offline for a while the widgets on HomePage would not reflect their actual state.

Also, as this functionality for ESPhome provides essentially a health check, this widget would basically fall into the same category as e.g. `ping`, `site-monitor`, `uptime` etc. All of these widgets _do_ expose the refreshInterval setting. So I would ask you to reconsider as at this point lacking this feature really is missing functionality.

If needed I'd be happy to make changes such that by default the refresh is disabled, but at least allows users to configure their system to their liking.

shamoon:
Almost all widgets pull data that can change over time, in general we do not support adding the parameter because its a slippery slope, homepage already generates a lot of requests and we want to minimize that.

RoboMagus:
>homepage already generates a lot of requests and we want to minimize that.

Understandable. It's ok to keep the refreshing on interval basis disabled by default. But at least allow for people to configure this as they see fit.
And the data  this widget shows is not something that _could eventually change at some point in time_. Its something that if it changes generally has some impact, just like what's done with the `ping` integration. Something that if it changes you don't want to refresh the dashboard for in order to find out.

Also, if it helps. the data that is pulled over the API from ESPHome for this ping request is just about as minimal as it gets. No big data structures to be discarded, but a simple key value mapping of all nodes and their states. I.e. the impact on the homepage backend is minimal.

PR #2363, opened 2023-11-22
Enhancement: Add configurable refresh interval and points for glances service

ni0ki:
## Proposed change

Add support for configurable refresh interval and amounts of points for the Glances service widget.

``` services.yaml
-CPU:
  widget:
    type: glances
    url: http://url:port
    metric: cpu
    refreshInterval: 2500
    pointsLimit: 50
```

This opens 2 possibilities, mainly useful for people with a home that stays open for a long period:
- Reduce the number of API calls by increasing the interval.
- Have different monitored periods (see below)

image

**A small note about network / disk**
Information about IO/s for net (RX, TX) and disk (Read, Write) might seems strange, but is still correct. It will simply show the value for the immediate last refresh on glances' side, regardless of the configured interval

(I also alphabetically sorted the allowed keys to ease finding existing values)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] If adding a service widget or a change that requires it, I have added corresponding documentation changes.
- [ ] If adding a new widget I have reviewed the [guidelines](https://gethomepage.dev/latest/more/development/#service-widget-guidelines).
- [ ] I have checked that all code style checks pass using pre-commit hooks and linting checks with `pnpm lint` (see development guidelines).
- [ ] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #3164, opened 2024-03-22
Fix: Glances cpu & info widgets information

sunnycloudy1337:
## Proposed change

This fix is for Glances Widget Service where the `info` metric without chart would only display the CPU instead of the OS, but most importantly fix the `cpu` metric to show the CPU brand / model as well as the temperature, instead of the OS as before.

The `cpu` metric now shows the temperatures as well, current and warn, just like the informational widget, and you can change the units to fahrenheit with `tempUnits: imperial`

Also updated the documentation to reflect `refreshInterval` and `pointsLimit` introduced in #2363 

image

Closes #3160 

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/latest/more/development/#code-formatting-with-pre-commit-hooks) 
- [X] I have checked that all code style checks pass using [linting checks](https://gethomepage.dev/latest/more/development/#code-linting).

### DIFF
--- a/docs/widgets/services/glances.md
+++ b/docs/widgets/services/glances.md
@@ -19,7 +19,6 @@ widget:
   password: pass # optional if auth enabled in Glances
   metric: cpu
   diskUnits: bytes # optional, bytes (default) or bbytes. Only applies to disk
-  tempUnits: imperial # optional, metric (default) or imperial. Only applies to cpu
   refreshInterval: 5000 # optional - in milliseconds, defaults to 1000 or more, depending on the metric
   pointsLimit: 15 # optional, defaults to 15
 ```
--- a/docs/widgets/services/glances.md
+++ b/docs/widgets/services/glances.md
@@ -19,6 +19,9 @@ widget:
   password: pass # optional if auth enabled in Glances
   metric: cpu
   diskUnits: bytes # optional, bytes (default) or bbytes. Only applies to disk
+  tempUnits: imperial # optional, metric (default) or imperial. Only applies to cpu
+  refreshInterval: 5000 # optional - in milliseconds, defaults to 1000 or more, depending on the metric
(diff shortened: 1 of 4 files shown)


shamoon (changes requested):
1. I have no idea why you think taking an [adversarial tone](https://github.com/gethomepage/homepage/discussions/3160#discussioncomment-8876694) is a good idea. Please drop the attitude or feel free to contribute elsewhere, we have no need for it.
2. What you’ve done here is mixed a feature (temperature units business) with a fix, which ain’t cool and isn’t good open source contributing. Feature PRs require up-votes, fixes don’t. Please change this PR to only address the cpu/info labels issue discussed and we’ll be happy to merge it

shamoon (approved):

sunnycloudy1337:
So now the CPU Small is a bit barren.

Before the merge, there used to be the hostname in there, but because the sensors proposition wasn't approved and because we don't have the system data anymore, can we populate that space with info from the already loaded `cpu` api call? Like maybe a core count?

## After Merge
image

## Proposed Change
image

Would you accept this change? Or do you have another idea about what info should we display there?
We only have [cpu API](https://glances.readthedocs.io/en/latest/api.html#get-cpu) and [quicklook API](https://glances.readthedocs.io/en/latest/api.html#get-quicklook) loaded atm.

Thanks!

shamoon:
Using the already existing api calls is a good start, yea. I’d just open a FR for whatever you think is the best option, or the temp option.

sunnycloudy1337:
Was trying to avoid opening a discussion and just quick-fixing the barren space 😅

PR #3195, opened 2024-03-28
Documentation: Update authentik api key info

ThorTheStorm:
## Proposed change

Add more information about what permissions are needed for the widget to retreive the fields it is intended to retrieve from Authentik.

Background for this is i wasted quite a bit of time figuring this out, and don't want a service account that retrieves statistics having more permissions than necessary.

### DIFF
--- a/docs/widgets/services/authentik.md
+++ b/docs/widgets/services/authentik.md
@@ -16,6 +16,12 @@ You will need to generate an API token for an existing user. To do so follow the
 5. Click the Create button on the dialog
 6. Click the copy button on the far right of the newly created API Token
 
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks, would you mind updating the instructions above to be way more succinct? Our docs shouldn’t be a list of UI elements to click in another software, it should be maybe one or two steps.

If not, I can do that

ThorTheStorm:
A bit unsure what you mean actually.

Should i write like a small guide on how to do it step-by-step, or maybe refer to Authentik-documentation on it as a first step?
Or is it generally just the point-list that is sub-standard?

shamoon (approved):
Took care of it, thanks

PR #3345, opened 2024-04-22
Documentation: clarify uptime kuma slug

NuAlex:
Took me a while to figure this out, so here's a small clarification that can same a lot of time.

## Proposed change

Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/latest/more/development/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/latest/more/development/#code-linting).

### DIFF
--- a/docs/widgets/services/uptime-kuma.md
+++ b/docs/widgets/services/uptime-kuma.md
@@ -5,7 +5,7 @@ description: Uptime Kuma Widget Configuration
 
 Learn more about [Uptime Kuma](https://github.com/louislam/uptime-kuma).
 
-As Uptime Kuma does not yet have a full API the widget uses data from a single "status page". As such you will need a status page setup with a group of monitored sites, which is where you get the slug (without the `/status/` portion). E.g. if your status page is https://url-address/status/MyStatusPage, insert `slug: mystatuspage`, in lower-case as slug is case-sensitive.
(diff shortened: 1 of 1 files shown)


NuAlex:
By just assuming that _Uptime kuma calls this value a slug in multiple places_, you're clearly not being inclusive of newbies. I'm new to Uptime Kuma and Homepage. I never seen this word before and after creating the status page you won't see the word "slug" anywhere! I had created the status page a couple of weeks ago, so I didn't recall seeing any "slug". 

As you mention, Uptime Kuma converts the slug to lower-case after you hit save, so you're assuming everyone will notice that?
I used some capitals when creating the status page, which **are preserved** in the page title, so I used that in the widget, because I use that same string in multiple things, just to realize later that slug in case-sensitive... 

And if not for that, the documentation does not explain what is a "slug" so adding a concrete example of how it works does a much better job of explaining things and can save people some troubleshooting and googlinhg. I only found out by myself after debugging the API calls since I was really motivated to fix, otherwise I would just quit trying and remove the widget.

Plus, this is just adding one short and single sentence to the doc, so disregarding such a small volunteered and humble contribution feels quite arrogant. Unless maybe, you're trying to save some characters on the internet... it beats me!

shamoon (approved):
Maintaining short and accurate docs is important imho. I didnt like being called a name, internet stranger.

PR #3352, opened 2024-04-22
Enhancement: Change default k8s label selector to be from resource

graytonio:
## Proposed change

Enhance default k8s pod selector by using the api to automatically discover the service used by the ingress resource. Reduces need to manually set pod selectors and does not break existing functionality. I opened a discussion on this but I was unclear if I needed to upvotes for this since it's not really a new feature but please let me know if I do.

Closes #3348

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [x] Other (please explain)

Enhancement of current functionality to integrate more natively in kubernetes.

## Checklist:

- [ ] (Did not test multiple front end platforms due to it being a backend change)

### DIFF
--- a/src/utils/kubernetes/kubernetes-utils.js
+++ b/src/utils/kubernetes/kubernetes-utils.js
@@ -1,14 +1,35 @@
 import { CoreV1Api, NetworkingV1Api } from "@kubernetes/client-node";
 
-export async function parseIngressSelector(ingressName, namespace, kc) {
+export async function parseIngressSelector(ingressName, namespace, kc, logger) {
   const coreApi = kc.makeApiClient(CoreV1Api);
   const networkApi = kc.makeApiClient(NetworkingV1Api);
 
-  const ingress = await networkApi.readNamespacedIngress(ingressName, namespace);
-  const serviceName = ingress.body.spec.rules[0].http.paths[0].backend.service.name;
+  const ingress = await networkApi
+    .readNamespacedIngress(ingressName, namespace)
+    .then((response) => response.body)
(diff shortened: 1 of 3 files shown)


shamoon:
I dont use k8s so our k8s integration could probably use some love. My question about this is isnt it going to be a completely breaking change for existing users who aren't using an explicit podSelector?

graytonio:
I'm working on that part right now I don't want it to be a breaking change.

The way I'm thinking about it right now if you have manually configured the app and namespace parameters in your config file either:
A. An ingress exists with that exact name and very likely that is the application you want to target
B. An ingress does not exist with that name and it will default to the current behavior

graytonio:
I was able to validate that you can configure the services manually in the config file and it will still work as intended.

both working

src/utils/kubernetes/kubernetes-utils.js
shamoon, line 1: So does this mean existing users will get these errors?

shamoon:
> I opened a discussion on this but I was unclear if I needed to upvotes for this since it's not really a new feature but please let me know if I do.
> 
> Closes #3348

Oh yea I just noticed this, the answer is yes. Also in general I'm going to be hesitant to make this change without getting confirmation from real k8s users. This setup has been around for a while and went through a few iterations (you can look through the history) to what seems to be one that makes folks happy, Im disinclined to suddenly change it without clear support & rationale.

If that does happen of course we can re-address this PR, but we dont leave PRs open in the meantime.

graytonio:
Understandable if it gets the support we can reopen this.

PR #3500, opened 2024-05-20
Site monitor url defaults to services href

neBM:
The approach where we don't allow for a fallback `siteMonitor` URL increases config duplication as it's common for administrators to set `siteMonitor` to the same value as the `href`. In the scenario where an administrator has many services with different `href` values and they want the `siteMonitor` to match `href`, they would need to configure both  `siteMonitor` and `href` values on each service item.

## Proposed change
The proposed changes address this issue by adding a new condition to handle the case where `siteMonitor` is `true`. In this case, it checks if the `href` property exists. If the `href` property does exist, it sets `monitorURL` to `href`. Otherwise, it _fails with message_. 

This change allows the use of the `href` property as the monitor target when `siteMonitor` is set to `true`.

> ```yaml
>   - dummyService:
>       href: https://github.com/
>       siteMonitor: true
> ```
> image
> ---
> 
> ```yaml
>   - dummyService:
>       siteMonitor: true
> ```
> image
> ```
> docker compose logs homepage
> homepage  | Listening on port 3000
> homepage  | [2024-05-20T13:12:22.578Z] error: <siteMonitor> Monitoring requested for service 'dummyService' but no url specified.
> homepage  | 	Either set monitor to a url or set href
> ```

## Type of change

- [ ] ~New service widget~
- [ ] ~Bug fix (non-breaking change which fixes an issue)~
- [ ] ~Documentation only~
- [ ] ~Other (please explain)~

## Checklist:

### DIFF
--- a/docs/configs/services.md
+++ b/docs/configs/services.md
@@ -125,6 +125,8 @@ You can also apply different styles to the ping indicator by using the `statusSt
 
 Services may have an optional `siteMonitor` property (formerly `ping`) that allows you to monitor the availability of a URL you chose and have the response time displayed. You do not need to set your monitor URL equal to your href or ping URL.
 
+If you set the `siteMonitor` property to `true`, the monitored URL be the same as `href`.
+
 !!! note
 
     The site monitor feature works by making an http `HEAD` request to the URL, and falls back to `GET` in case that fails. It will not, for example, login if the URL requires auth or is behind e.g. Authelia. In the case of a reverse proxy and/or auth this usually requires the use of an 'internal' URL to make the site monitor feature correctly display status.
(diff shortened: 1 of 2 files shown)


shamoon:
Thank you for the PR. I like the idea but as stated in the guidelines we require a feature request (that has been up-voted) first. Since we ask this of all PRs, it’s only fair that we hold all contributors to the same standards.

Please do so, and we’d be happy to re-address at a later point.

Thanks again for your interest in the project

neBM:
Hi @shamoon,

Thank you for the feedback! I'm glad you like what you see. I can absolutely raise a feature request for this.

I'm only new to contributing to opensource so I'm still learning the ropes and any feedback is greatly appreciated.

> Raised #3501

🤞🏻

shamoon:
Thanks. Every project has their own rules, we did that here just because we were getting a lot of overly-specific PRs (ie basically only for 1 person) and wanted to try and make sure to use our resources wisely.

PR #3617, opened 2024-06-09
Feature: stock market service and info widget

eldyl:
## Proposed change

Hello!

I have been working on an implementation of a stock market widget.

Relevant Feature Request -  https://github.com/gethomepage/homepage/discussions/297

widget_stocks_screenshot

homepage_widget_stocks_demo

## Type of change

- [x] New service widget
- [ ] Documentation only
- [x] Other (please explain) **Information widget**

## Checklist:

## Notes
- I felt like this widget fit best as an info widget, but I am happy to change things if the category is not appropriate.
- My goal was to show basic information in a simple way.
- Uses [Finnhub.io's](https//finnhub.io) API. You can sign up for a free API key.
	- [Finnhub allows up to 30 calls per second](https://finnhub.io/docs/api/rate-limit). I set the max items in the watchlist to 30 for this reason. This is a very high number, but I figured I would leave this up to the user.
	- [Finnhub allows up to 60 calls per minute](https://finnhub.io/pricing) on the free tier. I set default caching for this widget to 1 minute. Caching can be changed in the config.
- You can click on the widget to switch between the current price of a stock and the change in price of the stock since open.
- I created the widget as the "Stocks" widget rather than "Finnhub". The intention was to allow different APIs as providers. It should be relatively straightforward to implement more APIs.

Desktop

homepage_widget_stocks_desktop

Mobile

homepage_widget_stocks_mobile


shamoon (changes requested):
Thanks, this is great overall and I think people will be happy with it. Especially like that its extensible.

A couple small things before I double-check the code is that as-is this will allow duplicates in the list (thats prob OK, people can do crazy things, then again it'd be easy enough to solve) but Im also wondering if the limit of 30 (which I know comes from the API) is too much. It just starts to get too large. Maybe thats OK but my gut tells me we should trim it down. What do you think? 

Screenshot from 2024-06-09 18-56-25

eldyl:
I appreciate your feedback!

> as-is this will allow duplicates in the list 

I will take care of this.

> Im also wondering if the limit of 30 (which I know comes from the API) is too much. It just starts to get too large. Maybe thats OK but my gut tells me we should trim it down. What do you think?

Example with 6 Items


Example with 8 Items



Agreed. A max of 6 kind of feels like the sweet spot.  With a known max, I would also like to handle some viewport sizes a little differently. 

If it works, I can have these updates included tomorrow.

If there is anything else that looks like it could be improved, I am happy to make the changes.

shamoon:
Yea no problem, sounds good on all accounts. I’ll take a look at the code more closely after you push the changes but overall seems like pretty a clean PR thanks 👍

brhahlen:
Yay! Love this!
My only question would be if it would be possible to make a service widget for it? Would fit better with the CMC widget as welll :)
Thanks for the hard work!

eldyl:
> Yay! Love this! My only question would be if it would be possible to make a service widget for it? Would fit better with the CMC widget as welll :) Thanks for the hard work!

I could do both.

The component looks like it is pretty easily adaptable:
https://github.com/gethomepage/homepage/blob/main/src/widgets/coinmarketcap/component.jsx 

@shamoon What are your thoughts? Would it be better to pick one? Or would it be reasonable to have both component types as potential options?

shamoon:
It's not clear from the FR if people want a service widget or info, so it's of course debatable. But given this is closest to the CoinMarketCap widget (which btw is better suited to the list format) perhaps that is the better way to go, sorry for the changeup.

eldyl:
Happy to make the change. I agree that the format of the CoinMarketCap widget does work better overall. I can get started later today.

benphelps:
I think both should be options.

On the layout of the info widget, I think keeping the font size the same, and lowering the opacity of the price, would be more inline with the Homepage design direction.  This would also allow it to use slightly less vertical space.

eldyl:
> I think both should be options.

For my personal use, I do prefer having a few items in the header in most instances.

> On the layout of the info widget, I think keeping the font size the same, and lowering the opacity of the price, would be more inline with the Homepage design direction. This would also allow it to use slightly less vertical space.

I will look into these changes. Thanks for the tips.

---

Apologies for the delay. I have been busy with work :)

I was testing some other APIs to see if it makes sense to include them in the widget, but I think, at least for now, Finnhub seems like a better option with their free tier and limits.

I also have tested getting data from `https://finance.yahoo.com/quotes/${comma_separated_list_of_tickers}`. This would work without an API key, but will require some more processing. I have been testing headers to reduce the size of the received payload. 

I suggested this could be a `Stocks`, widget, and I'd like to make that work, but I would like to provide proof of concept with multiple APIs. Otherwise, it might make sense for for the Widget to be a `Finnhub` widget. I also want to make sure it won't be difficult to maintain multiple APIs. That being said, if changes in the future are accepted, I am happy to help maintain them.

shamoon:
I would just focus on this version, especially given it will now support two versions of the widget let’s keep the PR focused ie MVP. I’m confident this can be adapted for other APIs if there is need

eldyl:
> I think both should be options.
> 
> On the layout of the info widget, I think keeping the font size the same, and lowering the opacity of the price, would be more inline with the Homepage design direction. This would also allow it to use slightly less vertical space.

Maybe something closer to this?

homepage_stock_info_preview_1

arminius-smh:
> Example with 6 Items 
> 
> Example with 8 Items 
> 
> Agreed. A max of 6 kind of feels like the sweet spot. With a known max, I would also like to handle some viewport sizes a little differently.

Could this be reconsidered? For example, I don't use the resource monitors and have only the uptime there, so I have more space. 

The greeting info also allows me to write a paragraph in 4xl size in there if I want, I'd be cool if I want to have my header cluttered with stock info to be able to do that.

eldyl:
I will write with some more information in the morning regarding some of the choices I made, but this should be ready for review.

mobile
desktop

docs/widgets/info/stocks.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Stocks
+description: Stocks Information Widget Configuration
+---
+
+The Stocks Inormation Widget allows you to include basic stock market data in
```
eldyl, line 6: typo here

shamoon:
Looking good. Small thing, I wouldn’t bother with displaying (or translating) the errors or displaying things like has duplicates, it’s basically a config error so just throw an error with details in the log and / or truncate the list / remove the duplicates.

src/components/widgets/stocks/stocks.jsx
```diff
@@ -0,0 +1,102 @@
...
+      <Container options={options} additionalClassNames="information-widget-error">
+        <PrimaryText>{t(data?.error)}</PrimaryText>
+        <WidgetIcon icon={BiError} size="l" />
+      </Container>
+    );
+  }
+
+  if (error) {
+    return <Error options={error} />;
+  }
```
shamoon, line 35:
```suggestion
  if (data?.error || error) {
    return (
      <Container options={options} additionalClassNames="information-widget-error">
        <PrimaryText>{error ? error : data?.error}</PrimaryText>
        <WidgetIcon icon={BiError} size="l" />
      </Container>
    );
  }
```

Combine these two, also cant translate the error

src/components/widgets/stocks/stocks.jsx
```diff
@@ -0,0 +1,102 @@
...
+        )}
+      </div>
+    ));
+
+    return (
+      <Container>
+        <Raw>
+          <button
+            type="button"
+            onClick={() => (viewingPercentChange ? setViewingPercentChange(false) : setViewingPercentChange(true))}
```
shamoon, line 88:
```suggestion
            onClick={() => setViewingPercentChange(!viewingPercentChange)}
```

src/components/widgets/stocks/stocks.jsx
```diff
@@ -0,0 +1,102 @@
...
+            className="flex items-center w-full h-full hover:outline-none focus:outline-none"
+          >
+            <FaChartLine className="flex-none w-5 h-5 text-theme-800 dark:text-theme-200 mr-2" />
+            <div className="flex flex-wrap items-center gap-0.5">{stocks}</div>
```
shamoon, line 92: I think its a little more react-y to move the above `map` block to here, rather than save it to a const which is used once. Not a big thing, of course

src/components/widgets/stocks/stocks.jsx
```diff
@@ -0,0 +1,102 @@
...
+          </button>
+        </Raw>
+      </Container>
+    );
+  }
+}
+
+export default function Stocks({ options }) {
+  return <Widget options={{ ...options }} />;
```
shamoon, line 101: Same sorta thing, why not just move `Widget` component in here?

src/pages/api/widgets/stocks.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  if (watchlistArr.length > 8) {
+    return res.status(400).json({ error: "stocks.maxItemsReached" });
+  }
+
+  const hasDuplicates = new Set(watchlistArr).size !== watchlistArr.length;
+
+  if (hasDuplicates) {
+    return res.status(400).json({ error: "stocks.hasDuplicates" });
```
shamoon, line 24: see https://github.com/gethomepage/homepage/pull/3617#issuecomment-2199350926

src/pages/api/widgets/stocks.js
```diff
@@ -0,0 +1,73 @@
+import cachedFetch from "utils/proxy/cached-fetch";
+import { getSettings } from "utils/config/config";
+
+export default async function handler(req, res) {
+  const { watchlist, provider, cache } = req.query;
+
+  if (!watchlist) {
+    return res.status(400).json({ error: "stocks.missingWatchlist" });
```
shamoon, line 8: see https://github.com/gethomepage/homepage/pull/3617#issuecomment-2199350926

src/widgets/stocks/component.jsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+
+  if (watchlistHasDuplicates) {
+    return (
+      <Container service={service}>
+        <Block value={t("stocks.hasDuplicates")} />
+      </Container>
+    );
+  }
```
shamoon, line 136: See https://github.com/gethomepage/homepage/pull/3617#issuecomment-2199350926
Also I think these can be combined

src/widgets/stocks/component.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+    return {
+      ticker,
+    };
+  });
+
+  dupeCheckMap = null;
+
+  return { watchlistHasDuplicates, parsedWatchlist };
+}
```
shamoon, line 29: Why not the same as the other one, just using `Set`?

shamoon (changes requested):
I havent tested yet, just some thoughts looking at the code. Thanks

eldyl:
Reviewing now. Thank you for these great  suggestions. I will try and get them all added today.

src/widgets/stocks/component.jsx
eldyl: I will change this to act similarly as the other check. I was going on a journey to add some more features, but then I thought KISS.

I was looking at adding quantities to the service widget. I could add something like this in a future PR, but I have tabled it for now.

This is what it looked like going down that road: 

```js
function parseWatchlist(watchlist) {
  let watchlistNumberOfFieldsWithQuantity = 0;
  let watchlistHasDuplicates = false;
  let dupeCheckMap = new Map();

  // Parse quantity if it exists, and check for duplicates
  const parsedWatchlist = watchlist.map((ticker) => {
    const tickerAndQuantity = ticker.split("*");

    const tickerSymbol = tickerAndQuantity[0];
    const quantity = tickerAndQuantity[1] ? tickerAndQuantity[1] : null;

    if (quantity) watchlistNumberOfFieldsWithQuantity += 1;

    if (dupeCheckMap.has(tickerSymbol)) {
      watchlistHasDuplicates = true;
    } else {
      dupeCheckMap.set(tickerSymbol);
    }

    return {
      ticker: tickerSymbol,
      quantity,
    };
  });

  dupeCheckMap = null;

  return { watchlistHasDuplicates, watchlistNumberOfFieldsWithQuantity, parsedWatchlist };
}
```

shamoon: Yea good call, this is a big PR already. Get it shipped, watch for bugs, iterate if there's interest =)

src/components/widgets/stocks/stocks.jsx
eldyl: Should I simplify to use the Error component? 

```
  if (error || data?.error) {
    return <Error options={options} />;
  }

```

https://github.com/gethomepage/homepage/blob/810c321881d9acdde3666c4cac0fd6700d12735b/src/components/widgets/openmeteo/openmeteo.jsx#L20 

https://github.com/gethomepage/homepage/blob/810c321881d9acdde3666c4cac0fd6700d12735b/src/components/widgets/longhorn/longhorn.jsx#L15

https://github.com/gethomepage/homepage/blob/810c321881d9acdde3666c4cac0fd6700d12735b/src/components/widgets/glances/glances.jsx#L33

shamoon: Yea sounds good

eldyl:
> I havent tested yet, just some thoughts looking at the code. Thanks

Mentioned issues should be resolved: https://github.com/gethomepage/homepage/compare/878d62b1f832f17bb57d18ac3b00ad70dec0c2df..ae3df9eeeb045f256bce33751d1d504ca96cf6cd

shamoon (approved):
Great work, I think people will dig this. some minor tweaks but nothing major
Screenshot 2024-07-01 at 5 06 49 PM
Screenshot 2024-07-01 at 5 06 57 PM

eldyl:
> Great work, I think people will dig this. some minor tweaks but nothing major Screenshot 2024-07-01 at 5 06 49 PM Screenshot 2024-07-01 at 5 06 57 PM

Thanks again for your feedback throughout. I checked your changes, and I made some notes for myself as well.

Thanks!

rkmd:
@eldyl - thank you for this wonderful addition to Homepage! Similar to what you've done in the widget box with displaying the percentage price change (instead of the actual price), could this be an option for the information option in the top bar?

Just a suggestion!

suicidaleggroll:
> Is it possible to add crypto?

FYI - the implemented stock tracker already supports crypto.  While not explicitly called out, you can use "ETH-USD" or "BTC-USD", etc. for your tickers and it works correctly.

```
- stocks:
    provider: finnhub
    color: true
    cache: 1
    watchlist:
      - VOO
      - COIN
      - ETH-USD
      - BTC-USD
```

eldyl:
> FYI - the implemented stock tracker already supports crypto. While not explicitly called out, you can use "ETH-USD" or "BTC-USD", etc. for your tickers and it works correctly.
> 
> ```
> - stocks:
>     provider: finnhub
>     color: true
>     cache: 1
>     watchlist:
>       - VOO
>       - COIN
>       - ETH-USD
>       - BTC-USD
> ```
> 

I appreciate you sharing this! I was not aware these symbols received proper values from the Finnhub API. I think it would be a good idea to put in a PR for the widget documentation.

```
- stocks:
    provider: finnhub
    color: true # optional, defaults to true
    cache: 1 # optional, default caches results for 1 minute
    watchlist:
      - GME
      - AMC
      - NVDA
      - AMD
      - TSM
      - MSFT
      - BTC-USD
      - ETH-USD
```

Something like the above might help point out how you can query some crypto.

I will have some time within the next week to take a closer look at this.

I hope you are well.

Danie10:
Hi @eldyl this is working for me, but I'm not sure if I'm missing something for non-US stocks (syntax) or whether I must open a new issue for a feature. I'm looking to for example list a stock STX40 from Johannesburg stock exchange. Finnhub gives the exchange the code JO. But I tried adding STX40.JO and it gives an API error.

suicidaleggroll:
> > FYI - the implemented stock tracker already supports crypto. While not explicitly called out, you can use "ETH-USD" or "BTC-USD", etc. for your tickers and it works correctly.
> > ```
> > - stocks:
> >     provider: finnhub
> >     color: true
> >     cache: 1
> >     watchlist:
> >       - VOO
> >       - COIN
> >       - ETH-USD
> >       - BTC-USD
> > ```
> 
> I appreciate you sharing this! I was not aware these symbols received proper values from the Finnhub API. I think it would be a good idea to put in a PR for the widget documentation.
> 
> ```
> - stocks:
>     provider: finnhub
>     color: true # optional, defaults to true
>     cache: 1 # optional, default caches results for 1 minute
>     watchlist:
>       - GME
>       - AMC
>       - NVDA
>       - AMD
>       - TSM
>       - MSFT
>       - BTC-USD
>       - ETH-USD
> ```
> 
> Something like the above might help point out how you can query some crypto.
> 
> I will have some time within the next week to take a closer look at this.
> 
> I hope you are well.

After using this for a bit, I can't recommend the crypto tickers from finnhub unfortunately.  It does work, and during the day the numbers are correct, but after the markets close (before 9:30am or after 4pm Eastern) the crypto prices get wacky.  Sometimes it just throws an API error, sometimes it shows the "close" or "open" price of that day, sometimes it shows a price from 2-3 days ago.  It doesn't seem to follow any pattern that I can decipher.  I have noticed that the ticker values on finnhub's website still appear to be correct, so maybe it's just a problem with the API outside of standard trading hours.

PR #3657, opened 2024-06-21
Fix: Improve error handling for Glances widgets

mjsully:
## Proposed change

I have the Glances widgets deployed in my Homepage instance and noticed that when a host is unreachable the widgets do not handle the errors as the other widgets do. It results in an ugly error message which doesn't respect either the global or service-level `hideErrors`. Here is an example from my deployment:

glances-widget-bug

I have rewritten parts of the components such that the `Error` component is no longer needed, as it is now implemented within the `Container` component. The service and error attributes are now passed into the `Container` component, such that the service-level `hideErrors` flag can be checked. Additionally, the global settings are now imported into the `Container` component so the global `hideErrors` flag can be checked. Within each component, if the service data containers an "error" key,  the widget will now display an error message in line with other widgets. See the screenshot below:

glances-widget-fix-showErrors

Additionally, see the below screenshot testing the service-level `hideErrors` flag set to `true` for one widget:

glances-widget-fix-hideErrors

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/glances/components/container.jsx
+++ b/src/widgets/glances/components/container.jsx
@@ -1,19 +1,19 @@
 import { useContext } from "react";
 import { useTranslation } from "next-i18next";
 import { SettingsContext } from "utils/contexts/settings";
-import { Settings } from "luxon";
+import Error from "./error"
 
 export default function Container({ service, children, chart = true, error = false, className = "" }) {
   const { t } = useTranslation();
   const { settings } = useContext(SettingsContext);
-  const hideErrors = (service.widget.hide_errors || settings.hideErrors)
+  const hideErrors = (settings.hideErrors || service.widget.hide_errors)
   return (
     <div>
       {children}
       <div className={`absolute top-0 right-0 bottom-0 left-0 overflow-clip pointer-events-none ${className}`} />
       {chart && <div className="h-[68px] overflow-clip" />}
       {!chart && <div className="h-[16px] overflow-clip" />}
-      {error && !hideErrors && <div className="absolute bottom-2 left-2 z-20 text-red-400 text-xs opacity-75">{t("widget.api_error")}</div>}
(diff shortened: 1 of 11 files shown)


shamoon (changes requested):
I can take a real look at this soon but my first thought is that the logic to respect errors should just be moved inside Error, instead of all the other changes for removing it. and in info and gpu where you check for the error property presumably that should become the error for the Error component

shamoon:
Sure, but separating them logically makes sense and that's also how the rest of the app does it. Also would significantly reduce the diff here.

Let us know if youre OK making the change, I can also take care of it if thats better.

mjsully:
Sure, let me make the changes and commit.

mjsully:
I've now reintroduced the error component and import it within the container component. Following the implementation style in e.g. the Home Assistant widget https://github.com/gethomepage/homepage/blob/ec448d6c41ffa7c90c4a180a972eff309ebbcf3d/src/widgets/homeassistant/component.jsx#L10 the error handling is still done within the container component. Let me know if you want any other changes making.

shamoon:
That wasnt exactly what I had in mind. Please take a look at the changes here. (also FYI https://eslint.org/docs/latest/rules/no-prototype-builtins )

shamoon:
Edit: and let me know if you have any thoughts or concerns

mjsully:
Thanks for this, and thanks for the link that's good to know. As for the changes, it seems like there isn't a consistent implementation of the error handling (e.g. the Home Assistant ref I provided was similar to my own implementation but different to the one you have provided in the changes.) - either way the result will be the same and looks good to me.

shamoon:
Point taken, I've gone back to a bit of a mix

shamoon (approved):
Ok, thanks again

PR #3808, opened 2024-08-02
Fix/main workflow update

discretizer:
* Fix workflow for main branch


discretizer:
I hate github forks default behavior

PR #3826, opened 2024-08-07
Add custom arguments for the gamedig query

teooko:
## Proposed change

I reviewed [this feature request](https://github.com/gethomepage/homepage/discussions/2679) and tried to add custom arguments to the gamedig query according to [this format](https://github.com/gethomepage/homepage/discussions/2679#discussioncomment-8236622). I am unsure if this is the proper way of doing it, but I would appreciate some guidance if necessary.


shamoon:
This will not work, no. Did you test it? Generally a good idea before you PR something.

Unless specifically included options like this will get stripped out. There is very little interest on the FR, so we can reevaluate if that changes but this will likely need to be more explicit as well for security, not just allowing any arbitrary set of arguments.

thanks for your interest in the project

PR #3886, opened 2024-08-23
Refactor: change RomM API endpoint

InsertDisc:
## Proposed change

An update to the currently existing RomM widget. This update changes the endpoint from the currently set 'platforms' endpoint to the more concise and cleaner 'stats' endpoint. The stats endpoint gives three additional fields of information that the existing two give, also without needing the calculations of the existing endpoint. Documentation has been updated to reflect the changes as well.

### DIFF
--- a/docs/widgets/services/romm.md
+++ b/docs/widgets/services/romm.md
@@ -3,7 +3,7 @@ title: Romm
 description: Romm Widget Configuration
 ---
 
-Allowed fields: `["platforms", "totalRoms", "totalSaves", "totalStates", "totalScreenshots", "totalFilesize"]`.
+Allowed fields: `["platforms", "roms", "saves", "states", "screenshots", "totalfilesize"]`.
 
 ```yaml
 widget:
@@ -11,5 +11,5 @@ widget:
   url: http://romm.host.or.ip
   username: username # optional
   password: password # optional
-  fields: ["platforms", "totalRoms", "totalSaves", "totalScreenshots"]
+  fields: ["platforms", "roms", "saves", "screenshots"]
 ```
--- a/docs/widgets/services/romm.md
+++ b/docs/widgets/services/romm.md
@@ -3,7 +3,7 @@ title: Romm
 description: Romm Widget Configuration
 ---
 
-Allowed fields: `["platforms", "roms", "saves", "states", "screenshots", "totalfilesize"]`.
+Allowed fields: `["platforms", "totalRoms", "saves", "states", "screenshots", "totalfilesize"]`.
 If more than (4) fields are provided, only the first (4) will be used.
 
 ```yaml
@@ -12,5 +12,5 @@ widget:
   url: http://romm.host.or.ip
   username: username # optional
   password: password # optional
-  fields: ["platforms", "roms", "saves", "states"] # optional - default fields shown
+  fields: ["platforms", "totalRoms", "saves", "states"] # optional - default fields shown
 ```
--- a/docs/widgets/services/romm.md
+++ b/docs/widgets/services/romm.md
@@ -4,6 +4,7 @@ description: Romm Widget Configuration
 ---
 
 Allowed fields: `["platforms", "roms", "saves", "states", "screenshots", "totalfilesize"]`.
+If more than (4) fields are provided, only the first (4) will be used.
 
 ```yaml
 widget:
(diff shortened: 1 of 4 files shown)


gantoine (approved):

src/widgets/romm/component.jsx
shamoon, line 1: Please dont bother with `const`s that get used only once

InsertDisc: Thanks for the advice.

shamoon (changes requested):
This isnt "your fault" but I think we should take the time to fix it. This widget is not constructed correctly.

The blocks should be basically:

```
if error {}
if !response {}
response {}
```

In particular there is a redundant responseError block here, and no block for !response. In the case of no response it should just return the container with blocks but no values. This is what makes homepage look like its "waiting" until data is returned.

Let me know if that makes sense. If you'd prefer not to make these changes NBD, I can take care of it of course.

Thanks for the PR

InsertDisc:
> This isnt "your fault" but I think we should take the time to fix it. This widget is not constructed correctly.
> 
> The blocks should be basically:
> 
> ```
> if error { container with error }
> if !response { container with empty blocks }
> normal container
> ```
> 
> In particular there is a redundant responseError block here, and no block for !response. In the case of no response it should just return the container with blocks but no values. This is what makes homepage look like its "waiting" until data is returned.
> 
> Let me know if that makes sense. If you'd prefer not to make these changes NBD, I can take care of it of course.
> 
> Thanks for the PR

Added the empty blocks, removed the extra error handling, simplified label names, and updated local (en). I'm not entirely sure how the locale process is supposed to work, so apologies in advance.

InsertDisc:
> This isnt "your fault" but I think we should take the time to fix it. This widget is not constructed correctly.
> 
> The blocks should be basically:
> 
> ```
> if error { container with error }
> if !response { container with empty blocks }
> normal container
> ```
> 
> In particular there is a redundant responseError block here, and no block for !response. In the case of no response it should just return the container with blocks but no values. This is what makes homepage look like its "waiting" until data is returned.
> 
> Let me know if that makes sense. If you'd prefer not to make these changes NBD, I can take care of it of course.
> 
> Thanks for the PR

I added the empty blocks, removed the extra error handling, simplified label names, updated the readme with those changes, and edited the en locale. I'm not entirely sure how the locale process works so apologies in advance.

shamoon (changes requested):
Great thank you. One last thing sorry I didn't notice! Widgets should not have more than four blocks at a time. So this always requires a little bit of logic in terms of the defaults and limiting the total. You can look at other widgets that do that but it's relatively simple.

InsertDisc:
> Great thank you. One last thing sorry I didn't notice! Widgets should not have more than four blocks at a time. So this always requires a little bit of logic in terms of the defaults and limiting the total. You can look at other widgets that do that but it's relatively simple.

So you're saying there can be more than 4 available fields as long as it's limited to four? If so, do you know of an existing widget I could use as a guide?

shamoon:
So like: https://github.com/gethomepage/homepage/blob/0d25f5789b2b46b8adb85eb31883f31fe07f61df/src/widgets/gamedig/component.jsx#L16-L25

docs/widgets/services/romm.md
```diff
@@ -3,12 +3,13 @@ title: Romm
 description: Romm Widget Configuration
 ---
 
-Allowed fields: `["platforms", "totalRoms"]`.
+Allowed fields: `["platforms", "roms", "saves", "states", "screenshots", "totalfilesize"]`.
```
shamoon, line 6: One more thing, is `totalRoms` the same as `roms`? If so, please leave it as `totalRoms`, otherwise this is a breaking change. If its different we should mention that the old one has been removed.

InsertDisc: Oh, yes. I overlooked that. I will revert those changes. Once that is done I think we're good.

InsertDisc: Field name has been reverted in component, docs, and locale.

InsertDisc:
I think we should be all good now. Please let me know if further changes are needed. Thank you for your patience working with me through these changes.

shamoon (approved):
Great thanks for all the work

Screenshot 2024-08-23 at 7 25 25 PM

PR #3895, opened 2024-08-24
Chore: handle mealie API change

joncrangle:
## Proposed change

Mealie is introducing breaking API changes that are already merged in the "nightly" builds. See https://github.com/mealie-recipes/mealie/pull/3970. The `groups/statistics` endpoint is being deprecated and replaced with a `households/statistics` API.

I modelled the API handling on the Immich implementation https://github.com/gethomepage/homepage/pull/2284. If the version starts with "v2", or equals "nightly" it will use the new endpoint. Otherwise, it will keep using the existing endpoint to ensure this is non-breaking for people who pin to a v1 build.

Closes #

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/mealie.md
+++ b/docs/widgets/services/mealie.md
@@ -14,5 +14,5 @@ widget:
   type: mealie
   url: http://mealie-frontend.host.or.ip
   key: mealieapitoken
-  version: 1 # optional, defaults to 1
+  version: 2 # only required if version > 1, defaults to 1
 ```
--- a/docs/widgets/services/mealie.md
+++ b/docs/widgets/services/mealie.md
@@ -14,4 +14,5 @@ widget:
   type: mealie
   url: http://mealie-frontend.host.or.ip
   key: mealieapitoken
+  version: 1 # optional, defaults to 1
 ```
--- a/src/utils/config/service-helpers.js
+++ b/src/utils/config/service-helpers.js
@@ -402,7 +402,7 @@ export function cleanServiceGroups(groups) {
           // frigate
           enableRecentEvents,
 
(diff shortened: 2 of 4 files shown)


shamoon (changes requested):
Thanks. We’re going to start moving away from doing it this way because this will cause errors in one case or the other while the version call is pending.

Instead, the way to do this is use the `version` parameter in the widget config. And to be non-breaking it’ll need to default to the current one.

Not as “clean” for sure, but for widgets that don’t use a proxy it’s how we’ll handle stuff like this.

Let me know if that makes sense, and if you can make the changes

docs/widgets/services/mealie.md
```diff
@@ -14,4 +14,5 @@ widget:
   type: mealie
   url: http://mealie-frontend.host.or.ip
   key: mealieapitoken
+  version: 1 # optional, defaults to 1
```
shamoon, line 17:
```suggestion
  version: 2 # only required if version > 1, defaults to 1
```

shamoon (approved):
Thanks for the PR.

Note: this didn't actually work without my last commit there, always good to test things out. (I did of course):

Screenshot 2024-08-24 at 8 55 38 PM
Screenshot 2024-08-24 at 9 00 47 PM

docs/widgets/services/mealie.md
mcmasterp: I see no effect when setting this to version: 2. I'm trying to force the widget to retrieve stats but it will only use the v1 groups endpoint

Issue #1152, opened 2023-03-27
[Feature Request] Support for Technitium


iceman73:
### Description

Support for  https://github.com/TechnitiumSoftware/DnsServer/blob/master/APIDOCS.md#dashboard-api-calls

PR #3904, opened 2024-08-28
Feature: Technitium DNS Widget

bobby-driggs:
## Proposed change

Adds a new service widget for Technitium DNS Server, based on their [Dashboard API](https://github.com/TechnitiumSoftware/DnsServer/blob/master/APIDOCS.md#dashboard-api-calls)

The above API call takes the format of: `<server ip / domain>/api/dashboard/stats/get?token=<api-token>&type=<time_range>&utc=true`

The following query string parameters are used
- `token`: Is the api token that can be created on the Technitium DNS dashboard, this is a permanant token that wont expire, that is specifically used to API calls.
- `type`: Is really a date range, and can be one of: `[LastHour, LastDay, LastWeek, LastMonth, LastYear]`

The response of this API call returns a lot of information, but the one this widget cares about, is the `response.stats` property that has the following data:

```json
{
	"response": {
		"stats": {
			"totalQueries": 52321,
			"totalNoError": 49112,
			"totalServerFailure": 107,
			"totalNxDomain": 3102,
			"totalRefused": 0,
			"totalAuthoritative": 3393,
			"totalRecursive": 36952,
			"totalCached": 11976,
			"totalBlocked": 0,
			"totalDropped": 0,
			"totalClients": 4,
			"zones": 11,
			"cachedEntries": 15867,
			"allowedZones": 0,
			"blockedZones": 0,
			"allowListZones": 0,
			"blockListZones": 0
		},
	// additional properties removed for brevity ...
}
```

The above block is what gets translated to the data the component gets using the `useWidgetAPI` call.

I have a max 4 fields set as a default (as i saw utilized in other widgets), and a user can select which fields they would like to use.

Here is an example of its service configuration:
```yaml
    - Technitium:
        icon: technitium.png
        href: https://tdns.example.org
        description: T-DNS
        widget:
            type: technitium
            url: https://tdns.example.org
            key: <api_token>
            range: LastDay # optional, defaults to LastHour
```

Here is a screenshot of how i would personally use it, next to my Pi Hole widgets:

image

And here is a screenshot of the actual Technitium DNS dashboard for comparison:

image

I saw Technitium was mentioned back in 2023, with a handful of upvotes, so I thought Id take a stab at it.

Related issue: https://github.com/gethomepage/homepage/issues/1152
Related discussion: https://github.com/gethomepage/homepage/discussions/1319 (20 upvotes as of 8/2024)

## Type of change

- [X] New service widget

## Checklist:

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -323,6 +323,19 @@
         "seconds": "{{number}}s",
         "ago": "{{value}} Ago"
     },
+    "technitium": {
+        "totalQueries": "Queries",
+        "totalNoError": "Success",
+        "totalServerFailure": "Failures",
+        "totalNxDomain": "NX Domains",
+        "totalRefused": "Refused",
+        "totalAuthoritative": "Authoritative",
+        "totalRecursive": "Recursive",
+        "totalCached": "Cached",
+        "totalBlocked": "Blocked",
+        "totalDropped": "Dropped",
+        "totalClients": "Clients"
+    },
     "tdarr": {
         "queue": "Queue",
         "processed": "Processed",
--- a/src/widgets/technitium/widget.js
+++ b/src/widgets/technitium/widget.js
@@ -1,0 +1,17 @@
+import genericProxyHandler from "utils/proxy/handlers/generic";
+import { asJson } from "utils/proxy/api-helpers";
+
+const widget = {
+  api: "{url}/api/{endpoint}?token={key}&utc=true",
+  proxyHandler: genericProxyHandler,
+  mappings: {
+    stats: {
+      endpoint: "dashboard/stats/get",
+      validate: ["response", "status"],
+      params: ["type"],
+      map: (data) => asJson(data).response.stats,
+    },
+  },
+};
+
+export default widget;
--- a/docs/widgets/services/index.md
+++ b/docs/widgets/services/index.md
@@ -113,6 +113,7 @@ You can also find a list of all available service widgets in the sidebar navigat
 - [Syncthing Relay Server](syncthing-relay-server.md)
 - [Tailscale](tailscale.md)
 - [Tandoor](tandoor.md)
+- [Technitium DNS](technitium.md)
 - [TDarr](tdarr.md)
 - [Traefik](traefik.md)
 - [Transmission](transmission.md)
--- a/docs/widgets/services/technitium.md
+++ b/docs/widgets/services/technitium.md
@@ -24,5 +24,3 @@ This can be generated via the Technitium DNS Dashboard, and should be generated 
 #### Range
 
 `range` value determines how far back of statistics to pull data for. The value comes directly from Technitium API documentation found [here](https://github.com/TechnitiumSoftware/DnsServer/blob/master/APIDOCS.md#dashboard-api-calls), defined as `"type"`. The value can be one of: `LastHour`, `LastDay`, `LastWeek`, `LastMonth`, `LastYear`.
-
-**NOTE:** The value of `Custom` as specified in the API documentation is _NOT_ supported.
--- a/docs/widgets/services/technitium.md
+++ b/docs/widgets/services/technitium.md
@@ -1,0 +1,28 @@
+---
+title: Technitium DNS Server
+description: Technitium DNS Server Widget Configuration
+---
+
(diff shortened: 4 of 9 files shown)


src/widgets/technitium/widget.js
```diff
@@ -0,0 +1,16 @@
...
+import { asJson } from "utils/proxy/api-helpers";
+
+const widget = {
+  api: "{url}/api/{endpoint}/get?token={key}&type=LastHour",
+  proxyHandler: genericProxyHandler,
+  mappings: {
+    stats: {
+      endpoint: "dashboard/stats",
+      validate: ["response", "status"],
+      map: (data) => asJson(data).response.stats,
```
bobby-driggs, line 11: I tried specifying `params` here as specified here: https://github.com/gethomepage/homepage/blob/main/docs/widgets/authoring/metadata.md#params (and in the API block), but i couldnt get it to work.

I had changed the above roughly to:
```
  api: "{url}/api/{endpoint}/get?token={key}",
  proxyHandler: genericProxyHandler,
  mappings: {
    stats: {
      endpoint: "dashboard/stats",
      validate: ["response", "status"],
      params: ["type", "utc"],
      map: (data) => asJson(data).response.stats,
```
and was passing in the params via the `useWidgetAPI` like so:

```
  const params = {
    type: widget.range ?? 'LastHour',
    utc: widget.utc ?? false,
  }

  const { data: statsData, error: statsError } = useWidgetAPI(widget, "stats", params);
  ```
  And was getting API Errors, with vague JSON error messaging.
  
  I had an inkling that perhaps the proxy wasnt smart enough to _add_ the `type` and the `utc` to the existing `token` query string?
  
  Is doing something like i tried possible, or is there a better way?
  
  Secondly, how does one specify custom parameters for widgets? I assumed if I have `range` and `utc` specified in the `service.yaml` it would pass through, but they were both marked undefined.
  
  I was thinking of even allowing a user to use the `range: custom` and then additionally specifing the `start` and `end` timespans, and forwarding that to the Technitium DNS API call via the `Custom` query string parameters i mentioned on the description of the PR.
  
  Thoughts?

File: public/locales/en/common.json [hunk 2]
```diff
@@ -323,6 +323,19 @@
...
+        "totalServerFailure": "Failures",
+        "totalNxDomain": "NX Domains",
+        "totalRefused": "Refused",
+        "totalAuthoritative": "Authoritative",
+        "totalRecursive": "Recursive",
+        "totalCached": "Cached",
+        "totalBlocked": "Blocked",
+        "totalDropped": "Dropped",
+        "totalClients": "Clients"
+    },
```
bobby-driggs, line 338: Is it appropriate to just have the English translations? or is there something else i should do with other languages?

src/widgets/technitium/widget.js
shamoon: Passing in params is problematic from a security standpoint but they are supported. But really, we prioritize simplicity over anything else. Remember that homepage isn't meant to be a replacement for the tool itself, just a small window into some interesting / helpful data.

Honestly, again, the extra options sounds over-complicated. Keep it simple.

Extra fields need to be included in service-helpers.js

public/locales/en/common.json
shamoon: This is fine.

src/widgets/technitium/widget.js
bobby-driggs: Sweet, appreciate the feed back. Ill try to get this cleaned up before the end of the long weekend for a proper review!

bobby-driggs:
Sorry for the noise, i setup pre-commit and things should be good now. also i amin a dev container and the `pnpm run lint` wasnt working for me. complaining about some library

shamoon (approved):
Thanks, I simplified this a bit. In general theres no reason to creat a var (or const) for something that gets used 1 time. Also this wasnt handling 0 cases appropriately (maybe not such a real world occurrence):

Screenshot 2024-08-29 at 10 35 54 AM

Also, the i18n common.percent formatter should be used.

Minor stuff really, otherwise works great:
Screenshot 2024-08-29 at 10 51 04 AM

IconicLight:
@bobby-driggs Thanks for this. Been waiting on a widget for Technitium. Appreciate your work on this. Works great!

PR #3937, opened 2024-09-03
Added Kubernetes CRDs

Rikpat:
## Proposed change

Added CRDs as an option of Kubernetes service definition. This is useful if you want to keep the service configuration in yaml and coupled with your services instead of updating configmap for homepage after every service deployment. 

It also allows for using secretKeyRef and configMapKeyRef for widgets, to keep your secrets secret instead of having them readable in annotation. This was the primary reason for creating this PR (also resolves #2685)

### Testing:

To use or test the functionality, you need to create the CRD (should be added to helm chart after PR completion).

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: homepageservices.gethomepage.dev
spec:
  group: gethomepage.dev
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                enabled:
                  type: boolean
                instances:
                  type: array
                  items:
                    type: string
                app:
                  type: string
                namespace:
                  type: string
                href:
                  type: string
                name:
                  type: string
                group:
                  type: string
                weight:
                  type: integer
                icon:
                  type: string
                description:
                  type: string
                external:
                  type: boolean
                podSelector:
                  type: string
                ping:
                  type: string
                siteMonitor:
                  type: string
                statusStyle:
                  type: string
                widget:
                  type: object
                  properties:
                    type:
                      type: string
                  required:
                  - type
                  x-kubernetes-preserve-unknown-fields: true
                  # x-kubernetes-validations: 
                  # - rule: "self.all(key, !(key.matches('.+From$')) || (has(self[key].secretKeyRef.name) && has(self[key].secretKeyRef.key)) || (has(self[key].configMapKeyRef.name) && has(self[key].configMapKeyRef.key)))"
                  #   message: Some widget property ends with From, but doesn't contain secretKeyRef nor configMapKeyRef
  scope: Namespaced
  names:
    plural: homepageservices
    singular: homepageservice
    kind: HomepageService
```

You also need to allow your serviceAccount to list and get the custom resources and get secrets/configMaps (don't include list for this, as you specify name directly). This is very simple with current helm chart, just use extraClusterRoles:

```
 extraClusterRoles:
    - apiGroups:
        - gethomepage.dev
      resources:
        - homepageservices
      verbs:
        - get
        - list
    - apiGroups:
        - ""
      resources:
        - secrets
        - configmaps
      verbs:
        - get
```

In case you would just want to allow reading in a single namespace, it needs to be done manually as there's no regular roles and roleassignments in the helm chart.

To simplify testing, I've also pushed the modified app to public ghcr repo: **ghcr.io/rikpat/homepage:v1**

Closes #2685

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/configs/kubernetes.md
+++ b/docs/configs/kubernetes.md
@@ -143,3 +143,41 @@ If the `href` attribute is not present, Homepage will ignore the specific Ingres
 ## Caveats
 
 Similarly to Docker service discovery, there currently is no rigid ordering to discovered services and discovered services will be displayed above those specified in the `services.yaml`.
+
+## CRDs
+
+Homepage also comes with Kubernetes CRDs for services. These CRDs have same structure and properties as regular service YAML definition, with added properties of `group`, `weight`, `podSelector` and `instances`, used as described above. 
+
+Compared to annotations, CRD approach can use Kubernetes secrets and configMaps to populate attributes of the `widget` object. To do this, instead of using the name, ex. `key`, use `keyFrom`, to match kubernetes standards. Then use either `secretKeyRef` or `configMapKeyRef` object, see example:
+
+```yaml
+apiVersion: gethomepage.dev/v1
+kind: HomepageService
+metadata:
+  labels:
+    app.kubernetes.io/instance: sonarr
+  name: sonarr
+  namespace: media
+spec:
+  description: TV Show Management Application
+  group: Media
+  href: 'https://sonarr.example.org/'
+  icon: sonarr.svg
+  widget:
+    type: sonarr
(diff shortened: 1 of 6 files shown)


shamoon:
Thanks but please see the guidelines. We can readdress this in the future if they are met.

Also, I maintain that kubernetes is just too darn overl-complicated…

PR #3984, opened 2024-09-12
Fix: Technitium widget percentage display

bobby-driggs:
## Proposed change

During the cleanup commit from the original PR, the multiply by 100 while calculating the percentage number was accidently removed. Original code: https://github.com/gethomepage/homepage/pull/3904/commits/d8e45a2fbcc9ae1474a5d826cd5133edcd43154c#diff-bd42a75b6a61d8468c8570e33b7f13dfffcf6d6dca4997677b4d303ef56daab6R55

On my v9.8 dashboard, you can see the problem: 

image

And on the test bed you can see the fix:

image

## Type of change

## Checklist:

### DIFF
--- a/src/widgets/technitium/component.jsx
+++ b/src/widgets/technitium/component.jsx
@@ -52,8 +52,9 @@ export default function Component({ service }) {
   }
 
   function toPercent(value, total) {
+    const percentage = (value / total) * 100;
     return t("common.percent", {
-      value: !Number.isNaN(value / total) ? value / total : 0,
+      value: !Number.isNaN(percentage) ? percentage : 0,
       maximumFractionDigits: 2,
     });
   }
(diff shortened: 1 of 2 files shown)


shamoon:
Sure, thanks, but why don't you just limit the diff to just those two lines

bobby-driggs:
> Sure, thanks, but why don't you just limit the diff to just those two lines

If you look at the diff in the `<Block>` statements, the `toPercent` function was expecting two arguments, the value and the total, but was never the total, because it was never properly being passed in. 

`${toPercent(statsData.totalServerFailure / statsData.totalQueries)`
instead of what i changed it to:
`${toPercent(statsData.totalServerFailure, statsData.totalQueries)`

The other way would have been to have kept the `value / total` calc in the block, and do the `* 100` on a value arg on in the `toPercent` but sort of made sense to pass them all into the function and do it there.

shamoon:
Ah yea gotcha thanks. Just tightened up a bit and I noticed a random other thing I figured we can put in here

shamoon (approved):

PR #4118, opened 2024-10-12
Feature: Vikunja service widget

vhsdream:
## Proposed change

Created a new service widget for the task management application [Vikunja](https://vikunja.io) (feature request #1630).

### Widget with default settings
image

### Widget with `enableTaskList` option set to `true`
image

### API calls output (2 API calls in total)

[projects.json](https://github.com/user-attachments/files/17348680/projects.json)
[tasks.json](https://github.com/user-attachments/files/17348707/tasks.json)

## Type of change

- [x] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If applicable, I have reviewed the [feature](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).


src/widgets/vikunja/component.jsx
```diff
@@ -0,0 +1,101 @@
+import { useTranslation } from "next-i18next";
+
+import Container from "components/services/widget/container";
+import Block from "components/services/widget/block";
+import useWidgetAPI from "utils/proxy/use-widget-api";
+
+// return Projects that are not Saved Filters
+function getRealProjects(project) {
```
shamoon, line 8: It is generally pointless to extract these filter functions to separate functions if you’re only using them once. It’ll be cleaner to just include them as anonymous functions

src/utils/config/service-helpers.js
```diff
@@ -633,6 +636,9 @@ export function cleanServiceGroups(groups) {
         if (type === "lubelogger") {
           if (vehicleID !== undefined) cleanedService.widget.vehicleID = parseInt(vehicleID, 10);
         }
+        if (type === "vikunja") {
+          if (enableTaskList !== undefined) cleanedService.widget.enableTaskList = enableTaskList;
```
shamoon, line 640:
```suggestion
          if (enableTaskList !== undefined) cleanedService.widget.enableTaskList = !!enableTaskList;
```

src/widgets/vikunja/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    new Date(task.dueDate).getTime() > new Date("0001-01-01T00:00:00Z").getTime() &&
+    new Date(task.dueDate).getTime() <= new Date(Date.now() + 604800000)
+  );
+}
+
+// return Tasks that have a real due date in the past
+function getOverdue(task) {
+  return (
+    new Date(task.dueDate).getTime() > new Date("0001-01-01T00:00:00Z").getTime() &&
```
shamoon, line 23: What exactly is the point of this line?

shamoon (changes requested):
Quick notes for now. I’ll find some time to test

src/widgets/vikunja/component.jsx
vhsdream: For tasks that you don't set a due date for, Vikunja creates one anyway ("0001-01-01T00:00:00Z") and that was causing tasks with no due date to appear in the responses. This was my attempt to filter them out. Admittedly I am very new to Javascript (like a week new) so I am probably doing things inefficiently.

src/widgets/vikunja/component.jsx
vhsdream: All right - after a quick crash course in anonymous functions I have made some changes, I hope they are good!

shamoon:
Thanks for the changes. Overall pretty good, just made some changes as you'll see. Let me know if you have any questions / concerns of course

vhsdream:
Hi, I like all the changes and optimizations, I'm learning a lot! However I had to make a few small adjustments in order for the widget to work: add `> 0` back to the projectsData filter function - otherwise it returns everything in Projects, including Saved Filters, which are not technically Projects. I guess a way I could describe them as not being Projects, is if I delete a Saved Filter, Tasks that were in them are not deleted, whereas if I delete a Project, all Tasks under that Project are deleted too.

Then I re-appended `.getTime()` to a few places where you cleaned up the date calculations, since without it, my widget was not showing any tasks due in 7 days or less.

Without `> 0` and `.getTime()`:
image

With `> 0` and `.getTime()`:
image

If the dates were being correctly compared in your environment, perhaps there are differences between our environments (I'm using a Debian 12 LXC, which is what I run homepage-prod in) and I have an issue of some kind, like missing dependencies?

shamoon:
Those changes are all unnecessary.

`(project) => project.id` will evaluate the same as `(project) => project.id > 0` and calling `getTime()` is not needed for date comparison.

But if you test it out and notice anything let me know.

shamoon:
Oh sorry I didn't see your comment above, let me look

shamoon:
Oh lol it uses negative numbers for the saved filters project id! Ha

shamoon:
The problem was the conversion in widget.js was still getting turned back into a string, take a look now.

Screenshot 2024-10-12 at 7 20 01 AM

shamoon:
Actually we can do it like this (last commit), the default due date is always the same string and just add a flag for that to the tasks.

I also made it show tasks without a due date in the list, totally happy to remove that if you prefer, but kinda figure why not?

Screenshot 2024-10-12 at 7 22 27 AM

shamoon:
And just to close the loop, date comparison in JS doesnt need `getTime` (except `==` `!==`, see https://stackoverflow.com/a/493018

=)

Thanks for the widget, welcome to JS land

vhsdream (commented):
This is awesome!

vhsdream (commented):
👍👍👍

shamoon (approved):

PR #4199, opened 2024-10-26
Chore: filter Radarr movie response

markp-mckinney:
## Proposed change

The current usage of `/api/v3/movie`'s `all` mapping is only for mapping queue IDs to movie titles. Originally I planned to move to using the queue details title, but that is the full download title and not strictly just the movie title. To not change the UI, we can at least remove unnecessary attributes from the response which is currently very bloated. If it's fine to change from movie title to download title, then `all` can be removed completely.

A user reported for a library of ~10k movies the response was >4MB:
```
[2024-10-26T07:29:29.928Z] warn: API response for /api/services/proxy?group=Movies+and+Series+Management&service=Radarr&endpoint=movie exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit
```

This change won't fix the overall issue of the size of the radarr response, but will at least filter down to what the frontend needs.

The Sonarr API already does filtering like this. In my testing this reduced my response size with a library of 308 movies from 1.1 MB to 12.86 kB.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If applicable, I have reviewed the [feature](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).

### DIFF
--- a/src/widgets/radarr/widget.js
+++ b/src/widgets/radarr/widget.js
@@ -12,7 +12,10 @@ const widget = {
         wanted: jsonArrayFilter(data, (item) => item.monitored && !item.hasFile && item.isAvailable).length,
         have: jsonArrayFilter(data, (item) => item.hasFile).length,
         missing: jsonArrayFilter(data, (item) => item.monitored && !item.hasFile).length,
-        all: asJson(data),
+        all: asJson(data).map((entry) => ({
+          title: entry.title,
+          id: entry.id
+        })),
       }),
     },
     "queue/status": {
--- a/src/widgets/radarr/widget.js
+++ b/src/widgets/radarr/widget.js
@@ -14,7 +14,7 @@ const widget = {
         missing: jsonArrayFilter(data, (item) => item.monitored && !item.hasFile).length,
         all: asJson(data).map((entry) => ({
           title: entry.title,
-          id: entry.id
+          id: entry.id,
         })),
       }),
     },


shamoon:
Well, this is not reducing the payload from sonarr, just from the homepage 'api'. If you're saying Radarr can slim down the response, let's do that. Otherwise this is just silencing a warning and I doubt will make any actual difference anywhere since this API is by definition always being served to the homepage app

markp-mckinney:
Yes, my wording was a little unclear but this is just cleaning up the homepage response since it doesn't need all the data from radarr. I don't see any way to reduce the payload from radarr since the radarr UI is using all the data by design. Either way this is bringing the sonarr/radarr homepage responses closer to each other and there's no need to pass all the data along to homepage if it's not being used. https://github.com/gethomepage/homepage/blob/dev/src/widgets/sonarr/widget.js#L12

Another small benefit is for instances like mine it reduces the amount of WAN traffic since my homepage -> radarr requests are local. In an ideal world radarr would expose a library metadata api instead of requiring parsing the entire library response.

shamoon (approved):
For the future, especially for such a small PR, please do follow the contributing guidelines not just tick the box

PR #4327, opened 2024-11-25
Enhancement: resources network widget

ramazansancar:
upd: Docs updated for network usage.
add: Widget resorce iconChildren param added.
upd: config file default options added.
upd: resources units type added.

## Proposed change

Closes #3421 (Discussion)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [X] If applicable, I have reviewed the [feature](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).

### DIFF
--- a/docs/widgets/info/resources.md
+++ b/docs/widgets/info/resources.md
@@ -20,7 +20,6 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     cpu: true
     memory: true
     disk: /disk/mount/path
-    network: default # options: 'default' or active network interface name
     cputemp: true
     tempmin: 0 # optional, minimum cpu temp
     tempmax: 100 # optional, maximum cpu temp
@@ -28,6 +27,7 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     units: imperial # only used by cpu temp, options: 'imperial' or 'metric'
     refresh: 3000 # optional, in ms
     diskUnits: bytes # optional, bytes (default) or bbytes. Only applies to disk
+    network: true # uses 'default' or specify a network interface name
 ```
 
 You can also pass a `label` option, which allows you to group resources under named sections,
--- a/docs/widgets/info/resources.md
+++ b/docs/widgets/info/resources.md
@@ -27,7 +27,7 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     units: imperial # only used by cpu temp, options: 'imperial' or 'metric'
     refresh: 3000 # optional, in ms
     diskUnits: bytes # optional, bytes (default) or bbytes. Only applies to disk
-    network: true # uses 'default' or specify a network interface name
+    network: true # optional, uses 'default' if true or specify a network interface name
 ```
 
 You can also pass a `label` option, which allows you to group resources under named sections,
--- a/docs/widgets/info/resources.md
+++ b/docs/widgets/info/resources.md
@@ -20,11 +20,12 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     cpu: true
     memory: true
     disk: /disk/mount/path
+    network: default # options: 'default' or active network interface name
     cputemp: true
     tempmin: 0 # optional, minimum cpu temp
     tempmax: 100 # optional, maximum cpu temp
     uptime: true
-    units: imperial # only used by cpu temp
+    units: imperial # only used by cpu temp widget, options: 'imperial' or 'metric'
     refresh: 3000 # optional, in ms
     diskUnits: bytes # optional, bytes (default) or bbytes. Only applies to disk
 ```
--- a/docs/widgets/info/resources.md
+++ b/docs/widgets/info/resources.md
@@ -25,7 +25,7 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     tempmin: 0 # optional, minimum cpu temp
     tempmax: 100 # optional, maximum cpu temp
     uptime: true
-    units: imperial # only used by cpu temp widget, options: 'imperial' or 'metric'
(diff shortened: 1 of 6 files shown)


docs/widgets/info/resources.md
```diff
@@ -20,11 +20,12 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
     cpu: true
     memory: true
     disk: /disk/mount/path
+    network: default # options: 'default' or active network interface name
     cputemp: true
     tempmin: 0 # optional, minimum cpu temp
     tempmax: 100 # optional, maximum cpu temp
     uptime: true
-    units: imperial # only used by cpu temp
+    units: imperial # only used by cpu temp widget, options: 'imperial' or 'metric'
```
shamoon, line 28:
```suggestion
    units: imperial # only used by cpu temp, options: 'imperial' or 'metric'
```

src/components/widgets/resources/network.jsx
```diff
@@ -0,0 +1,62 @@
...
+import { FaNetworkWired, FaAngleUp, FaAngleDown } from "react-icons/fa";
+import { useTranslation } from "next-i18next";
+
+import Resource from "../widget/resource";
+import Error from "../widget/error";
+
+export default function Network({ options, refresh = 1500 }) {
+  const { t } = useTranslation();
+
+  const { data, error } = useSWR(`/api/widgets/resources?type=network${(options.network || ( options.network !== 'default' && options.network === `false`)) ? `&interfaceName=${options.network}` : '' }`, {
```
shamoon, line 11: Did you really mean to test for network === “false” the string here?

src/pages/api/widgets/resources.js
```diff
@@ -8,6 +8,7 @@ const si = require("systeminformation");
 
 export default async function handler(req, res) {
   const { type, target } = req.query;
+  let { interfaceName } = req.query;
```
shamoon, line 11: This presumably should be a const like the others

shamoon (changes requested):
Thanks. This does need some work. I noted some small things on my phone but I need to take a closer look. It also won’t pass lint at the moment.

src/components/widgets/resources/network.jsx
ramazansancar: I wanted to check the case where the value is false and not equal to "default". @shamoon

src/pages/api/widgets/resources.js
ramazansancar: If we define it as constant const, I cannot change its value and control it in case it comes as default.

https://github.com/gethomepage/homepage/pull/4327/files#diff-0b3b2a1073c6f2cb86bf9a749bc02609827974f397d6847e3f658bcdf551300bR71-R72

ramazansancar:
> Thanks. This does need some work. I noted some small things on my phone but I need to take a closer look. It also won’t pass lint at the moment.

I will look at the lint issue again. I have provided answers for other change requests.

As an additional feature, I have in mind to separate live network usage and total network usage, i.e. define it separately from the config, what is your opinion on this? (Those who want only one can open it from the config as they wish. Those who want both can add it by writing 2 lines in the config.)

src/components/widgets/resources/network.jsx
shamoon: That's not what this is doing anyway.

src/pages/api/widgets/resources.js
shamoon: You can initialize it to a default

shamoon (changes requested):
You can take a look my changes, this is cleaner already.

Screenshot 2024-11-24 at 8 30 18 PM

But I'm still not happy with the aesthetics.

1. This is the only widget item with two lines by default, which is a problem.
2. Yes, we are not going to show both of them. I'd suggest we make the throughput and totals each their own line and the widget defaults to throughput and the totals is in the 'expanded' version.

ramazansancar:
> That's not going to work, and will make it overly-complicated. Take a look a the recent changes.
> 
> Screenshot 2024-11-24 at 8 41 05 PM
> Screenshot 2024-11-24 at 8 40 22 PM
> 
> The final issue here is the percentage bar, which you've just set... to 0?


The most logical thing to do at the moment seems to be to keep it at 0. We can't do anything because we won't have any max data.

shamoon (approved):
Screenshot 2024-11-24 at 10 30 23 PM

o2e:
I would like to add a tip, if you want to get the host's NIC information, then the docker container may need to be modified to `network_mode: host`. If you want to modify the port in host network mode, you also need to add `environment: - PORT=3000`(3000 is your port)

o2e:
> I would like to add a tip, if you want to get the host's NIC information, then the docker container may need to be modified to `network_mode: host`. If you want to modify the port in host network mode, you also need to add `environment: - PORT=3000`(3000 is your port)

I think it would be a good idea to add this note to the documentation, otherwise some newbies might not use it, and at least some of my friends don't know the difference between docker's network modes

shamoon:
The problem with doing that is it will almost certainly break existing integrations with other containers. It was always my beef with this widget but then again the same can apply to other resources in some situations

o2e:
Yes, it's a matter of trade-offs. Despite his slight flaws, it's worth it to me because I don't have any other easy and intuitive way to see what's going on with the network. `homepage` serves as the homepage of my browser, the monitor window of my portable screen, and having a small, sophisticated display of network information is great for me, and I can live with a few tweaks for that (I know that Glances has a similar feature, but I don't really like it! )
image

ramazansancar:
I forgot to thank you. Thanks for supporting the development and pushing it forward at some point. It was a nice feature to the project. @shamoon

PR #4520, opened 2024-12-30
Add Hoarder service widget

gatesry:
Add a new service widget for Hoarder to the homepage.


gatesry:
Disregard, still testing

PR #4564, opened 2025-01-05
Chore: migrate dashboard-icons to homarr-labs

walkxcode:
Info: https://github.com/homarr-labs/dashboard-icons/discussions/797#discussioncomment-11736264.

_Note: The walkxcode/dashboard-icons repository will continue to function indefinitely but may no longer receive updates._

### DIFF
--- a/docs/configs/services.md
+++ b/docs/configs/services.md
@@ -134,7 +134,7 @@ Services may have descriptions,
 
 ## Icons
 
-Services may have an icon attached to them, you can use icons from [Dashboard Icons](https://github.com/walkxcode/dashboard-icons) automatically, by passing the name of the icon, with, or without `.png` or with `.svg` to use the svg version.
+Services may have an icon attached to them, you can use icons from [Dashboard Icons](https://github.com/homarr-labs/dashboard-icons) automatically, by passing the name of the icon, with, or without `.png` or with `.svg` to use the svg version.
 
 You can also specify prefixed icons from:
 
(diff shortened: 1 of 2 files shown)


shamoon (changes requested):
walkxcode himself, cool.

Well thanks for taking the time to PR, thanks for your project.

I think this contains an error at the moment, it's re-declaring the const and the (unintentionally?) added webp block is unreachable

walkxcode:
> walkxcode himself, cool.
> 
> Well thanks for taking the time to PR, thanks for your project.
> 
> I think this contains an error at the moment, it's re-declaring the const and the (unintentionally?) added webp block is unreachable

The constant uses the same logic as the SVG and PNG retrieval. Folder links to jsDelivr show as unreachable because we’ve exceeded the ‘package’ limit. However, direct file links, such as https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/webp/1password.webp, will still work.

walkxcode:
Ah seems like we're missing the `if (icon.endsWith(".webp")) {`!

walkxcode:
Ideally the PNGs should be removed completely as they have the same function as the WebPs but with a bigger size, but that's all up to you ofcourse. Let me know if you want the WebPs or PNGs :)

shamoon:
Thanks. Yea we can make the fallback to webp, its support is pretty darn good at this point

shamoon:
Actually I take that back. That would be breaking or at least kinda confusing. You can add webp if you like but please leave the png for now

shamoon (approved):
Thanks again

PR #4643, opened 2025-01-23
Enhancement: support for Kubernetes gateway API

djeinstine:
## Proposed change

Feature Request #4021

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:


src/utils/config/kubernetes.js
```diff
@@ -6,26 +6,50 @@ import { KubeConfig } from "@kubernetes/client-node";
...
+  let route = "ingress";
+  if (config?.route === "gateway") {
+    route = "gateway";
+  }
+
+  // traefik
+  let traefik = true;
+  if (config?.traefik === "disable") {
+    traefik = false;
+  }
```
shamoon, line 24:
```suggestion
  const route = config?.route ?? "ingress";
  const traefik = config?.traefik !== "disable";
```

or even `const route = config?.route === "gateway" ? "gateway" : "ingress";` if you want to be explicit

src/utils/config/kubernetes.js
```diff
@@ -6,26 +6,50 @@ import { KubeConfig } from "@kubernetes/client-node";
...
   checkAndCopyConfig("kubernetes.yaml");
 
   const configFile = path.join(CONF_DIR, "kubernetes.yaml");
   const rawConfigData = readFileSync(configFile, "utf8");
   const configData = substituteEnvironmentVars(rawConfigData);
   const config = yaml.load(configData);
-  const kc = new KubeConfig();
+  let kubeData;
 
   switch (config?.mode) {
```
shamoon, line 42:
```
  if (config?.mode === "disabled") {
    kubeData = { config: null };
  } else {
    kubeData = extractKubeData(config);
  }
```

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+const kc = kubeArguments.config;
+
+const apiGroup = "gateway.networking.k8s.io";
+const version = "v1";
+
+let crd;
+let core;
+let networking;
+let routingType;
+let traefik;
```
shamoon, line 18: Please dont use global variables like this, they should be passed to the methods as needed

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+        return httpRoute;
+      }),
+    );
+
+    httpRouteList = httpRouteListUnfiltered.filter((httpRoute) => httpRoute !== undefined);
+  }
+  return httpRouteList;
+}
+
+async function getIngressList(ANNOTATION_BASE) {
```
shamoon, line 108: Do not write method parameters like this

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+        (ingress) => ingress.metadata.annotations && ingress.metadata.annotations[`${ANNOTATION_BASE}/href`],
+      );
+      ingressList.items.push(...traefikServices);
+    }
+  }
+
+  return ingressList.items;
+}
+
+export async function getRouteList(ANNOTATION_BASE) {
```
shamoon, line 169:
```suggestion
export async function getRouteList(ANNOTATION_BASE) {
```

Same as above, do not write method parameters like this

shamoon (changes requested):
This needs some work I think, will mark as draft while we're awaiting input from dudo

src/utils/config/service-helpers.js
```diff
@@ -202,128 +174,70 @@ export async function servicesFromKubernetes() {
...
-      .then((response) => response.body)
-      .catch((error) => {
-        logger.error("Error getting ingresses: %d %s %s", error.statusCode, error.body, error.response);
-        logger.debug(error);
-        return null;
-      });
-
-    const traefikContainoExists = await checkCRD(kc, "ingressroutes.traefik.containo.us");
-    const traefikExists = await checkCRD(kc, "ingressroutes.traefik.io");
+    const routeList = await getRouteList(ANNOTATION_BASE);
```
dudo, line 177: I like the shuffle, here, but I think we can do better. I'm thinking something like this, where the logic for each resource is encapsulated within its own file. Then in each of the list_ functions, we can early out with the config check `config?.ingress`, `config?.gateway`, `config?.traefik`, respectively.

```js
export async function servicesFromKubernetes() {
  const { instanceName } = getSettings();

  checkAndCopyConfig("kubernetes.yaml");

  try {
    const kc = getKubeConfig();
    if (!kc) {
      return [];
    }

    const [ingressList, traefikIngressList, httpRouteList] = await Promise.all([
      kubernetes.listIngress(kc),
      kubernetes.listTraefikIngress(kc),
      kubernetes.listHttpRoute(kc)
    ]);

    const resources = [ ...ingressList, ...traefikIngressList, ...httpRouteList ];
    if (!resources) {
      return [];
    }

    const services = resources
      .filter(resource => kubernetes.isDiscoverable(resource, instanceName))
      .map(resource => kubernetes.constructedServiceFromResource(resource))
      // I think we lift ANNOTATION_BASE and ANNOTATION_WIDGET_BASE to config/kubernetes.js so they don't need to be passed here.
      // For href, I think we can duck type it. We could over engineer it, and instantiate each resource as an object/class, and call href(), but I think we can just check for spec.rules[0].host, then else if spec.rules[0].matches
    
    const mappedServiceGroups = services.reduce((groups, serverService) => {
      let serverGroup = groups.find(group => group.name === serverService.group);
    
      if (!serverGroup) {
        serverGroup = {
          name: serverService.group,
          services: []
        };
        groups.push(serverGroup);
      }
    
      const { name: serviceName, group: _, ...pushedService } = serverService;
    
      serverGroup.services.push({
        name: serviceName,
        ...pushedService
      });
    
      return groups;
    }, []);
    
    return mappedServiceGroups;
  } catch (e) {
    if (e) logger.error(e);
    throw e;
  }
}
```

dudo (changes requested):

docs/installation/k8s.md
```diff
@@ -216,6 +216,15 @@ rules:
     verbs:
       - get
       - list
+  # if using gateway api add the following:
```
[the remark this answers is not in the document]
dudo: We'll want to submit a PR in the helm chart for this as well.

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+      logger.error("Error getting gateways: %d %s %s", error.statusCode, error.body, error.response);
+      logger.debug(error);
+      return "";
+    });
+  return schema;
+};
+
+async function getUrlFromHttpRoute(ingress) {
+  const urlHost = ingress.spec.hostnames[0];
+  const urlPath = ingress.spec.rules[0].matches[0].path.value;
```
dudo, line 57: Should we guard against type here? This won't work if type is `RegularExpression`.

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+    .catch((error) => {
+      logger.error("Error getting gateways: %d %s %s", error.statusCode, error.body, error.response);
+      logger.debug(error);
+      return "";
+    });
+  return schema;
+};
+
+async function getUrlFromHttpRoute(ingress) {
+  const urlHost = ingress.spec.hostnames[0];
```
dudo, line 56: When using httproutes for [GAMMA](https://gateway-api.sigs.k8s.io/mesh/) (configuring the service mesh), there is no `hostnames` key. We'll want to guard against that as well.

docs/configs/kubernetes.md
```diff
@@ -8,6 +8,7 @@ The Kubernetes connectivity has the following requirements:
 - Kubernetes 1.19+
 - Metrics Service
 - An Ingress controller
+  - Optionally: Gateway-API
```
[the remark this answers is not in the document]
dudo: Oh, I see the text above - "requirements". An ingress controller is required regardless, so I'd pull out the gateway api line entirely.

src/utils/kubernetes/kubernetes-routes.js
dudo: [Example](https://github.com/gitops-ci-cd/greeting-service-deployment/blob/main/base/grpc-route.yaml). This is grpc, but is the same for http. These aren't publicly accessible, and only used to configure a service mesh, so we can exclude entirely.

.github/workflows/sync-dev-integration.yml
shamoon, line 1: Please remove this file

djeinstine: Sorry I messed up. Removed.

src/utils/kubernetes/kubernetes-routes.js
djeinstine: Ok done. I returned a null url if the spec has no hostname key.

src/utils/kubernetes/kubernetes-routes.js
djeinstine: You're right. It only works with a rule match type of `PathPrefix`. So I only return a url when the type is `PathPrefix`

src/utils/kubernetes/kubernetes-routes.js
```diff
@@ -0,0 +1,211 @@
...
+const apiGroup = "gateway.networking.k8s.io";
+const version = "v1";
+
+let crd;
+let core;
+let networking;
+let routingType;
+let traefik;
+
+export async function checkCRD(name) {
```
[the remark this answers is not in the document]
djeinstine: Done.

src/utils/kubernetes/kubernetes-routes.js
dudo: Or Exact. https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.PathMatchType

src/utils/config/kubernetes.js
```diff
@@ -6,26 +6,50 @@ import { KubeConfig } from "@kubernetes/client-node";
 
 import checkAndCopyConfig, { CONF_DIR, substituteEnvironmentVars } from "utils/config/config";
 
-export default function getKubeConfig() {
+const extractKubeData = (config) => {
```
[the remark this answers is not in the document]
djeinstine: I reverted the old behaviour. I think that should be fine.

docs/configs/kubernetes.md
```diff
@@ -19,6 +20,12 @@ The Kubernetes connection is configured in the `kubernetes.yaml` file. There are
 mode: default
 ```
 
+To enable Kubernetes gateway-api compatibility, add the following setting:
+
+```yaml
+route: gateway
```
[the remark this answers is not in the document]
djeinstine: I agree. My only concern with that approach is what if a user sets everything true by accident? What kind of behaviour would be expected then?

I guess we could just default to ingress in that case?

```
ingress: true
traefik: true
gateway: true
```

docs/installation/k8s.md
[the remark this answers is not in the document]
djeinstine: Done. Once this PR is accepted I'll make the PR to the helm chart.

src/utils/config/service-helpers.js
djeinstine: So I just started the reshuffle as you described and the only thing I see is unfortunately ANNOTATION_BASE is needed for the traefikList filter :(

```(ingress) => ingress.metadata.annotations && ingress.metadata.annotations[`${annotationBase}/href`],```

Comment (djeinstine) [reply src/utils/config/service-helpers.js:177]: Please check the latest files I've uploaded. I haven't fully integrated them into service-helpers.js yet but I've created the functions you outlined above. Let me know what you think.

```
listIngress
listTraefikIngress
listHttpRoute
isDiscoverable
constructedServiceFromResource
```

File: docs/configs/kubernetes.md (hunk 13 above)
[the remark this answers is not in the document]
Comment (dudo) [reply docs/configs/kubernetes.md:26]: My understanding is that they all can exist simultaneously, no?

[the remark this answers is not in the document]
Comment (Bricklou) [reply docs/configs/kubernetes.md:26]: > My understanding is that they all can exist simultaneously, no?

Traefik is able to understand all the resources and make them available. So yes, everything can exist at the same time. Ideally, the behaviour would be to scan all the resources detected and merge them on the page based on the configuration.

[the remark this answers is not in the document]
Comment (djeinstine) [reply docs/configs/kubernetes.md:26]: Understood.

File: src/utils/kubernetes/kubernetes-traefik-list.js [hunk 14]
```diff
@@ -0,0 +1,60 @@
...
+            return [];
+        });
+
+        const traefikIngressList = [...(traefikIngressListContaino?.items ?? []), ...(traefikIngressListIo?.items ?? [])];
+
+        if (traefikIngressList.length > 0) {
+            const traefikServices = traefikIngressList.filter(
+                (ingress) => ingress.metadata.annotations && ingress.metadata.annotations[`${annotationBase}/href`],
+            );
+            traefikList.items.push(...traefikServices);
```
dudo, line 55: `.items` doesn't exist on the array we're building up - that's on the original objects retrieved from the cluster.

```suggestion
            traefikList.push(...traefikServices);
```

src/utils/config/service-helpers.js
dudo: Right, so for those constants, set them within `utils/config/kubernetes.js`

```js
export const ANNOTATION_BASE = "gethomepage.dev";
export const ANNOTATION_WIDGET_BASE = `${ANNOTATION_BASE}/widget.`;
```

Then just import them in the files you need them. No need to pass them around as arguments, since they're, well, constant.

```js
import { ANNOTATION_BASE, ANNOTATION_WIDGET_BASE } from "utils/config/kubernetes";
```

src/utils/kubernetes/kubernetes-routes.js
dudo, line 1: This file can be deleted, yes?

src/utils/kubernetes/kubernetes-traefik-list.js
```diff
@@ -0,0 +1,60 @@
+import CustomObjectsApi from "@kubernetes/client-node";
+import createLogger from "utils/logger";
+
+export async function listIngress(annotationBase,kc) {
```
dudo, line 4:
```suggestion
export async function listTraefikIngress(kc) {
```

src/utils/kubernetes/kubernetes-traefik-list.js
dudo, line 1: Nit, but on these filenames, I'd drop the `kubernetes-` prefix, personally.

src/utils/kubernetes/kubernetes-service-helpers.js
```diff
@@ -0,0 +1,64 @@
+import getUrlSchema from "utils/kubernetes/kubernetes-routes";
```
dudo, line 1:
```suggestion
import getUrlSchema from "utils/kubernetes/kubernetes-routes";
import * as shvl from "utils/config/shvl";
import { ANNOTATION_BASE, ANNOTATION_WIDGET_BASE } from "utils/config/kubernetes";
```

src/utils/kubernetes/kubernetes-httproute-list.js
```diff
@@ -0,0 +1,50 @@
+
+import { CustomObjectsApi, CoreV1Api } from "@kubernetes/client-node";
+import createLogger from "utils/logger";
+
+
+
+export async function getHttpRouteList(kc) {
+  
+    const logger = createLogger("service-helpers");
```
dudo, line 9: You'll want to create this logger outside of the function, and name it accordingly (based on the filename seems to be the pattern)

src/utils/kubernetes/kubernetes-service-helpers.js
dudo, line 1: naming things is hard... I'd call this `resource-helpers`

src/utils/kubernetes/kubernetes-traefik-list.js
```diff
@@ -0,0 +1,60 @@
...
+    
+    const logger = createLogger("service-helpers");
```
dudo, line 6: You'll want to create this logger outside of the function, and name it accordingly (based on the filename seems to be the pattern)

src/utils/kubernetes/kubernetes-ingress-list.js
```diff
@@ -0,0 +1,18 @@
+import NetworkingV1Api from "@kubernetes/client-node";
```
dudo, line 1:
```suggestion
import NetworkingV1Api from "@kubernetes/client-node";
import createLogger from "utils/logger";

const logger = createLogger("ingress-list");
```

dudo (commented):
This is taking shape! We're getting to a point where you've shuffled the code around that was interleaved together into their own files, then added a new file to capture the new logic.

I found a handful of bugs, and missing imports. Thank you for the iterations.

src/utils/config/kubernetes.js
```diff
@@ -27,5 +22,54 @@ export default function getKubeConfig() {
...
+  const traefik = config?.traefik !== "disable";
+
+  return {
+    config: kc,
+    route,
+    traefik,
+  };
+};
+
+export default function getKubeArguments() {
```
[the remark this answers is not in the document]
djeinstine: Done.

djeinstine:
Hi @dudo. I believe I've updated everything to reflect what was discussed. I've tested both HttpRoutes and Ingresses and everything works as intended. It would be nice if someone with the Traefik CRDs installed could test that. Let me know what you think!

src/utils/kubernetes/httproute-list.js
```diff
@@ -0,0 +1,54 @@
...
+const kc = getKubeConfig();
+
+export default async function listHttpRoute() {
+  
+    const crd = kc.makeApiClient(CustomObjectsApi);
+    const core = kc.makeApiClient(CoreV1Api);
+    const { gateway } = getKubernetes();
+    let httpRouteList = [];
+
+    if (gateway === true) {
```
dudo, line 17:
```suggestion
    if (!!gateway) {
```

src/utils/kubernetes/ingress-list.js
```diff
@@ -0,0 +1,29 @@
...
+const kc = getKubeConfig();
+
+
+export default async function listIngress() {
+
+    const networking = kc.makeApiClient(NetworkingV1Api);
+    const { ingress } = getKubernetes();
+    let ingressList = []
+    
+    if (ingress===true){
```
dudo, line 16:
```suggestion
    if (!!ingress){
```

dudo (commented):
Lots of inconsistent spacing. Lint those bad boys, and let's get this tested! 

Good work, though. This looks pretty good!

djeinstine:
> Lots of inconsistent spacing. Lint those bad boys, and let's get this tested!
> 
> Good work, though. This looks pretty good!

Hi. So I merged your commits and they don't pass the built-in linter. Also where is the inconsistent spacing? They also do not show up on the pre-build linter.

image

shamoon:
> Hi. So I merged your commits and they don't pass the built-in linter. Also where is the inconsistent spacing? They also do not show up on the pre-build linter.

Yes they do, presumably you haven't installed pre-commit.

I ran the code style linter and fixed the double-negations, I dont see why they're necessary anyway

I've also merged the latest dev branch.

Am I to understand the PR is ready?

dudo (approved):
Other than some linting nits, this looks good, and should alleviate some of the pain from working in this particular section of code.

dudo:
@shamoon just checked your commit, we don't want !. !! or none at all.

src/utils/config/kubernetes.js
```diff
@@ -2,18 +2,21 @@ import path from "path";
 import { readFileSync } from "fs";
 
 import yaml from "js-yaml";
-import { KubeConfig } from "@kubernetes/client-node";
+import { KubeConfig, ApiextensionsV1Api } from "@kubernetes/client-node";
 
 import checkAndCopyConfig, { CONF_DIR, substituteEnvironmentVars } from "utils/config/config";
 
-export default function getKubeConfig() {
+export default function getKubernetes() {
```
dudo, line 9: I would lose the default here, too, even. A default method is most useful if it's the _only_ method, but once you mix it in with named methods, it's a bit janky.

djeinstine:
Looks like my version wasn't updated with the latest dev packages. Apparently there was a new ObjectsApi for the @kubernetes/client-node module. I updated the code to reflect the changes.

dudo:
Nice catch!

What's integration testing look like for this @shamoon? Just build the image locally, or do you have any CI in place for that?

shamoon:
CI will build the container, and again lint is failing. I can fix later if you cant. The updated package was to patch a security vulnerability and yes when I merged back in dev it got updated here

shamoon:
Ps. The PR template does explain how to install e.g. https://gethomepage.dev/more/development/#code-formatting-with-pre-commit-hooks

shamoon (approved):
There has been so much back-and-forth on the code here I won't quibble with it.

Thanks to you both for all your work. I will refer all k8s complaints to you guys going forward 😜

PR #4673, opened 2025-01-28
[Bug fix] Fixed a wrong translation

Sergiy3013:
The previous translation used the word "gravity," which was incorrect and unrelated to "Domains on Adlists."  
This PR corrects the translation to reflect the intended meaning.


shamoon:
Thank you, all translation is handled via crowdin

Sergiy3013:
Can you share the link please?

PR #4682, opened 2025-02-01
Add initial back background for dark system theme

v1ack:
## Proposed change

As for now, if dark theme selected, we can see while flash on page load.

This is an example of page reloading

https://github.com/user-attachments/assets/96a12ff6-3370-4787-99bb-0f27ae8dd9bd

My suggestion is to add default dark background if user set system dark theme.

https://github.com/user-attachments/assets/1920cbd8-74a4-4053-8071-98d0ca4c4af0

- it won't affect users with light system and light homepage
- it won't affect users with light system and dark homepage, there will still be white flash
- ✅ there won't be light flash for users with dark system and dark homepage
- ⚠️ it will affect users with dark system and light homepage - there will be dark flash

It's bad change only for last type of users, but I think it's acceptable

The way it works - adding dark style in <head> section of page, to be loaded in first

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/pages/_app.jsx
+++ b/src/pages/_app.jsx
@@ -26,6 +26,16 @@ function MyApp({ Component, pageProps }) {
           name="viewport"
           content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no"
         />
+        <style
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{
+            __html: `@media (prefers-color-scheme: dark) {
+              #page_wrapper {
+                background-color: rgb(30, 41, 59);
+              }
+            }`,
+          }}
+        />
       </Head>
       <ColorProvider>
         <ThemeProvider>


shamoon:
Thanks but this is a hack so I’m not going to merge it, it also has the negative effect on light mode users you mentioned, it’s just swapping one problem for another (which doesn’t happen to affect you).

I believe this is browser dependent because we do other things to prevent it. So maybe try a better browser? 🙂 

If you have an alternative (better) solution we’d be open to that.

thanks again for contributing

v1ack:
> So maybe try a better browser? 🙂

Flash appears on Safari and Ark (chromium)

> it’s just swapping one problem for another

I agree, but as my opinion, it's the lowest category of users - with dark system and light homepage

> If you have an alternative (better) solution we’d be open to that.

Suddenly I found solution - add my code to `custom.css`. It has `preload` property, so it works as I expected

PR #4988, opened 2025-03-16
Change: re-enable k8s ingress by default

shamoon:
## Proposed change

@djeinstine @dudo sorry to keep bothering you guys, but now that I can run a little k3d cluster I was able to look at this, I was considering making `ingress: true` the default (aka not make 1.0 changes breaking). I believe this change will do it but the question is whether this will have other effects?

Thanks again for all your help

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [x] Other (please explain)

### DIFF
--- a/docs/configs/kubernetes.md
+++ b/docs/configs/kubernetes.md
@@ -25,13 +25,13 @@ To configure Kubernetes gateway-api, ingress or ingressRoute service discovery, 
 Example settings:
 
 ```yaml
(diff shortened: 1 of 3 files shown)


dudo:
Ingress and traefik being true would be original behavior is my understanding.

djeinstine:
Setting the default to those values won't cause any issues. 👍

shamoon:
Thanks guys

PR #5081, opened 2025-03-28
Fix: support shvl on customapi dynamic list target

matheusvellone:
## Proposed change

Support SHVL syntax on dynamic list target templating
There is no way to access nested fields on dynamic lists `target` field

```
  target: https://example.com/{subject.id}
```
Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] If applicable, I have reviewed the [feature / enhancement](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).
- [ ] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/more/development/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/more/development/#code-linting).

### DIFF
--- a/src/widgets/customapi/component.jsx
+++ b/src/widgets/customapi/component.jsx
@@ -246,10 +246,14 @@ export default function Component({ service }) {
                 const itemName = shvl.get(item, name, "");
                 const itemLabel = shvl.get(item, label, "");
 
-                const itemUrl = target ? [...target.matchAll(/\{(.*?)\}/g)].map((match) => match[1]).targetReplaces.reduce((url, targetTemplate) => {
-                  const value = shvl.get(item, targetTemplate, "");
-                  return url.replaceAll(`{${targetTemplate}}`, value);
-                }, target) : null;
+                const itemUrl = target
+                  ? [...target.matchAll(/\{(.*?)\}/g)]
+                      .map((match) => match[1])
+                      .reduce((url, targetTemplate) => {
+                        const value = shvl.get(item, targetTemplate, item[targetTemplate]) ?? "";
+                        return url.replaceAll(`{${targetTemplate}}`, value);
+                      }, target)
+                  : null;
                 const className =
                   "bg-theme-200/50 dark:bg-theme-900/20 rounded-sm m-1 flex-1 flex flex-row items-center justify-between p-1 text-xs";
 
--- a/src/widgets/customapi/component.jsx
+++ b/src/widgets/customapi/component.jsx
@@ -245,7 +245,11 @@ export default function Component({ service }) {
               listItems.map((item, index) => {
                 const itemName = shvl.get(item, name, "");
                 const itemLabel = shvl.get(item, label, "");
-                const itemUrl = target ? target.replace(/\{([^}]+)\}/g, (_, key) => item[key] || "") : null;
+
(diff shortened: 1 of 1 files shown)


src/widgets/customapi/component.jsx
```diff
@@ -245,7 +245,12 @@ export default function Component({ service }) {
               listItems.map((item, index) => {
                 const itemName = shvl.get(item, name, "");
                 const itemLabel = shvl.get(item, label, "");
-                const itemUrl = target ? target.replace(/\{([^}]+)\}/g, (_, key) => item[key] || "") : null;
+
+                const targetReplaces = [...target.matchAll(/\{(.*?)\}/g)].map((match) => match[1]);
```
matheusvellone, line 249: I used a different regex here.
I think this one is a bit simpler than the current one, its just a `.*` with the lazy modifier `?`

shamoon, line 249:
```suggestion
                const itemUrl = target ? 
                     [...target.matchAll(/\{(.*?)\}/g)].map((match) => match[1]).reduce((url, targetTemplate) => {
                         const value = shvl.get(item, targetTemplate, "");
                         return url.replaceAll(`{${targetTemplate}}`, value);
                     }, target) : null;
```

shamoon (commented):
Hmm, so I can see the value of this but Im a little concerned this adds complexity (maybe slows things down but lets be honest, probably not) where there doesnt need to be, for example the code as written will fail if target is null. Also this will fail if the property has a '.' in it, a rare but apparently a thing sometimes.

It's obviously not a big change, Im just trying to be careful about constantly changing / breaking things without a good reason (because we've done a fair bit of that lately)

src/widgets/customapi/component.jsx
matheusvellone: > Also this will fail if the property has a '.' in it, a rare but apparently a thing sometimes.

Yeah, indeed. But at least this would fail in all other shvl usages too
With a more complex regex this can be fixed, but i don't know if this is necessary for now.

Fixed with your suggestion ✅

yamek:
So you are working on how to build a link

src/widgets/customapi/component.jsx
```diff
@@ -245,7 +245,11 @@ export default function Component({ service }) {
               listItems.map((item, index) => {
                 const itemName = shvl.get(item, name, "");
                 const itemLabel = shvl.get(item, label, "");
-                const itemUrl = target ? target.replace(/\{([^}]+)\}/g, (_, key) => item[key] || "") : null;
+
+                const itemUrl = target ? [...target.matchAll(/\{(.*?)\}/g)].map((match) => match[1]).targetReplaces.reduce((url, targetTemplate) => {
+                  const value = shvl.get(item, targetTemplate, "");
```
shamoon, line 250:
```suggestion
                  const value = shvl.get(item, targetTemplate, item[targetTemplate]);
```

would fallback to getting the prop if it has a '.'

shamoon (approved):
Screenshot 2025-03-28 at 4 14 22 PM

This didn't actually work as-is, so I've fixed it as suggested.

Totally happy for you to contribute here, I would humbly suggest the best PRs are for high-value things that are requested 👍 

Thanks

PR #5199, opened 2025-04-20
Fix: kubernetes statistics not respecting selector

xethlyx:
## Proposed change

After upgrading to v1, I noticed that my Kubernetes widgets were reporting invalid statistics:

image

This clearly can't be correct, since `kubectl top` reports that it only uses about 1.5gb of memory:

```
NAME                                                     CPU(cores)   MEMORY(bytes)
...
prometheus-grafana-0                                     19m          384Mi
prometheus-kube-prometheus-operator-796b68b4c-qjswh      1m           35Mi
prometheus-kube-state-metrics-c44f8f948-r7mvd            4m           58Mi
prometheus-prometheus-kube-prometheus-prometheus-0       290m         1294Mi
prometheus-prometheus-node-exporter-bnzs7                1m           17Mi
prometheus-prometheus-node-exporter-gr6h6                1m           15Mi
prometheus-prometheus-node-exporter-wsss6                1m           27Mi
...
```

This issue can be reproduced with the following configuration:
```yaml
- My First Group:
    - My First Service:
        href: http://localhost/
        description: Homepage is awesome
        app: homepage
        namespace: services # any namespace with multiple pods
        podSelector: app=homepage # restrict this to a single pod
```

Instead of showing the statistics of the pods in `podSelector`, homepage will show the statistics of the entire namespace *multiplied* by the amount of pods found in the podSelector. This is because `.getPodMetrics(namespace, pod.items)` is returning the metrics for the entire namespace, as it does not accept a second argument.

This pull requests fixes the issue by filtering the metrics by the pods found in the namespace. Since `.getPodMetrics` can only filter by namespace, I also pull the fetch outside to reduce the number of calls from the amount of matched pods to just one. I considered using a simple filter per pod for a minimal diff, but I believe using a Set is the better choice here, in case there are a lot of selected pods, to avoid a `O(n^2)` time complexity.

Here is the same entry, with the fixes applied:

image

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/pages/api/kubernetes/stats/[...service].js
+++ b/src/pages/api/kubernetes/stats/[...service].js
@@ -53,9 +53,11 @@ export default async function handler(req, res) {
       return;
     }
 
+    const podNames = new Set();
     let cpuLimit = 0;
     let memLimit = 0;
     pods.forEach((pod) => {
+      podNames.add(pod.metadata.name);
       pod.spec.containers.forEach((container) => {
         if (container?.resources?.limits?.cpu) {
           cpuLimit += parseCpu(container?.resources?.limits?.cpu);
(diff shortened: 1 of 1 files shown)


shamoon (approved):
lgtm thanks

PR #5237, opened 2025-05-02
Feature: Satisfactory Server widget

InsertDisc:
## Proposed change

Closes #4010

## Type of change

- [X] New service widget
- [X] New feature or enhancement (non-breaking change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:


src/widgets/satisfactory/transforms.js
```diff
@@ -0,0 +1,98 @@
+export function formatDuration(seconds) {
```
shamoon, line 1: We have one of these already

src/widgets/satisfactory/transforms.js
shamoon, line 1: So we seriously need a file of hardcoded values for these?

shamoon (changes requested):
Oh, I finally looked what satisfactory is, it seems its a game?

So then you should use https://github.com/gamedig/node-gamedig/issues/442

src/widgets/satisfactory/transforms.js
InsertDisc: It's not completely necessary but the naming conventions the server api uses is basically useless for meaningful information. I.E. 'Schematic_2-1' isn't very 'human' friendly. The names for some of these things on the backend are never displayed that way in the frontend and as far as I could tell there is no human-friendly translations directly available in the api.

InsertDisc:
> Oh, I finally looked what satisfactory is, it seems its a game?
> 
> So then you should use [gamedig/node-gamedig#442](https://github.com/gamedig/node-gamedig/issues/442)

This wasn't a thing when I started the project, sadly. That's the entire reason I did it. So would we be able to use the built in gamedig widget?

src/widgets/satisfactory/transforms.js
InsertDisc: I could remove this but it was built to display the duration exactly as you would see on the server stats page.

InsertDisc:
> Oh, I finally looked what satisfactory is, it seems its a game?
> 
> So then you should use [gamedig/node-gamedig#442](https://github.com/gamedig/node-gamedig/issues/442)

Could you elaborate on this? Are you saying to use GameDig instead of Homepage or am I missing some implementation?

shamoon:
> This wasn't a thing when I started the project, sadly. That's the entire reason I did it. So would we be able to use the built in gamedig widget?

I would assume so based on the linked gamedig issue...

InsertDisc:
> > This wasn't a thing when I started the project, sadly. That's the entire reason I did it. So would we be able to use the built in gamedig widget?
> 
> I would assume so based on the linked gamedig issue...

The current implementation of the GameDig widget, after much trial and error, requires a password-less server. With a password set the most I could get to display was 'Online', the server name, and ping. Without a password the player count was available as well. I don't see/know of a way to pass a password or token with the GameDig widget. The node-gamedig core also doesn't appear to even provide much information more than those four anyway. I'm happy to go either way. It's nice to know at least we can get 'something' from the GameDig widget now, but having to leave the Player Password empty is disappointing. It's entirely possible to get the information using the Custom Api widget as well which is what I was doing. There was just that feature request I was trying to fulfill.

InsertDisc:
Also, I did try to add images to the PR showing things in action but I clearly did something wrong.

InsertDisc:
sds_widget_info_screenshot
sds_widget_screenshot

shamoon:
Thanks. I think the best course of action would be to PR the gamedig repository to add pw support, if thats possible, and something like that could make it back into homepage. But in general, we're not going to start adding support for multiple individual games like this. Sorry for not realizing earlier what this was.

InsertDisc:
> Thanks. I think the best course of action would be to PR the gamedig repository to add pw support, if thats possible, and something like that could make it back into homepage. But in general, we're not going to start adding support for multiple individual games like this. Sorry for not realizing earlier what this was.

To be clear, you CAN pass a token to GameDig to be authorized on the server, I just don't think there is currently a way to do that with the widget.

PR #5241, opened 2025-05-03
Chore: change to ical.js for ical parsing

matheusvellone:
## Proposed change
Swap ical lib parser to a maintained one, as discussed on #5077 
The new lib support more iCal features than the old one, so we can expect the widget to show more data

Closes #5077 

## Type of change
- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] (not applicable, i guess) If applicable, I have added corresponding documentation changes.


src/widgets/calendar/integrations/ical.jsx
```diff
@@ -25,85 +24,117 @@ export default function Integration({ config, params, setEvents, hideErrors, tim
...
+      const vCalendar = new ICAL.Component(jCal)
+
+      const buildEvent = (event, type) => {
+        return {
+          id: event.getFirstPropertyValue("uid"),
+          type,
+          title: event.getFirstPropertyValue("summary"),
+          rrule: event.getFirstPropertyValue("rrule"),
+          dtstart: event.getFirstPropertyValue("dtstart") || event.getFirstPropertyValue("due") || event.getFirstPropertyValue("completed") || ICAL.Time.now(),
+          dtend: event.getFirstPropertyValue("dtend") || event.getFirstPropertyValue("due") || event.getFirstPropertyValue("completed") || ICAL.Time.now(),
```
matheusvellone, line 48: The last `ICAL.Time.now()` handles the case where an event has already been completed in the past.

I feel like this entire date resolving code should be more explicit. It's a lot confusing on why this need to be like this. Would like your opinion on this

shamoon:
Screenshot 2025-05-04 at 8 24 02 AM

This seems to work great which is impressive considering how messy calendar stuff is. There may be issues Im not seeing but I'd be inclined to merge this into dev and see if people are willing to test more.

src/widgets/calendar/integrations/ical.jsx
shamoon: I dont understand the question here, "why this needs to be like this"? As for the code itself, yea it seems odd to just assign now to either the event start or end

matheusvellone: > it seems odd to just assign now to either the event start or end

Exactly. The logic behind the `ICAL.Time.now()` isn't explicit enough. It handles events from calendar that doen't have a date, like a note or something

shamoon: Im still a bit confused about your question. If we want to filter out events like that, perhaps just set the time to some minimum value that we know will always cause it to get ignored?

i.e. instead of `|| ICAL.Time.now(),` something like `|| 0,` (if that actually works, I dont know)

Or let me know if Im still not getting your q.

matheusvellone: `|| ICAL.Time.now()` handles events without a date. I've seen them on "open" reminders that doesn't have any kind of date, which makes sense.
In these cases, I expect them to show as "today", which I think is intuitive.

But the code isn't explict about this rule, which might be understood in the future and removed.

Probably just a comment to explain the case `|| ICAL.Time.now()` is covering should be enough

Got it?

shamoon: Sure yea

shamoon (approved):

PR #5255, opened 2025-05-09
feat: Allow working behind http proxy

amaczuga:
## Proposed change

Related discussion: #4012

Use npm global-agent@3.0.0 dependency, allowing homepage service to reach Internet URL's from behind squid-type http proxy. Configuration is based on environment variables, as described here: https://github.com/gajus/global-agent?tab=readme-ov-file#environment-variables

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "dockerode": "^4.0.4",
     "follow-redirects": "^1.15.9",
     "gamedig": "^5.2.0",
+    "global-agent": "^3.0.0",
     "i18next": "^24.2.3",
     "ical.js": "^2.1.0",
     "js-yaml": "^4.1.0",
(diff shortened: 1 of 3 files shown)


shamoon:
FYI, the lock file stuff is all malformed, but more importantly we are not going to merge this. 

global-agent monkey patches Node.js global HTTP and HTTPS behavior and we are not going to introduce a change that could potentially cause issues for a tiny subset of users.

You are of course welcome to maintain your own fork of the project.

Thanks for contributing

amaczuga:
@shamoon thank you for explaining the issue. From your point of view, what would be the right way to introduce environment variable http proxy support to the project? I don't expect tutoring, just a hint would suffice. Thanks in advance.

shamoon:
Homepage already constructs and manages its own HTTP(S) agents, and does not rely on Node.js global HTTP behavior. That makes introducing global-agent redundant and potentially dangerous, because:

- It globally monkey-patches http/https and would override existing agents
- We'd lose per-request proxy configurability
- It could lead to hard-to-debug behavior in environments that don’t expect proxying (e.g., CI, development)

You could probably do this cleanly with something like https://www.npmjs.com/package/https-proxy-agent (and http), and put it behind an ENV. Again, not like this.

PR #5274, opened 2025-05-15
Add ability to use Open ID SSO - such as keycloak

Sudo-JHare:
## Type of change - Security

## Checklist:

- [ ] If applicable, I have added corresponding documentation changes. - new readme file, kept original intact.

to:do  add ENV option to disable SSO so dash loads with or without SSO based on option.
.


shamoon:
We have said before we will not be including user management/auth inside homepage and that hasn’t changed. Starting with a discussion (or using an existing feature request discussion) as is required to open a PR by the way, would have revealed this and perhaps saved you the time. Thank you for the PR anyway, sorry we won’t be merging it

PR #5285, opened 2025-05-17
Enhancement: support specifying protocol for Docker Socket Proxy

grbkdd:
## Proposed change

Closes #4658

## Type of change

- [ ] New service widget
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or enhancement (non-breaking change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If applicable, I have added corresponding documentation changes.
- [x] If applicable, I have reviewed the [feature / enhancement](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).
- [x] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/more/development/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/more/development/#code-linting).
- [x] If applicable, I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.

PR #5304, opened 2025-05-22
Enhancement: Support more docker API connection options.

apollo13:
## Proposed change

Following #5285, this change adds even more options to customize the connection to the docker API. I'd argue it also makes it more secure if you are running with a reverse proxy because you can now add basic auth :)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] If applicable, I have reviewed the [feature / enhancement](https://gethomepage.dev/more/development/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/more/development/#service-widget-guidelines).

### DIFF
--- a/docs/configs/docker.md
+++ b/docs/configs/docker.md
@@ -79,6 +79,18 @@ my-docker:
 
     Note: This does not require TLS certificates if the proxy handles encryption. Do not use `protocol: https` unless you’re sure the target host supports HTTPS.
 
+You can further customize the connection options via `pathPrefix` and `headers`:
+
+```yaml
+my-docker:
+  host: dockerproxy
+  port: 443
+  protocol: https
+  pathPrefix: /docker_api/ # If a reverse proxy serves the API on a subpath
+  headers:
+    Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ== # Send an authorization header if the API access is secured via basic auth
+```
+
 ## Using Socket Directly
 
 If you'd rather use the socket directly, first make sure that you're passing the local socket into the Docker container.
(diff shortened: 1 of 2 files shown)


shamoon:
Thanks very much. Would like to make sure theres some interest on the feature request before merging, happy to readdress at that time.

apollo13:
@shamoon Not being familiar with the process here, what is the indicator for interest? The discussion has a few thumbs up already and the same arguments that applied to `protocol` before imo apply here as well (with the added benefit that with this change you can actually safely expose the socket, so a net security win).

shamoon:
I’m a bit surprised to see that many votes in a few hours since it’s literally never come up before tbh, but assuming those are actually people who want this (or did you post it somewhere?) that’s fine.

Mostly we’re just trying to avoid maintenance burden for changes that no one besides the author wants. I understand this is expected to be low maintenance since we’re just passing on the settings…

shamoon (approved):

apollo13:
> I’m a bit surprised to see that many votes in a few hours since it’s literally never come up before tbh, but assuming those are actually people who want this (or did you post it somewhere?) that’s fine.

Yes I did post it, but the people upvoting actually want this :D We are trying to use it as part of https://gitlab.com/rocketduck/paas to make a dashboard over multiple servers and TLS + basic auth is easier to handle than client TLS for small setups.

> Mostly we’re just trying to avoid maintenance burden for changes that no one besides the author wants. I understand this is expected to be low maintenance since we’re just passing on the settings…

That is a reason I can get behind, another cleanup might probably to pass all/most parameters on to the modem configuration, then the whole special casing could get removed.

shamoon:
Exactly, maintenance. I’ll remove it for now, thanks for the heads up

apollo13:
@shamoon If possible just the pathPrefix, the header work. Also host can be specified as https://… meaning protocol isn't required either

PR #5476, opened 2025-07-03
Enhancement: support for new grafana alerting api

Lappihuan:
## Proposed change

This PR leaves the existing legacy `alerts` API call unchanged and adds support for querying both Grafana's `grafana-alerting` data source and the `alertmanager` data source so that firing alerts from either source will be displayed.

I have opened a discussion here: https://github.com/gethomepage/homepage/discussions/5475

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/docs/widgets/services/grafana.md
+++ b/docs/widgets/services/grafana.md
@@ -5,11 +5,18 @@ description: Grafana Widget Configuration
 
 Learn more about [Grafana](https://github.com/grafana/grafana).
 
+| Grafana Version | Homepage Widget Version |
+| -------------- | ----------------------- |
+| <= v10.4       | 1 (default)             |
+| >= v10      | 2                       |
+
 Allowed fields: `["dashboards", "datasources", "totalalerts", "alertstriggered"]`.
 
 ```yaml
 widget:
   type: grafana
+  version: 2 # optional, default is 1
+  alerts: alertmanager # optional, default is grafana
   url: http://grafana.host.or.ip:port
   username: username
   password: password
--- a/docs/widgets/services/grafana.md
+++ b/docs/widgets/services/grafana.md
@@ -8,7 +8,7 @@ Learn more about [Grafana](https://github.com/grafana/grafana).
 | Grafana Version | Homepage Widget Version |
 | --------------- | ----------------------- |
 | <= v10.4        | 1 (default)             |
-| >= v10          | 2                       |
+| > v10.4         | 2                       |
 
 Allowed fields: `["dashboards", "datasources", "totalalerts", "alertstriggered"]`.
 
--- a/docs/widgets/services/grafana.md
+++ b/docs/widgets/services/grafana.md
@@ -6,9 +6,9 @@ description: Grafana Widget Configuration
 Learn more about [Grafana](https://github.com/grafana/grafana).
 
 | Grafana Version | Homepage Widget Version |
-| -------------- | ----------------------- |
-| <= v10.4       | 1 (default)             |
-| >= v10      | 2                       |
+| --------------- | ----------------------- |
+| <= v10.4        | 1 (default)             |
+| >= v10          | 2                       |
 
 Allowed fields: `["dashboards", "datasources", "totalalerts", "alertstriggered"]`.
 
--- a/src/widgets/grafana/component.jsx
+++ b/src/widgets/grafana/component.jsx
@@ -11,22 +11,26 @@ export default function Component({ service }) {
   const { data: statsData, error: statsError } = useWidgetAPI(widget, "stats");
   const { data: alertsData, error: alertsError } = useWidgetAPI(widget, "alerts");
   const { data: alertmanagerData, error: alertmanagerError } = useWidgetAPI(widget, "alertmanager");
+  const { data: grafanaData, error: grafanaError } = useWidgetAPI(widget, "grafana");
 
   let alertsInt = 0;
 
   if (alertsError || !alertsData || alertsData.length === 0) {
-    if (alertmanagerData) {
+    if (alertmanagerData.length > 0) {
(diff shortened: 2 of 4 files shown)


shamoon (changes requested):
Thanks. Rather than just adding it and making 4 API calls, how about incrementing the version of the widget and then we can just avoid the un-needed ones? Yes, it will require a config update for users but if these endpoints are truly deprecated the current behavior feels sorta 'lazy', if you know what I mean

Lappihuan:
Sure! Is there a widget that was versioned, to use as example?

shamoon:
There are many, if you search the docs. Immich for example but lots of others

shamoon (changes requested):
This is definitely a step in the right direction but at the moment you've effectively created 3 'versions'. What we want is to say the current behavior is homepage version 1, version 2 presumably, would be opting into this new behavior. And to make the changes non-breaking the default should be 1.

Lappihuan:
That situation is admittedly a bit confusing, so I will try to elaborate my understanding of it without using versions first, and then we can decide whether versioning is even the right abstraction here.

1. The legacy `/api/alerts` endpoint was removed in Grafana 10.4 [Legacy alerting removal…](https://grafana.com/blog/2024/04/04/legacy-alerting-removal-what-you-need-to-know-about-upgrading-to-grafana-alerting/).
2. The replacement is now under `/api/alertmanager/<implementation>/api/v2/alerts`, but there are two possible implementations:

   * `alertmanager` the prometheus-style implementation, widely used (e.g. in kube-prometheus-stack).
   * `grafana` grafanas "Unified Alerting" system [blog post](https://grafana.com/blog/2021/06/14/the-new-unified-alerting-system-for-grafana-everything-you-need-to-know/) but i have no idea how its adoption is.

Since we’re on Grafana v12 (and `/api/alerts` has been gone since v10.4), remove `/api/alerts` entirely and introduce an `alerts` parameter with allowed values `alertmanager` and `grafana`, with default `grafana` to be [as close to BC as possible](https://github.com/gethomepage/homepage/blob/1d990ef7c71fa94248d48a45a38c0374605aa731/src/widgets/grafana/widget.js#L12).

```yaml
- My Third Group:
    - My Third Service:
        href: http://localhost/
        description: Homepage is 😎

    - grafana:
        href: http://grafana.host.or.ip:port
        description: my grafana
        widget:
          type: grafana
          alerts: alertmanager # or grafana
          url: http://grafana.host.or.ip:port
          username: admin
          password: ******
```

But thats up to you, if you prefer this or the `version` parameter.

shamoon:
Good question. Annoyingly it might be both? So v1 = same, i.e. users running ≤ 10.4 homepage v2 = alertsmanager or grafana, default to whichever is more sane, allow choosing the other.

Again, we still need to support people who havent (aren't planning to) upgrade, so the rule is "dont break existing installs, support progressive enhancement".

LMK if that makes sense / seems OK?

shamoon:
> 19:56  Error: React Hook "useWidgetAPI" is called conditionally. React Hooks must be called in the exact same order in every component render.  react-hooks/rules-of-hooks

Yea this is why other widgets instead use conditionals to set _which_ endpoint to call (not wrap the entire call itself). If you need to not call an endpoint at all at all you can use null / empty string (I forget tbh)

Lappihuan:
`version 1` remains unchanged.
`version 2` removes the deprecated `/api/alerts` endpoint and introduces a new `alerts` parameter, which defaults to `grafana` (the endpoint currently in use).

Lappihuan:
oh i see, i will change that.

src/widgets/grafana/component.jsx
```diff
@@ -6,27 +6,59 @@ import useWidgetAPI from "utils/proxy/use-widget-api";
...
   const { t } = useTranslation();
-
   const { widget } = service;
+
+  const { version = 1, alerts = "grafana" } = widget;
+
+  let alertsError = null;
+  const allowedEndpoints = ["alertmanager", "grafana"];
+  if (!allowedEndpoints.includes(alerts)) {
+    alerts = ""; // Reset to empty string if invalid
```
shamoon: Not sure why lint misses it but this is reassigning a const

src/widgets/grafana/component.jsx
```diff
@@ -6,27 +6,59 @@ import useWidgetAPI from "utils/proxy/use-widget-api";
...
+    alertsError = new Error(
+      `Invalid alerts endpoint: ${alerts}, allowed endpoints are: ${allowedEndpoints.join(", ")}`,
```
shamoon: Error message not so helpful if you've reassigned `alerts` to `""` =)

docs/widgets/services/grafana.md
```diff
@@ -5,11 +5,18 @@ description: Grafana Widget Configuration
 
 Learn more about [Grafana](https://github.com/grafana/grafana).
 
+| Grafana Version | Homepage Widget Version |
+| --------------- | ----------------------- |
+| <= v10.4        | 1 (default)             |
+| >= v10          | 2                       |
```
shamoon:
```suggestion
| > v10.4         | 2                       |
```

shamoon (changes requested):
Just some minor stuff

src/widgets/grafana/component.jsx
Lappihuan: I wanted to do a early exit but the conditional hook call prevents that.
I think i can just get rid of the validation all together. If its a invalid endpoint it will throw a "API Error: Unsupported service endpoint".

shamoon (approved):
Screenshot 2025-07-12 at 8 27 45 AM

Great thanks

PR #5657, opened 2025-08-11
Enhancement: use docker dashboard portainer API instead of Docker HTTP API reverse proxy

rayanmestiri:
## Proposed change

This makes use of the `/docker/{environmentId}/dashboard` Portainer API : https://app.swaggerhub.com/apis/portainer/portainer-ce/2.27.9#/docker/dockerDashboard instead of the reverse-proxy to the Docker HTTP API `/api/endpoints/<environmentId>/docker` which was used before.  

It allows getting the data we need (containers running, stopped & total) without requiring the user to expose the Docker HTTP API.

These changes are transparent for the end user.

Screenshot 2025-08-11 at 12 20 54

Closes #5656

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/portainer/component.jsx
+++ b/src/widgets/portainer/component.jsx
@@ -70,11 +70,6 @@ export default function Component({ service }) {
     );
   }
 
-  if (dockerDashboard.error || dockerDashboard.message) {
-    // dockerDashboard can be itself an error object e.g. if environment fails
-    return <Container service={service} error={dockerDashboard?.error ?? dockerDashboard} />;
-  }
-
   return (
     <Container service={service}>
       <Block label="portainer.running" value={dockerDashboard.containers.running} />
--- a/src/widgets/portainer/component.jsx
+++ b/src/widgets/portainer/component.jsx
@@ -10,12 +10,9 @@ export default function Component({ service }) {
     widget.fields = widget.kubernetes ? ["applications", "services", "namespaces"] : ["running", "stopped", "total"];
   }
 
-  const { data: containersCount, error: containersError } = useWidgetAPI(
+  const { data: dockerDashboard, error: dockerError } = useWidgetAPI(
     widget,
-    widget.kubernetes ? "" : "docker/containers",
-    {
-      all: 1,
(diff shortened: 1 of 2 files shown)


shamoon:
1. The spirit of opening a feature request first is to gauge community interest, which of course we really can’t when you immediately open the PR.
2. I’m a little unclear on what the benefits are here while it’s hard to imagine there is no chance for this to introduce a regression. (In other words, I’m worried it will). Do you know how long the desktop api has been around and can you ensure it won’t drop users on older versions of portainer? Any other potential impacts? I guess in sum I’m a little worried the risks outweigh the benefits here.

given the above I’ll close this for now, we can certainly readdress in the future

lufiia:
To answer about what the benefits are. As I said in the description, the current implementation requires the user to have the Docker HTTP API exposed when it doesn't need to be. This isn't mentioned in the widget documentation. 

I found out precisely because the widget was not working, was not returning any useful error message so I had to dig in order understand what the issue was. This can happen to anyone which, like me, didn't expose the Docker HTTP API.

Besides, this means we're basically using the underlying Docker API which kind of defeats the purpose of this being a Portainer Widget imo.

As for retro-compatibility, the `/docker/{environmentId}/dashboard` has been introduced in version `2.22.0` of Portainer released in Oct 2, 2024 so I agree this can break old versions. If this feature were to be included, documentation should be updated in order to inform users of the requirements.

I needed the widget to work for myself so I figured it would be useful to share what I already worked on. But I understand this should be discussed in the feature request and I understand why this PR was closed. 

Thanks.

shamoon:
Yea makes sense, my concern would just being sure were sticking to the ‘greater good’ here. The portainer widget has been around a for a long time and we honestly get almost no complaints™ so I’m hesitant to change it. I suppose another approach would be to support both APIs, though it would make the logic overall more complicated instead of less, and again my question would be how many folks out there have a need for it.

Thanks for your understanding and interest in contributing

PR #5683, opened 2025-08-20
Feature: Unraid widget

dkaser:
## Proposed change

Adds a service widget for [Unraid](https://unraid.net/). The widget provides a large selection of fields to accommodate varying configurations -- e.g., some users might not have an array to monitor, other users might have multiple pools, etc.

One unusual feature about how the widget operates is related to the `fields` option. When displaying information about a pool, the label is dynamic -- it includes the pool name and the metric being displayed (e.g., "TESTPOOL USED (%)" in the example rendering). This doesn't work with the filtering that `fields` does since the label no longer matches the field.

To work around that issue, I switch the option provided to the component from `fields` to `display` in `cleanServiceGroups`. I did it this way so that the user configuration remains consistent with other widgets. If you would rather I make that a unique option instead, I can do that too.

### API

Unraid 7.2 (currently in beta) and later provide a GraphQL API. Instructions to create an API key are linked in the doc.

#### Example Request

```
query {
  array {
    state
    capacity {
      kilobytes {
        free
        total
        used
      }
    }
    caches {
      name
      fsType
      fsSize
      fsFree
      fsUsed
    }
  }
  metrics {
    memory {
      active
      available
      percentTotal
    }
    cpu {
      percentTotal
    }
  }
  notifications {
    overview {
      unread {
        total
      }
    }
  }
}
```

#### Example Response

```
{
  "data": {
    "array": {
      "state": "STARTED",
      "capacity": {
        "kilobytes": {
          "free": "54563033",
          "total": "80462479",
          "used": "25899446"
        }
      },
      "caches": [
        {
          "name": "testpool",
          "fsType": "zfs",
          "fsSize": 51506053,
          "fsFree": 41501696,
          "fsUsed": 10004357
        },
        {
          "name": "testpool2",
          "fsType": null,
          "fsSize": null,
          "fsFree": null,
          "fsUsed": null
        }
      ]
    },
    "metrics": {
      "memory": {
        "active": 3248586752,
        "available": 2974113792,
        "percentTotal": 52.205416748397525
      },
      "cpu": {
        "percentTotal": 5.897821187077385
      }
    },
    "notifications": {
      "overview": {
        "unread": {
          "total": 0
        }
      }
    }
  }
}
```

### Example Rendering

image

## Type of change

- [X] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

See #4514

## Checklist:

### DIFF
--- a/docs/widgets/services/unraid.md
+++ b/docs/widgets/services/unraid.md
@@ -1,0 +1,25 @@
+---
+title: Unraid
+description: Unraid Widget Configuration
+---
+
+Learn more about [Unraid](https://unraid.net/).
+
+The Unraid widget allows you to monitor the resources of an Unraid server.
+
+You will need an API key with the **GUEST** (read only) role: [Managing API Keys](https://docs.unraid.net/API/how-to-use-the-api/#managing-api-keys)
+
+The widget can display metrics for selected Unraid pools. If using one of the "pool" fields, you must also add the pool name to the settings.
+
+Allowed fields: `["cpu","memoryPercent","memoryAvailable","memoryUsed","notifications","arrayFreeSpace","arrayUsedSpace","arrayUsedPercent","status","pool1UsedSpace","pool1FreeSpace","pool1UsedPercent","pool2UsedSpace","pool2FreeSpace","pool2UsedPercent","pool3UsedSpace","pool3FreeSpace","pool3UsedPercent","pool4UsedSpace","pool4FreeSpace","pool4UsedPercent"]`
+
+```yaml
+widget:
+  type: unraid
+  url: https://unraid.host.or.ip
+  key: api-key
+  pool1: # optional, name of pool for pool1... fields
+  pool2: # optional, name of pool for pool2... fields
+  pool3: # optional, name of pool for pool3... fields
+  pool4: # optional, name of pool for pool4... fields
+```
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -1083,5 +1083,30 @@
         "nextMonthlyCost": "Next Month",
         "previousMonthlyCost": "Prev. Month",
         "nextRenewingSubscription": "Next Payment"
+    },
+    "unraid": {
+        "STARTED": "Started",
+        "STOPPED": "Stopped",
+        "NEW_ARRAY": "New Array",
+        "RECON_DISK": "Reconstructing Disk",
+        "DISABLE_DISK": "Disk Disabled",
+        "SWAP_DSBL": "Swap Disable",
+        "INVALID_EXPANSION": "Invalid Expansion",
+        "PARITY_NOT_BIGGEST": "Parity Not Biggest",
+        "TOO_MANY_MISSING_DISKS": "Too Many Missing Disks",
+        "NEW_DISK_TOO_SMALL": "New Disk Too Small",
+        "NO_DATA_DISKS": "No Data Disks",
+        "notifications": "Notifications",
+        "status": "Status",
+        "cpu": "CPU",
+        "memoryPercent": "Memory Used (%)",
+        "memoryUsed": "Memory Used",
+        "memoryAvailable": "Memory Available",
+        "arrayUsedSpace": "Array Used",
+        "arrayFreeSpace": "Array Free",
+        "arrayUsedPercent": "Array Used (%)",
+        "poolUsedSpace": "{{pool}} Used",
+        "poolFreeSpace": "{{pool}} Free",
+        "poolUsedPercent": "{{pool}} Used (%)"
     }
 }
--- a/src/widgets/unraid/component.jsx
+++ b/src/widgets/unraid/component.jsx
@@ -1,0 +1,53 @@
+import Block from "components/services/widget/block";
+import Container from "components/services/widget/container";
+import { useTranslation } from "next-i18next";
+
+import useWidgetAPI from "utils/proxy/use-widget-api";
+
+const UNRAID_DEFAULT_FIELDS = ["status", "cpu", "memoryPercent", "notifications"];
+const MAX_ALLOWED_FIELDS = 4;
+
+export default function Component({ service }) {
+  const { t } = useTranslation();
+  const { widget } = service;
+
+  const { data, error } = useWidgetAPI(widget);
+
+  if (error) {
+    return <Container service={service} error={error} />;
+  }
+
+  if (!data) {
+    if (!widget.display?.length > 0) {
+      widget.display = UNRAID_DEFAULT_FIELDS;
+    } else if (widget.display.length > MAX_ALLOWED_FIELDS) {
+      widget.display = widget.display.slice(0, MAX_ALLOWED_FIELDS);
+    }
+
+    return (
+      <Container service={service}>
+        {widget.display.map((field) => {
+          const poolMatch = field.match(/^(pool\d)(.+)/);
+          if (poolMatch) {
+            const poolName = widget?.[poolMatch[1]] || poolMatch[1];
+            const fieldName = "pool" + poolMatch[2];
+            return <Block key={field} label={t(`unraid.${fieldName}`, { pool: poolName })} />;
+          }
+          return <Block key={field} label={`unraid.${field}`} />;
+        })}
+      </Container>
+    );
+  }
+
+  return (
+    <Container service={service}>
+      {data.map((data) => (
+        <Block
+          key={data.key ?? data.label}
+          label={data.poolName ? t(data.label, { pool: data.poolName }) : data.label}
+          value={t(data.t, { value: data.value })}
+        />
+      ))}
+    </Container>
+  );
+}
--- a/src/widgets/unraid/proxy.js
+++ b/src/widgets/unraid/proxy.js
@@ -1,0 +1,211 @@
+import getServiceWidget from "utils/config/service-helpers";
+import createLogger from "utils/logger";
+import { asJson } from "utils/proxy/api-helpers";
+import { httpProxy } from "utils/proxy/http";
+
+const logger = createLogger("unraidProxyHandler");
+
+const UNRAID_DEFAULT_FIELDS = ["status", "cpu", "memoryAvailable", "notifications"];
+const MAX_ALLOWED_FIELDS = 4;
+
+const graphqlQuery = `
+{
+  array {
+    state
+    capacity {
+      kilobytes {
+        free
+        total
+        used
+      }
+    }
+    caches {
+      name
+      fsType
+      fsSize
+      fsFree
+      fsUsed
+    }
+  }
+  metrics {
+    memory {
+      active
+      available
+      percentTotal
+    }
+    cpu {
+      percentTotal
+    }
+  }
+  notifications {
+    overview {
+      unread {
+        total
+      }
+    }
+  }
+}
+`;
+
+function getBlock(field, data, widget) {
+  if (field.startsWith("pool")) {
+    const pool = field.match(/^pool[1-4]/)?.[0]; // Match pool1, pool2, etc.
+    if (pool) {
+      // Get the last part of the field, e.g., "UsedSpace" from "pool1UsedSpace"
+      const param = field.replace(pool, "");
+      const poolName = widget?.[pool] || pool;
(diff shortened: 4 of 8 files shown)


public/locales/en/common.json
```diff
@@ -1083,5 +1083,30 @@
...
+        "STOPPED": "Stopped",
+        "NEW_ARRAY": "New Array",
+        "RECON_DISK": "Reconstructing Disk",
+        "DISABLE_DISK": "Disk Disabled",
+        "SWAP_DSBL": "Swap Disable",
+        "INVALID_EXPANSION": "Invalid Expansion",
+        "PARITY_NOT_BIGGEST": "Parity Not Biggest",
+        "TOO_MANY_MISSING_DISKS": "Too Many Missing Disks",
+        "NEW_DISK_TOO_SMALL": "New Disk Too Small",
+        "NO_DATA_DISKS": "No Data Disks",
```
shamoon, line 1098: What are these used for?

docs/widgets/services/unraid.md
```diff
@@ -0,0 +1,25 @@
...
+
+Learn more about [Unraid](https://unraid.net/).
+
+The Unraid widget allows you to monitor the resources of an Unraid server.
+
+You will need an API key with the **GUEST** (read only) role: [Managing API Keys](https://docs.unraid.net/API/how-to-use-the-api/#managing-api-keys)
+
+The widget can display metrics for selected Unraid pools. If using one of the "pool" fields, you must also add the pool name to the settings.
+
+Allowed fields: `["cpu","memoryPercent","memoryAvailable","memoryUsed","notifications","arrayFreeSpace","arrayUsedSpace","arrayUsedPercent","status","pool1UsedSpace","pool1FreeSpace","pool1UsedPercent","pool2UsedSpace","pool2FreeSpace","pool2UsedPercent","pool3UsedSpace","pool3FreeSpace","pool3UsedPercent","pool4UsedSpace","pool4FreeSpace","pool4UsedPercent"]`
```
shamoon: Is 4 pools arbitrary? Yea Im concerned about the dynamic pools fields thing, and I do think it should use `fields` not `display`

shamoon (changes requested):
Thanks, interesting approach. I guess theres no real harm in releasing this widget for a beta version of unraid, but it should be noted in the docs.

If you can provide from sample API output that would be helpful, Im not sure I have the energy to get something like this setup in a VM or whatever.

shamoon:
Oh, and did you notice this doesnt yet meet the PR threshold? second checkbox

public/locales/en/common.json
dkaser: These are the possible values for array.state:

image

docs/widgets/services/unraid.md
dkaser: 4 pools was based on the fact that widgets are supposed to have no more than 4 blocks. An Unraid instance could have more pools than that though.

dkaser:
> Oh, and did you notice this doesnt yet meet the PR threshold? second checkbox

I did notice that, although given that there's a [dedicated installation page](https://gethomepage.dev/installation/unraid/) for Unraid it seemed that "community interest" was already met. (I meant to comment on that when I opened the PR, forgot to)

I'm certain that the discussion can get a few more upvotes though if that's what we need :)

dkaser:
> Thanks, interesting approach. I guess theres no real harm in releasing this widget for a beta version of unraid, but it should be noted in the docs.
> 
> If you can provide from sample API output that would be helpful, Im not sure I have the energy to get something like this setup in a VM or whatever.

I've added a note to the docs indicating the minimum OS requirements.

I have sample API output in the PR description, was there something else you were looking for?

There's also now 21 upvotes on the feature request. 👍

shamoon:
> I have sample API output in the PR description, was there something else you were looking for?

Missed that, thanks.

docs/widgets/services/unraid.md
shamoon: Yea I figured, sure thats fine but I do think we should use `fields`. I guess I'm a little unsure why not use the same construct as other widgets where the blocks are all there and let homepage do the filtering—with the exception of these pool fields which would need a bit of custom logic.

If / when we make broader changes to the app its really preferable that all widgets basically do the same thing wrt blocks

dkaser: Having a label that is both localizable and contains user-provided data (the pool name) is the challenge. It doesn't seem like with the current Component and Block there's a way to do both that doesn't result in the block just getting filtered out of view. (There's no way to pass values to the translated label, so you have to do t() inside the <Block /> tag, meaning it no longer matches.)

https://github.com/gethomepage/homepage/blob/6d7be1c7f219115d4fdb4a75a5abdb1fe37a76fb/src/components/services/widget/block.jsx#L16

A clean (but broader) solution would be to have Component filter on "key" first (or maybe add something like "field" so that it doesn't mess with anything already using key), and then "label" if it doesn't exist:

https://github.com/gethomepage/homepage/blob/6d7be1c7f219115d4fdb4a75a5abdb1fe37a76fb/src/components/services/widget/container.jsx#L40

`let matches = fullField === child?.props?.key ?? child?.props?.label;`

... but my initial approach was to keep the uniqueness to the widget instead of going for app-wide changes.

shamoon: I'd be ok with that, it could be helpful in other situations and yea it would allow this widget to go back to the more expected

```js
      <Container>
        <Block />
        <Block />
        <Block />
      </Container>
```

dkaser: Sounds good. I'll work on that some tonight.

dkaser: Updated; the change adds the ability to set a `field` attribute on a block, which is then used for filtering. If `field` is not set, it uses `label` as it does today.

Using `key` would have been problematic since some widgets currently set that to a value that is different than `label`. Filtering on `key` would change/break those widgets.

I looked through all of the custom components in widgets/ -- nothing uses `field` as a parameter, so it won't affect the behavior of any existing widgets.

shamoon: Nice thanks! I should have some time to test it out later

src/widgets/unraid/component.jsx
shamoon, line 1: Agree with the DRY part but why the `createElement`? It’s not really a homepage convention

dkaser: Because my brain decided to skip the simple solution and do the more complicated one 🙃

dkaser:
I'll have a small update to the docs tomorrow -- the Unraid staff are going to get me a permalink for the API key instructions.

src/widgets/unraid/proxy.js
```diff
@@ -0,0 +1,146 @@
...
+    query: graphqlQuery,
+  });
+
+  const [status, , data] = await httpProxy(url, params);
+
+  let resultData = data;
+
+  if (resultData.error?.url) {
+    resultData.error.url = sanitizeErrorURL(url);
+  }
```
shamoon:
```suggestion
  if (data.error?.url) {
    data.error.url = sanitizeErrorURL(url);
  }
```

src/widgets/unraid/proxy.js
```diff
@@ -0,0 +1,146 @@
...
+      "Error getting data from Unraid for service '%s' in group '%s': %d.  Data: %s",
+      service,
+      group,
+      status,
+      data,
+    );
+    return res.status(status).send({ error: { message: "Error calling Unraid API.", data } });
+  }
+
+  const result = processUnraidResponse(resultData);
```
shamoon:
```suggestion
  const result = processUnraidResponse(data);
```

src/widgets/unraid/component.jsx
```diff
@@ -0,0 +1,81 @@
...
+  const { t } = useTranslation();
+  const { widget } = service;
+
+  const { data, error } = useWidgetAPI(widget);
+
+  if (error) {
+    return <Container service={service} error={error} />;
+  }
+
+  if (!widget.fields?.length > 0) {
```
shamoon:
```suggestion
  if (!widget.fields?.length) {
```

src/widgets/unraid/proxy.js
```diff
@@ -0,0 +1,146 @@
...
+    "Content-Type": "application/json",
+    Accept: `application/json`,
+    "X-API-Key": `${widget.key}`,
+  };
+
+  const params = {
+    method: "POST",
+    headers,
+    withCredentials: true,
+    credentials: "include",
```
shamoon:
```suggestion
```

Think you dont need these here

src/widgets/unraid/proxy.js
```diff
@@ -0,0 +1,146 @@
...
+    response["arrayUsedPercent"] =
+      (data?.array?.capacity?.kilobytes?.used / data?.array?.capacity?.kilobytes?.total) * 100 ?? null;
+
+    response["caches"] = {};
+    if (data?.array?.caches) {
+      data.array.caches.forEach((cache) => {
+        if (cache.fsType) {
+          response.caches[cache.name] = {
+            fsFree: cache.fsFree * 1000,
+            fsUsed: cache.fsUsed * 1000,
```
shamoon, line 70: should these be 1024 if its kb

public/locales/en/common.json
```diff
@@ -1083,5 +1083,30 @@
...
+        "notifications": "Notifications",
+        "status": "Status",
+        "cpu": "CPU",
+        "memoryPercent": "Memory Used (%)",
```
shamoon:
```suggestion
        "memoryPercent": "Memory Used",
```

I think it's redundant to include these (%) given the number above will show `%`. In general people complain a lot about labels that are too long so trying to keep them short

public/locales/en/common.json
```diff
@@ -1083,5 +1083,30 @@
...
+        "memoryUsed": "Memory Used",
+        "memoryAvailable": "Memory Available",
+        "arrayUsedSpace": "Array Used",
+        "arrayFreeSpace": "Array Free",
+        "arrayUsedPercent": "Array Used (%)",
```
shamoon:
```suggestion
        "arrayUsedPercent": "Array Used",
```

public/locales/en/common.json
```diff
@@ -1083,5 +1083,30 @@
...
+        "poolUsedSpace": "{{pool}} Used",
+        "poolFreeSpace": "{{pool}} Free",
+        "poolUsedPercent": "{{pool}} Used (%)"
```
shamoon:
```suggestion
        "poolUsedPercent": "{{pool}} Used"
```

docs/widgets/services/unraid.md
```diff
@@ -0,0 +1,28 @@
...
+
+```yaml
+widget:
+  type: unraid
+  url: https://unraid.host.or.ip
+  key: api-key
+  pool1: # optional, name of pool for pool1... fields
+  pool2: # optional, name of pool for pool2... fields
+  pool3: # optional, name of pool for pool3... fields
+  pool4: # optional, name of pool for pool4... fields
```
shamoon:
```suggestion
  pool1: pool1name # required only if using pool1 fields
  pool2: pool2name # required only if using pool2 fields
  pool3: pool3name # required only if using pool3 fields
  pool4: pool4name # required only if using pool4 fields
```

src/widgets/unraid/proxy.js
dkaser: This is disk space, GUI measures in xB (not xiB), so we want 1000 here to match.

public/locales/en/common.json
shamoon, line 1: Please don’t reformat all this other stuff that’s not in the PR

dkaser: I thought I had undone that, will fix, vscode hid the whitespace changes from me on the diff

shamoon:
Thanks. If you don’t mind making sure this still works with your live unraid (as opposed to me testing sample api output) we should be good

dkaser:
Looks good!

shamoon (approved):

brankko:
I just saw the translation strings for Unraid and run here immediately to share my enthusiasm. I'm happy that Unraid support is coming, because workaround was to use Glances container and expose everything via API. Great work @dkaser

PR #5761, opened 2025-09-10
Update traefik.md for note a potential API Error solution

Rycochet:
## Proposed change

Add note about potentially needing `--api.insecure=true` in the traefik config to allow the Traefik widget API access. Included a link to the official Traefik docs for how to do this and only quoted the CLI change for brevity.

## Type of change

- [ ] New service widget
- [x] Documentation only
- [ ] Other (please explain)

### DIFF
--- a/docs/widgets/services/traefik.md
+++ b/docs/widgets/services/traefik.md
@@ -8,6 +8,8 @@ Learn more about [Traefik](https://github.com/traefik/traefik).
 No extra configuration is required.
 If your traefik install requires authentication, include the username and password used to login to the web interface.
 
+If you get an API Error then you may need to set `--api.insecure=false` to support this: <https://doc.traefik.io/traefik/master/operations/api/#insecure>
+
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks, since this really hasn't been an issue (for anyone?) in the ~3years the widget has been around Im not going to merge this and suggest that authoritatively it should be set. I think the circumstances are specific (for example, I dont have it set and use `https://traefik:443` as my url) and adding this line will potentially just create noise / confusion.

Thanks again.

Rycochet:
Given that I'm not using SSL / 443 and am on the standard 8080 port - with no special arguments for the api except `--api.dashboard=true` it's worth noting that this is going to be the case for some people who try it and it doesn't work - and then decide simply to not use it (or spend time figuring out how to make it work) - this will be in the history if anyone checks on github at least ;-)

shamoon:
Yea as a discussion might be more useful than in the docs, again given that it hasn’t really come up (I see one similar one but not exactly)

PR #5894, opened 2025-10-23
Unifi Widget: Allow both LAN and WLAN blocks to display

therumbler:
## Proposed change

Before this change some Unifi LAN stats would not show if WLAN was showing, and some WLAN stats would not show if LAN stats were showing.

This change allows both WLAN and LAN stats to display, regardless of the status of the other

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/unifi/component.jsx
+++ b/src/widgets/unifi/component.jsx
@@ -62,15 +62,11 @@ export default function Component({ service }) {
       {wan.show && <Block label="unifi.wan" value={wan.status === "ok" ? t("unifi.up") : t("unifi.down")} />}
 
       {lan.show && <Block label="unifi.lan_users" value={t("common.number", { value: lan.num_user })} />}
-      {lan.show && (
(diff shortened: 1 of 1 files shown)


shamoon:
Thanks but this is intentional behavior as it has been for 3 years or so, and Im not aware of a request to change it (let alone support for that change, as the PR template asks)

therumbler:
Thank you for the review! So.. for me i have multiple (non-wireless) switches, and (wireless) access points in Unifi, but Homepage won't let me see all those at once. Is that intentional?

shamoon:
I suppose the answer is 'no'. I dont have such strong feelings about changing this but your PR is a 'breaking change' in the sense that other users would suddenly see something different without changing anything, which we try to avoid. And again it's been 3 years. So the alternative would be to open a feature request and see if anyone else is interested and then either implement as a non-breaking change (unless there is some groundswell of support for it, but again, since youre the first to bring it up...)

PR #5913, opened 2025-10-31
Fix Authentik widget API endpoints for compatibility with Authentik 2025.10.0

gezuka77:
## Summary
The Authentik widget was using deprecated `/per_month/` endpoints that no longer exist in recent Authentik versions (tested with 2025.10.0), causing 404 errors and preventing login statistics from displaying (showing "NaN" instead).

## Changes
- Updated all endpoint mappings from deprecated `/events/events/per_month/` to current `/events/events/volume/` API
- Fixed query parameter syntax (changed `&&` to `&` in v2 endpoints)
- Simplified component logic to use consistent data format across versions
- Removed version-specific data processing since all versions now use the same endpoint format

## Testing
- Tested with Authentik 2025.10.0
- Confirmed the `/api/v3/events/events/volume/` endpoint works correctly
- Widget now displays login and failed login counts properly

## Fixes
Resolves the issue where the widget displayed "NaN" for login counts and generated HTTP 404 errors in the logs when calling the non-existent `/per_month/` endpoint.

## Related
This issue was discussed in the community and affects users running recent versions of Authentik where the `/per_month/` API endpoint has been removed or changed.

### DIFF
--- a/src/widgets/authentik/component.jsx
+++ b/src/widgets/authentik/component.jsx
@@ -11,11 +11,9 @@ export default function Component({ service }) {
 
   const { data: usersData, error: usersError } = useWidgetAPI(widget, "users");
 
-  const loginsEndpoint = widget.version === 2 ? "loginv2" : "login";
-  const { data: loginsData, error: loginsError } = useWidgetAPI(widget, loginsEndpoint);
+  const { data: loginsData, error: loginsError } = useWidgetAPI(widget, "login");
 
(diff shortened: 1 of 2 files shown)


shamoon:
One thing AI doesnt do well is that it makes obvious mistakes a human wouldn't, like breaking the backwards compatibility of the widget by changing the URLs for existing endpoints.

Thanks, rather than have Claude provide another revision I will take care of it.

shamoon:
https://github.com/gethomepage/homepage/discussions/5912#discussioncomment-14841979

PR #6006, opened 2025-11-23
Enhancement: support authentication for Frigate widget

Romloader:
## Proposed change

Frigate has implemented internal authentication via username/password.
This means that the handling of authentication is needed for the homepage widget too.
With this PR I am adding a custom proxy to frigate to handle exactly that.
Frigate authentication works via cookies, so the proxy firstly tries to login (if not already logged in) and then tries to use the retrieved cookie for API operations.

Related to #4697 #5399 #5649

## Type of change

- [ ] New service widget
- [X] New feature or enhancement (non-breaking change which adds functionality)
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/frigate/proxy.js
+++ b/src/widgets/frigate/proxy.js
@@ -21,8 +21,6 @@ export default async function frigateProxyHandler(req, res, map) {
     if (widget) {
       const url = formatApiCall(widgets[widget.type].api, { endpoint, ...widget });
 
-      let status, data;
-
       const params = {
         method: "GET",
         headers: {
@@ -30,7 +28,7 @@ export default async function frigateProxyHandler(req, res, map) {
         },
       };
 
-      [status, , data] = await httpProxy(url, params);
+      let [status, , data] = await httpProxy(url, params);
 
       if (status === 401 && widget.username && widget.password) {
         const loginUrl = `${widget.url}/api/login`;
@@ -49,7 +47,6 @@ export default async function frigateProxyHandler(req, res, map) {
             error: {
               message: `HTTP Error ${status} while trying to login to Frigate`,
               url: sanitizeErrorURL(url),
-              data: Buffer.isBuffer(data) ? Buffer.from(data).toString() : data,
             },
           });
         }
@@ -65,7 +62,6 @@ export default async function frigateProxyHandler(req, res, map) {
           error: {
             message: `HTTP Error ${status} from Frigate`,
             url: sanitizeErrorURL(url),
-            data: Buffer.isBuffer(data) ? Buffer.from(data).toString() : data,
           },
         });
       }


Romloader:
Frigate authentication goes through a `/login` endpoint which sends back a cookie with a `frigate_token` and some other parameters like expiration date (which is so in the future it's like it's not there from my tests).
Then, requests need to send that `frigate_token` cookie to the api for a correct use.
The user system in frigate is quite simple, with an admin and viewer roles: in theory the user is supposed to create a viewer user because homepage does not need access to settings, but that is up to the user.

I have tried to make use of the cookie jar from the `utils` folder to avoid requesting another cookie if one is present already, and at the same time, a failure of auth with the previous cookie triggers a login to try and get an updated one.
The proxy code should handle each relevant http response code.

However, this is the first time I submit a PR to Homepage so please let me know if you would have done anything differently or if there is anything to change here.
I also think I am not requesting a login (+ cookie) unless needed (so basically I should not be creating unnecessary sessions) but if that's not the case just tell me and I'll fix that.

The previous behavior, so without auth, is intact: the proxy checks if the widget has `username` and `password` set and uses authentication only if they are present.

Romloader:
One approach that isn't included is that frigate supports authentication with reverse proxies passing headers to it, but I am not sure if it would be applicable in the context of a homepage widget.
Anyway, for now internal accounts are handled.

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!widgets?.[widget.type]?.api) {
+      return res.status(403).json({ error: "Service does not support API calls" });
+    }
+
+    if (widget) {
+      // If there are more than one question marks, replace others to &
+      let urlString = formatApiCall(widgets[widget.type].api, { endpoint, ...widget }).replace(/(?<=\?.*)\?/g, "&");
+      if (widget.type === "customapi" && widget.url?.endsWith("/")) {
+        urlString += "/"; // Ensure we dont lose the trailing slash for custom API calls
```
shamoon: So did you use a LLM to generate this PR?

Romloader: I did not, i took inspiration from the general purpose proxy, the npm one, and the openmediavault one

If i remember correctly that check for question marks and trailing slash is in the generic proxy too
Update: found it https://github.com/gethomepage/homepage/blob/cf990063b9f267a96ad82dc6f445d75045378855/src/utils/proxy/handlers/generic.js#L21-L25
https://github.com/gethomepage/homepage/commit/6e16adc460909c90b479070700ce55b58dbd1ea1

But yes looking at those lines without the context, they do appear like something that slipped while I was writing the functionality, and looking at it popst-mortem they might not be needed even.
Might have to recheck that if it doesn't make much sense

I spent a couple of days on this pr because i genuinely wanted to help, if there are mistakes they are my mistakes

Romloader: Hello there!

Now that I got the time I checked the code again, and I noticed I was actually confusing the concept of custom api (as in having the custom connection behavior to the service) and the homepage [Custom API](https://gethomepage.dev/widgets/services/customapi/) widget.
It does make sense to be present in the generic proxy indeed because it handles all basic use cases, but has no use here.
I'll remove that with another commit.

Apart from that, I am checking again if we need that check for the url still, but probably not because the user needs to set it up correctly in the first place anyway.
One thing that could still be useful maybe is checking for the trailing slash just in case.

What do you think @shamoon?
And yes now I understand why you were wondering about LLM, it didn't make much sense to have `customapi` there

shamoon: Yea that is why I was wondering. Yes that wouldn’t really apply here, feel free to push any changes but I’ll give it a proper review when I have a second

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+      const url = new URL(urlString);
+      const loginUrl = `${widget.url}/api/login`;
+
+      let status, contentType, data, responseHeaders;
+
+      const params = {
+        method: "GET",
+        headers: {
+          "Content-Type": "application/json",
+          //Cookie: `frigate_token=${token}`,
```
shamoon:
```suggestion
```

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    if (widget) {
+      // If there are more than one question marks, replace others to &
+      let urlString = formatApiCall(widgets[widget.type].api, { endpoint, ...widget }).replace(/(?<=\?.*)\?/g, "&");
+      if (widget.type === "customapi" && widget.url?.endsWith("/")) {
+        urlString += "/"; // Ensure we dont lose the trailing slash for custom API calls
+      }
+      const url = new URL(urlString);
+      const loginUrl = `${widget.url}/api/login`;
```
shamoon:
```suggestion
      const url = formatApiCall(widgets[widget.type].api, { endpoint, ...widget });
      const loginUrl = `${widget.url}/api/login`;
```

Yea, this proxy will only be used by frigate, so `widget.type` should always be `frigate` (and not `customapi`)

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+        resultData.error.url = sanitizeErrorURL(url);
+      }
+
+      if (status === 200) {
+        if (!validateWidgetData(widget, endpoint, data)) {
+          return res
+            .status(status)
+            .json({ error: { message: "Invalid data", url: sanitizeErrorURL(url), data: data } });
+        }
+        if (map) data = map(data);
```
shamoon: Now that this will have a custom proxy, just move the `map` calls out of widget.js and put helpers here, its much easier to track

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+      if (status == 401) {
+        if (widget.username || widget.password)
+          [status, contentType, data, responseHeaders] = await tryLogin(
+            loginUrl,
+            widget.username,
+            widget.password,
+            service,
+          );
+
+        if (status === 200) {
```
shamoon: There almost certainly should not be two separate `status === 200` blocks

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+          addCookieToJar(url, responseHeaders);
+          setCookieHeader(url, params);
+          [status, , data] = await httpProxy(url, params);
+        } else if (status === 401) {
```
shamoon: Again, why two 401 blocks

src/widgets/frigate/proxy.js
```diff
@@ -0,0 +1,139 @@
...
+          url.pathname,
+        );
+        return res.status(status).json({
+          error: {
+            message: `HTTP Error ${status} from Frigate`,
+            url: sanitizeErrorURL(url),
+            data: Buffer.isBuffer(data) ? Buffer.from(data).toString() : data,
+          },
+        });
+      }
```
shamoon: Same as above, there are way too many if / else clauses here. Generally:

200 = good
401 (or 403, depending on frigate I dont know) = auth
everything else > 403 = error

shamoon:
Oh nice, they have a demo 😎

Screenshot 2025-11-24 at 12 12 59 PM

shamoon:
I’ll take care of the testing part, you don’t really need to worry about that

Romloader:
Ah, I see, I was wandering around trying to find an RTSP link to provide for testing, phew (anyway, docker example for Frigate (only) here at https://docs.frigate.video/frigate/installation/#docker)
I'll apply the suggested corrections

src/widgets/frigate/proxy.js
Romloader: From my testing on my own instance Frigate returns a 401 Unauthorized usually and not a 403

This was to try directly using the cached token if we have one with an API request
Then if that fails try to login and handle the response codes accordingly, but yes that duplicates if/else checks
We could simplify the logic as you said yes, I'll commit a refactored version

src/widgets/frigate/proxy.js
Romloader: The second one was handling the status of the api call after the login function.
I'll modify it according to https://github.com/gethomepage/homepage/pull/6006#discussion_r2557590294 then

src/widgets/frigate/proxy.js
Romloader: https://github.com/gethomepage/homepage/pull/6006#discussion_r2557866055

src/widgets/frigate/proxy.js
Romloader: Mappings moved accordingly  https://github.com/gethomepage/homepage/pull/6006/commits/db2ddeca734f8e1c8a30837704dda4e22223c927
I hope I got what you said properly

Romloader:
Took the opportunity to apply your latest commit to the PR too

shamoon:
I did a bit more cleanup. If you want to just make sure this still authenticates with your instance that would be great. Still works with the demo.

One question is whether this is accurate:

https://github.com/Romloader/homepage/blob/4d44915900171ad01aaad0cfe2c69b8cc15df29e/src/widgets/frigate/proxy.js#L36

As in, do you actually always need username and password (not OR)?

Romloader:
Saw your commits, nice stuff, I am always learning more about the repo structure from this pr.
And yes both username and passwords are needed: in Frigate, you either login without credentials or with a username and password.

I just tried from Frigate GUI and you can't set up an user without a password, it just doesn't let you.

Romloader:
Did a quick build with the new commits but I get a 
```
error: ReferenceError: username is not defined
at frigateProxyHandler (webpack-internal:///(api-node)/./src/widgets/frigate/proxy.js:50:31)
```
Will check back later in case I need to recompile or delete the bundler cache etc cause I'm a bit busy right now

Romloader:
I modified user and password parameters because they kept the function parameter names
I now get a 
```
error: SyntaxError: Unexpected token '<', "<html>
<h"... is not valid JSON
    at JSON.parse (<anonymous>)
    at asJson (webpack-internal:///(api-node)/./src/utils/proxy/api-helpers.js:43:27)
    at frigateProxyHandler (webpack-internal:///(api-node)/./src/widgets/frigate/proxy.js:82:83)
```
Will check back later

shamoon:
This works for me currently, let me know

Screenshot 2025-11-25 at 9 44 22 AM

Screenshot 2025-11-25 at 9 48 22 AM

Romloader:
Yes it's working fine here too!
image
Events work too (you can understand why I didn't put a screenshot for them)

There is one last thing I wonder: frigate gives tokens that are valid for >50 years in my testing (if I remember correctly).
Since their lifetime is not so limited, if you change credentials in the widget, and then change them to something invalid, then the stats will keep working because the http proxy has the previous cookie in its cache.
Is this something that will be left to the user to figure out? Should we put something in the docs that says to restart the instance in case you change account? Should we just do nothing and the user should use a grain of salt?

src/widgets/frigate/proxy.js
```diff
@@ -96,9 +96,7 @@ export default async function frigateProxyHandler(req, res, map) {
           uptime: data?.service?.uptime,
           version: data?.service.version,
         });
-      }
-
-      if (endpoint == "events") {
+      } else if (endpoint == "events") {
```
Romloader: Curiosity: what is the difference between having two separate `if` blocks and an `else if` here?
It does "feel" cleaner to the eye but the behaviour would be the same, right?

shamoon:
That sounds like a frigate issue, ie it should invalidate the cookie and it isnt. Either way, no not something we need to handle. I dont imagine users are going to be changing frigate users on-the-fly...

Romloader:
Fair enough then

shamoon (approved):

PR #6366, opened 2026-02-27
fix(resources): use direct networkStats(iface) call for named interfaces in Docker

jonathan0104:
## Description

When Homepage runs inside a Docker container, the network widget fails to return stats for a named host interface (e.g. `enp6s18`), returning a 404 even when `/sys` is correctly mounted.

## Root Cause

`networkStats("*")` enumerates interfaces via `os.networkInterfaces()`, which is Docker network-namespace-aware. Inside a container, this only returns container interfaces (`eth0`, `lo`) — never the host NIC. The subsequent `.filter()` call finds no match and returns 404.

This is a different failure mode from the one fixed in #6102. That PR correctly added `"*"` to enumerate all interfaces, but the enumeration itself is broken in Docker for host NICs.

## Fix

When an explicit `interfaceName` is provided, call `si.networkStats(interfaceName)`  directly instead of `networkStats("*")` + filter. This reads from `/sys/class/net/<iface>/statistics/` via sysfs, bypassing namespace enumeration entirely and correctly returning host NIC stats.

The default interface path (`network: true`) is unchanged.

## Changes

- `src/pages/api/widgets/resources.js` — restructured named-interface path
- `src/__tests__/pages/api/widgets/resources.test.js` — updated existing test, added new test for Docker namespace fix
- `docs/widgets/info/resources.md` — added callout documenting required `/sys/class/net:/sys/class/net:ro` volume mount for named host NICs

## Testing

Tested on Homepage v1.10.1 running in Docker with host interface `enp6s18` and `/sys:/sys:ro` mounted. Widget correctly returns live rx/tx stats after this fix.

## AI Assistance

This fix was developed with AI assistance (Claude). The bug was discovered and diagnosed while debugging my own Homepage deployment, the root cause analysis and fix were developed collaboratively, and the tests and documentation were written with AI help. The fix has been verified working on my own system.

### DIFF
--- a/docs/widgets/info/resources.md
+++ b/docs/widgets/info/resources.md
@@ -15,6 +15,15 @@ _Note: unfortunately, the package used for getting CPU temp ([systeminformation]
 
 **Any disk you wish to access must be mounted to your container as a volume.**
 
+**To monitor a named host network interface** (e.g. `network: enp6s18`), the host sysfs must be accessible inside the container. Add the following volume mount to your Docker Compose service:
+
+```yaml
+volumes:
+  - /sys/class/net:/sys/class/net:ro
+```
+
+Without this mount, the widget can only report statistics for the container's own network interfaces, not the host machine's physical NICs. Using `network: true` (which monitors the container's default interface) works without any additional mounts.
+
 ```yaml
 - resources:
     cpu: true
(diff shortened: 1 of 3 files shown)


shamoon:
Im going to close this PR. While I appreciate the effort, this highlights the issue with LLM-generated PRs: whether the diagnosis is correct or not, this PR most certainly is not. It incorrectly treats missing interfaces as returning `[]`, which they do not. This is not to mention the behavior on macOS, WSL, etc which I assume you have not tested. Rather than have you pipe this feedback to your LLM for a revision, I will look into this.

In the future, I would suggest stopping at the diagnosis point, while I appreciate that LLMs "lower the barrier" to opening a PR, in reality, like seen here what they often are doing is shifting the burden to us to actually figure out how the PR works (or doesnt).

Thanks for your understanding

PR #6468, opened 2026-03-27
Documentation: fix kubernetes config examples

mylegitches:
## Summary

Fix incorrect Kubernetes configuration examples in :

- change  to 
- change  to 

## Why

The surrounding text says to set cluster mode and enable gateway compatibility, but the examples showed different values. This updates the examples to match the documented behavior.


shamoon (approved):
Thanks

PR #6477, opened 2026-03-28
Fix: fix compatibility with flood changes

SkeLLLa:
## Proposed change

Fixes flood widget to work without credentials by using proper value empty `body` . In new versions of flood fastify+zod validation is used. Therefore `body: null` in combination of request header `content-type: application/json` is invalid, since null is not proper JSON.

For more info see `FST_ERR_CTP_EMPTY_JSON_BODY` (in case of `null` in body) and `FST_ERR_VALIDATION` (in case of `""`) fastify errors.

Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [ ] If applicable, I have reviewed the [feature / enhancement](https://gethomepage.dev/widgets/authoring/getting-started/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/widgets/authoring/getting-started/#service-widget-guidelines).

### DIFF
--- a/src/widgets/flood/proxy.js
+++ b/src/widgets/flood/proxy.js
@@ -12,7 +12,7 @@ async function login(widget) {
   const loginParams = {
     method: "POST",
     headers: { "Content-Type": "application/json" },
-    body: null,
+    body: "{}",
   };
 
   if (widget.username && widget.password) {
--- a/src/widgets/flood/proxy.test.js
+++ b/src/widgets/flood/proxy.test.js
@@ -45,7 +45,7 @@ describe("widgets/flood/proxy", () => {
     expect(httpProxy).toHaveBeenCalledTimes(3);
     expect(httpProxy.mock.calls[0][0].toString()).toBe("http://flood/api/stats");
     expect(httpProxy.mock.calls[1][0]).toBe("http://flood/api/auth/authenticate");
-    expect(httpProxy.mock.calls[1][1].body).toBeNull();
+    expect(httpProxy.mock.calls[1][1].body).toBe("{}");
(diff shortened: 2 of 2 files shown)


shamoon:
A bit more info about where this came from (ie was introduced to flood) would be helpful since we haven’t gotten any reports about it. And have you checked this is backwards compatible?

SkeLLLa:
@shamoon sure.

It was introduced with release https://github.com/jesec/flood/releases/tag/v4.12.1 where express was replaced by fastify. Regardless `fix!` which means breaking change the version bump was semver minor (but I don't dig into flood's versioning strategies). Fastify framework is much stricter about types and standards. So the combination of `null` in body and `content-type application/json` was technically incorrect from the start, since `null` value is not valid json. But express framework allowed lot's of incorrect stuff to work.

Considering backward compatibility it will be compatible, since this line: https://github.com/jesec/flood/blame/28b3e53f04d1d31b821321fb0af9c0b7531e7ceb/server/routes/api/auth.ts#L75 (on express) and now in fastufy: https://github.com/jesec/flood/blob/master/server/routes/api/auth.ts#L90. So if in flood you had no auth, it will exit early with default credentials regardless of what actually will be in the `body`. 
And now it's failing because of validation happen much earlier and it doesn't even reach this code.

> we haven’t gotten any reports about it

It might be that many flood users use flood's auth instead of external one. Also there's a simple workaround - just supply any username/password and they will not be actually used. And speaking of reports that's how it looks right now with the latest version of flood & no username/password in config.
image

shamoon (approved):
Great thanks

PR #6525, opened 2026-04-05
Handle bookmark URLs without protocol in hostname fallback

Vellis59:
## Summary
This small change avoids failures when a bookmark URL is entered without a protocol.

Homepage already handles protocol-less URLs in some places (for example Quick Launch), but bookmark hostname fallback was still doing a direct `new URL(bookmark.href)`, which can throw for values like `example.com`.

## Changes
- add a small shared helper to normalize protocol-less URLs
- use it in bookmark hostname fallback
- add tests for the helper

## Example
Before:
- `href: example.com`
- bookmark hostname fallback could throw when no custom description was set

After:
- `href: example.com`
- fallback hostname resolves correctly to `example.com`

I’m working on a fork of Homepage and ran into this while testing real-world configs, so I thought this small fix would also make sense upstream.

### DIFF
--- a/src/components/bookmarks/item.jsx
+++ b/src/components/bookmarks/item.jsx
@@ -2,9 +2,10 @@ import classNames from "classnames";
 import ResolvedIcon from "components/resolvedicon";
 import { useContext } from "react";
 import { SettingsContext } from "utils/contexts/settings";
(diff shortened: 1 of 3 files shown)


shamoon:
Feel free to maintain in your own fork, there havent been any open requests / reports around this.

PR #6569, opened 2026-04-16
fix(immich): default to v2 API endpoints, fixing breakage on Immich >= v2.7

emak:
## Summary

- Immich v2.7.x removed the legacy `/server-info/*` API endpoints entirely, causing the widget to return 404s and break for any user who hadn't manually set `version: 2` in their config
- Changes the default `widget.version` from `1` to `2` in `component.jsx` so the widget works out of the box with all current Immich installs
- Users on very old Immich (< v1.118, which pre-dates the `/server/` endpoints) can opt back in with `version: 1`
- Updates docs and tests to reflect the new default

## Root cause

The Immich project introduced `/server/*` endpoints in v1.118 and, as of v2.7.x, **removed** the old `/server-info/*` endpoints. Homepage already had full support for the new endpoints under the `version: 2` config flag — the only problem was that the default remained `1`, so most users were silently broken after upgrading Immich.

Verified against the [Immich v2.7.5 OpenAPI spec](https://github.com/immich-app/immich/blob/v2.7.5/open-api/immich-openapi-specs.json): `/server-info/*` paths are absent, and `/server/version` + `/server/statistics` are present with identical response shapes to what the widget already expects.

## Test plan

- [x] `src/widgets/immich/component.test.jsx` — updated default-behaviour test to expect `version_v2`/`statistics_v2` endpoints; legacy v1 test now requires explicit `version: 1`
- [x] `src/widgets/immich/widget.test.js` — added assertions for the v2 endpoint mappings
- [x] All 4 Immich tests pass (`pnpm test`)
- [ ] Manual smoke test against a live Immich >= v2.7 instance recommended before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -83,7 +83,7 @@ jobs:
 
       - name: Log into registry ${{ env.REGISTRY }}
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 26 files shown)


shamoon:
1. Ai slop bypassing the PR template
2. Unclean commit history
3. A breaking change no one asked for

we’ll close this out, thanks

PR #6784, opened 2026-06-18
Feature: Pulse service widget

Pootch:
## Proposed change
Adds a service widget for [Pulse](https://github.com/rcourtman/Pulse), 
a real-time monitoring dashboard for Proxmox VE, PBS, and Docker/host infrastructure.

The widget shows:
- Nodes online (X/Y)
- VMs running (X/Y)  
- LXCs running (X/Y)

Example API response (`GET /api/resources`):
{"resources": [{"type": "node", "status": "online", "name": "pve1"}, ...]}

## Type of change
✅ New service widget

## Checklist:
✅ Documentation added (docs/widgets/services/pulse.md)
❌ Tests (not yet added)
✅ Reviewed service widget guidelines
❌ Pre-commit hooks (not yet run)
✅ AI tools disclosure: This PR was developed with AI assistance (Claude by Anthropic)

### DIFF
--- a/docs/widgets/services/pulse.md
+++ b/docs/widgets/services/pulse.md
@@ -1,0 +1,20 @@
+---
+title: Pulse
+description: Pulse Widget Configuration
+---
+
+Learn more about [Pulse](https://github.com/rcourtman/Pulse).
+
+Pulse is a real-time monitoring dashboard for Proxmox VE, Proxmox Backup Server, and Docker/host infrastructure.
+
+An API token with the **Kiosk / Dashboard** scope is required. Generate one in Pulse under **Settings → Security → API Tokens**.
+
+Allowed fields: `["nodes", "vms", "lxcs"]`
+
+```yaml
+widget:
+  type: pulse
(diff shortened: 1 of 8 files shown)


shamoon:
I hate to ask, but did you actually test this or just use an LLM? It doesnt work and testing it would've shown the obvious reason why

Screenshot 2026-06-18 at 2 56 58 PM

Pootch:
You're right, thanks for catching this. The previous version used `headers: {"X-API-Token": "{key}"}` in `widget.js`, but the generic proxy handler doesn't substitute template strings inside headers — so the token was never actually sent.

Fixed by switching to `credentialedProxyHandler` and adding `pulse` to the `X-API-Token` header list in `credentialed.js` (same pattern as `autobrr`/`jellystat`). Verified directly against a live Pulse instance — confirmed 200 response and correct node/VM/LXC counts rendering in the widget.

Should have tested this properly before opening the PR. Apologies for the wasted review cycle — re-tested end to end this time.

PR #6811, opened 2026-06-26
fix(config): skeleton fallback for read-only config dir

### LINKED ISSUE #2172: [Bug] Kubernetes Deployment Fails Due to RO ConfigMap
### Description

With the kubernetes deployment in https://gethomepage.dev/v0.7.3/installation/k8s/#deployment, you are told to do the following:
```yaml
volumeMounts:
                      - name: homepage-config
                        mountPath: /app/config
                      - name: logs
                        mountPath: /app/config/logs
            volumes:
                - name: homepage-config
                  configMap:
                      name: homepage
```

This makes the /app/config folder _readonly_ which results in `error: Error: EROFS: read-only file system, copyfile '/app/src/skeleton/custom.css' -> '/app/config/custom.css'` Making some settings breaks.

### Steps to reproduce

Create homepage following the Kubernetes docs
See erorr

### homepage version

v0.7.3

### Installation method

Other (please describe above)

### Configuration

_No response_

### Container Logs

error: Error: EROFS: read-only file system, copyfile '/app/src/skeleton/custom.css' -> '/app/config/custom.css'

### Browser Logs

_No response_

### Troubleshooting

I was able to fix this, by mounting the configmap to /tmp/config, and adding the following hook:
```yaml
lifecycle:
            postStart:
              exec:
                command: ["/bin/sh", "-c", "cp -r -L /tmp/config/* /app/config/; chmod -R 777 /app/config"]
````

The CSS files should be copied into another dir, or mount the config map separately and copy it.

### Other

_No response_

### Before submitting, I have made sure to

- [X] Check [the documentation](https://gethomepage.dev/)
- [X] Follow [the troubleshooting guide](https://gethomepage.dev/latest/more/troubleshooting/) (please include output above if applicable).
- [X] Search [existing issues](https://github.com/gethomepage/homepage/search?q=&type=issues) and [discussions](https://github.com/gethomepage/homepage/search?q=&type=discussions).

emanuelbesliu:
## Proposed change

Homepage calls `process.exit(1)` when it cannot copy a default config file into the config directory. This happens whenever the config dir is **read-only and a file is missing** — most commonly when individual files are mounted via `subPath` from a Kubernetes ConfigMap (which makes the dir non-writable). The missing file triggers an `EACCES`/`EROFS` copy error and the whole app dies with CrashLoopBackOff.

#2178 (merged) only documented a workaround in the K8s docs — enumerate every config file as a `subPath` mount. That is fragile: each release that adds a new skeleton file silently re-breaks deployments until the user mounts it too. The `process.exit(1)` is still in `config.js` on `dev` today.

This PR makes config loading resilient instead:

- `checkAndCopyConfig()` now **warns and continues** rather than exiting when it cannot seed a default file.
- A new `getConfigPath()` helper resolves each read to the user's file when present, otherwise to the bundled `src/skeleton/` copy. Skeletons are valid (mostly-commented) YAML/CSS/JS, so Homepage starts with whatever subset of files the user provided.
- All readers (`settings`, `services`, `docker`, `widgets`, `bookmarks`, `kubernetes`, `proxmox`, `/api/hash`) go through `getConfigPath()`.

Closes #2172

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [x] If applicable, I have added or updated tests for new features and bug fixes.
- [ ] If applicable, I have reviewed the feature / service widget guidelines.
- [x] I have checked that all code style checks pass (eslint clean on touched files).
- [ ] If applicable, I have tested my code on mobile & desktop.

Tests: updated the copy-failure test to assert warn-and-continue (no exit) plus `getConfigPath` cases, and added `getConfigPath` to the affected mocks. AI tools (GitHub Copilot CLI) assisted; the diagnosis, design, code and tests were human-reviewed and verified locally.

### DIFF
--- a/src/__tests__/pages/api/hash.test.js
+++ b/src/__tests__/pages/api/hash.test.js
@@ -7,10 +7,11 @@ function sha256(input) {
   return createHash("sha256").update(input).digest("hex");
 }
 
-const { readFileSync, checkAndCopyConfig, CONF_DIR } = vi.hoisted(() => ({
+const { readFileSync, checkAndCopyConfig, CONF_DIR, getConfigPath } = vi.hoisted(() => ({
   readFileSync: vi.fn(),
   checkAndCopyConfig: vi.fn(),
   CONF_DIR: "/conf",
+  getConfigPath: vi.fn((c) => `/conf/${c}`),
 }));
 
 vi.mock("fs", () => ({
@@ -20,6 +21,7 @@ vi.mock("fs", () => ({
 vi.mock("utils/config/config", () => ({
   default: checkAndCopyConfig,
   CONF_DIR,
+  getConfigPath,
 }));
 
 import handler from "pages/api/hash";
--- a/src/pages/api/hash.js
+++ b/src/pages/api/hash.js
@@ -1,8 +1,7 @@
 import { createHash } from "crypto";
 import { readFileSync } from "fs";
-import { join } from "path";
 
-import checkAndCopyConfig, { CONF_DIR } from "utils/config/config";
+import checkAndCopyConfig, { getConfigPath } from "utils/config/config";
 
 const configs = [
   "docker.yaml",
@@ -23,7 +22,7 @@ function hash(buffer) {
 export default async function handler(req, res) {
   const hashes = configs.map((config) => {
     checkAndCopyConfig(config);
-    const configYaml = join(CONF_DIR, config);
+    const configYaml = getConfigPath(config);
     return hash(readFileSync(configYaml, "utf8"));
   });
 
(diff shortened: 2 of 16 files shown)


shamoon:
Thanks for working on this, but I’m going to close this one.

The original Kubernetes read-only ConfigMap issue was handled in #2178 by documenting explicit `subPath` mounts for each config file, and we haven’t seen enough recent reports to justify changing the core config-loading mechanism.

My main concern is that this significantly changes homepage's behavior, which almost guarantees it will cause more issues than it solves. If a Kubernetes user forgets to mount a file, or any deployment has a broken config path, Homepage can now start with example config instead of failing. `checkAndCopyConfig()` also returns `true` after copy failure, so validation no longer makes the initialization problem obvious.

That feels more confusing than the [edge case] issue it solves. I’d rather keep the documented explicit-file mount approach

Thanks again

PR #6842, opened 2026-07-06
Add Guacamole widget

joszz:
## Summary
- Adds a `guacamole` widget that shows active session, connection, and user counts for an [Apache Guacamole](https://guacamole.apache.org/) server.
- Guacamole's REST API authenticates via `POST /api/tokens` (username/password) and requires the returned token as a `?token=` query param on every subsequent request, so this adds a custom proxy handler (`src/widgets/guacamole/proxy.js`) that logs in, caches the token + resolved data source, and re-authenticates automatically on a 401/403 — modeled on the existing `npm` widget's login/retry pattern.
- An optional `datasource` config field lets users pin a specific auth provider (e.g. `ldap`) instead of the default one returned at login.

## Test plan
- [x] `pnpm vitest run src/widgets/guacamole` — 7 tests passing (widget config shape, proxy login/retry/datasource-override, component loading/error/data states)
- [x] `pnpm eslint` / `pnpm prettier --check` clean on new files
- [x] Ran `pnpm dev` locally against a live Guacamole instance and hit `/api/services/proxy` directly for all three mappings (`activeConnections`, `connections`, `users`); each returned real data (correct auth token flow, correct counts) matching the instance's actual state

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/widgets/services/guacamole.md
+++ b/docs/widgets/services/guacamole.md
@@ -1,0 +1,21 @@
+---
+title: Guacamole
+description: Guacamole Widget Configuration
+---
+
+Learn more about [Apache Guacamole](https://guacamole.apache.org/).
+
+This widget shows the number of active sessions, total configured connections, and total users for a Guacamole server.
+
+Use the username and password of an account that has permission to view connections and users on the data source you want to report on.
+
+```yaml
+widget:
+  type: guacamole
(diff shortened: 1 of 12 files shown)


joszz (commented):
.

PR #817, opened 2023-01-13
use abbr as icon fallback

Jose134:
Hello, I submitted a [feature request](https://github.com/benphelps/homepage/issues/814) and was able to implement it myself. This uses the abbr property of bookmarks as a fallback in case it doesn't find the icon by replacing the "logo" string in the image's alt text. It's a tiny change, but I think it's a good one since previously abbr would just be ignored if the icon failed to load.

Here's a screenshot:
Screenshot from 2023-01-13 22-44-05

Using the following bookmarks.yaml:
```
---
# For configuration options and examples, please see:
# https://gethomepage.dev/en/configs/bookmarks

- Dashboard icon:
    - Icon with abbr:
        - icon: github-light.png
          abbr: GH
          href: https://github.com/
    - Icon without abbr:
        - icon: github-light.png
          href: https://github.com/
    - Bad icon with abbr:
        - icon: github-light1.png
          abbr: GH
          href: https://github.com/
    - Bad icon without abbr:
        - icon: github-light1.png
          href: https://github.com/

- Direct or Relative URL:
    - Icon with abbr:
        - icon: /favicon-32x32.png
          abbr: RE
          href: https://github.com/
    - Icon without abbr:
        - icon: /favicon-32x32.png
          href: https://github.com
    - Bad icon with abbr:
        - icon: /notanimage.png
          abbr: RE
          href: https://github.com/
    - Bad icon without abbr:
        - icon: /notanimage.png
          href: https://github.com/

- MDI Icon:
    - Icon with abbr:
        - icon: mdi-calendar-badge.svg
          abbr: GH
          href: https://github.com/
    - Icon without abbr:
        - icon: mdi-calendar-badge.svg
          href: https://github.com/
    - Bad icon with abbr:
        - icon: mdi-doesntexist.svg
          abbr: GH
          href: https://github.com/
    - Bad icon without abbr:
        - icon: mdi-doesntexist.svg
          href: https://github.com/

- No icon:
    - With abbr:
        - abbr: GH
          href: https://github.com
    - Without abbr:
        - href: https://github.com
```

Issue #254, opened 2022-09-23
[Bug] Docker Integration doesn't work with Podman


leruaa:
I've configured the docker integration for my services and I see the little green dot, but when I click on it, I have the following error:

> Application error: a client-side exception has occurred (see the browser console for more information).

Looking at the browser console I see:

```
TypeError: Cannot read properties of undefined (reading 'rx_bytes')
    at D (index-f387f442492a0fa8.js:1:8855)
    at Eo (framework-b3ca2e9a9574e304.js:1:63191)
    at Pu (framework-b3ca2e9a9574e304.js:1:74879)
    at xi (framework-b3ca2e9a9574e304.js:1:121314)
    at bs (framework-b3ca2e9a9574e304.js:1:109367)
    at gs (framework-b3ca2e9a9574e304.js:1:109295)
    at vs (framework-b3ca2e9a9574e304.js:1:109158)
    at as (framework-b3ca2e9a9574e304.js:1:105938)
    at ls (framework-b3ca2e9a9574e304.js:1:104489)
    at S (framework-b3ca2e9a9574e304.js:1:138448)
```

I think the error is from there: https://github.com/benphelps/homepage/blob/0b43f83daabae5655c1c950377e37fd1ca3f0631/src/components/services/widgets/service/docker.jsx#L57

When I call the following API (corresponding to one of my services : `/api/docker/stats/gitea/my-docker`, I get that:

```json
      "networks":{
         "network":{
            "rx_bytes":8465590,
            "rx_packets":0,
            "rx_errors":0,
            "rx_dropped":0,
            "tx_bytes":41127436,
            "tx_packets":0,
            "tx_errors":0,
            "tx_dropped":0
         }
      }
```

Looks like the network is expected to be `eth0`, but with Podman, it's `network`.

It would be great if we can support Podman :)

benphelps:
Thank you for the details, we'll get this fixed in the next release.

benphelps:
This will be fixed once fe7fa5c goes out.

PR #6847, opened 2026-07-08
feat(dockhand): support API token (Bearer) authentication

jheaton141:
## What

Adds an optional `key` to the Dockhand widget so it can authenticate with a Dockhand API token (sent as `Authorization: Bearer <key>`) instead of a username/password.

## Why

Dockhand added user-scoped API token auth in v1.0.25 (docs: https://dockhand.pro/manual/#api-authentication), but the widget still only supported local username/password login. Putting a full account username/password in `services.yaml` grants the widget the whole account for what is a read-only status readout. An API token is scoped to the creating user and can be revoked without changing the password — the safer, more conventional pattern for a dashboard integration.

Dockhand implemented this partly with homepage in mind (Finsys/dockhand#729, also #543 / #817 / #254).

## Behavior

- If `key` is set, it is sent as a bearer token and takes precedence over username/password.
- The existing 401 -> session-login retry is skipped when a token is used (a 401 with a token means the token itself is invalid, so retrying via login is wrong).
- No change to the existing username/password flow.

## Testing

**Unit** (`src/widgets/dockhand/proxy.test.js`) — added two cases: the bearer token is sent with no login call, and there is no login fallback on a 401 when a token is configured. Existing tests unchanged. `vitest run`, `prettier --check`, and `eslint` all pass.

**Verified against a real instance** — tested end-to-end against a live **Dockhand v1.0.36** deployment on the `/api/dashboard/stats` endpoint the widget uses:

| Scenario | Result |
| --- | --- |
| No auth header | `HTTP 401` (auth enforced) |
| Invalid bearer token | `HTTP 401` |
| Valid `Authorization: Bearer <token>` | `HTTP 200` + full stats JSON |

The returned JSON contained every field the widget maps (`running`, `stopped`, `paused`, `total`, `cpu`, `memory`, `images`, `volumes`, `events_today`, `pending_updates`, `stacks`), confirming the token path returns the same shape as the existing username/password path.

---

_Disclosure: this PR was drafted with the help of an AI coding assistant. All changes were reviewed by me and verified working against a real Dockhand instance as described above._

### DIFF
--- a/docs/widgets/services/dockhand.md
+++ b/docs/widgets/services/dockhand.md
@@ -5,10 +5,22 @@ description: Dockhand Widget Configuration
 
 Learn more about [Dockhand](https://dockhand.pro/).
 
-Note: The widget currently supports Dockhand's **local** authentication only.
+Authenticate with either a Dockhand [API token](https://dockhand.pro/manual/#api-authentication) (recommended) or Dockhand's local username/password. A token is scoped to the user that created it and can be revoked without changing your password, so it's preferable to storing account credentials in your config. If both are provided, the token is used.
 
 **Allowed fields:** (max 4): `running`, `stopped`, `paused`, `total`, `cpu`, `memory`, `images`, `volumes`, `events_today`, `pending_updates`, `stacks`.
 **Default fields:** `running`, `total`, `cpu`, `memory`.
+
+Using an API token:
+
+```yaml
+widget:
+  type: dockhand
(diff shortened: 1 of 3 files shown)


shamoon:
Sigh, AI is making OSS much less fun. This diff is about 100x longer than it needs to be. Rather than going through all 60 lines or having Claude do another turn, I'll take care of it, thanks.

PR #6866, opened 2026-07-13
Fix: auto-scale binary rate units instead of pinning to MiB/s

TowyTowy:
## Proposed change

The `rate` i18n formatter hardcoded the unit index to `2` (MiB/s) whenever `binary: true`, so binary byte/bit rates never scaled with the value. The qbittorrent widget uses this via `common.bibyterate`, so a 15 KiB/s download rendered as `0.0 MiB/s`, 300 KB/s as `0.3 MiB/s`, and 2 GiB/s as `2,048.0 MiB/s`.

The unit index is now computed from the value with the correct base (`k` is already `1024` for binary / `1000` for decimal), matching the existing non-binary path. The regression was introduced in "Fix rate unit displays" (712fbb53).

Before → after (binary byte rate): `15 KiB/s`: `0.0 MiB/s` → `15.0 kiB/s`; `800 KiB/s`: `0.8 MiB/s` → `800.0 kiB/s`; `2 GiB/s`: `2,048.0 MiB/s` → `2.0 GiB/s`. Decimal and bit paths are unchanged; the full `pnpm test` suite passes (1467 tests).

## Type of change

## Checklist

- [x] Code style checks pass (prettier + eslint) on the changed file.
- [x] AI tool disclosure: this fix was prepared with AI assistance (Claude) and reviewed and verified by me before submitting.

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -136,7 +136,7 @@ module.exports = {
 
           const dm = options.decimals ? options.decimals : 0;
 
-          const i = options.binary ? 2 : Math.floor(Math.log(value) / Math.log(k));
+          const i = Math.floor(Math.log(value) / Math.log(k));
 
           const formatted = new Intl.NumberFormat(lng, { maximumFractionDigits: dm, minimumFractionDigits: dm }).format(
             parseFloat(value / k ** i),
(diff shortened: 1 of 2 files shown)


shamoon (changes requested):
The hard-coded binary exponent does appear incorrect, and the proposed behavior matches qBittorrent’s historical dynamic scaling. However, this PR has no linked report and adds no regression test; the changed config is excluded from coverage. The change also affects binary bit rates despite the description saying bit paths are unchanged, and the new exponent should be bounded for sub-unit and oversized values. Please add focused tests, and handle exponent bounds before merge.

shamoon (approved):

PR #6874, opened 2026-07-16
fix(calendar): show all days of multi-day ical events

aschaber1:
> Note: This PR was developed with the assistance of OpenCode using Claude Opus 4.8.

## Proposed change

Multi-day iCal events were only displayed on their first day. In `getOcurrencesFromRange`, a non-recurring event returned only `event.dtstart`, so a multi-day all-day event (for example multi-day PTO from an HR calendar) produced a single calendar entry. The parsed `dtend` was only used as a range filter and never to expand the event across the days it spans.

This PR expands each event into one occurrence per day it covers:

- All-day (date-only) events treat `DTEND` as exclusive per the iCal spec, so `DTSTART:20990101` / `DTEND:20990104` correctly covers Jan 1, 2 and 3.
- Timed events ending exactly at midnight are treated as exclusive too, so an event from Monday 20:00 to Tuesday 00:00 only occupies Monday.
- The original start time is preserved on the event's first day (used by the `showTime` option); subsequent days are plain dates.
- Recurring events are expanded by their duration as well, so multi-day recurring events also render every day.
- The non-recurring range filter was changed to an overlap check so partially visible multi-day events still render their visible days.
- The per-day expansion loop is clamped to the visible range instead of iterating the full event duration, so an event with a far-future `DTEND` no longer performs tens of thousands of wasted iterations.

No changes were needed in the calendar views: each expanded day flows through the existing per-occurrence rendering path.

### Testing

Added unit tests to `src/widgets/calendar/integrations/ical.test.jsx` covering:

- All-day multi-day expansion (one entry per day, `DTEND` exclusive)
- Clipping a multi-day event to the visible range
- Yearly recurring single-day events
- Timed event ending at midnight not spilling into the next day
- An event starting before the window with a far-future end date
- A performance guard asserting the day loop stays bounded to the visible window

All calendar tests pass and lint is clean.

Closes # (issue)

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/src/widgets/calendar/integrations/ical.jsx
+++ b/src/widgets/calendar/integrations/ical.jsx
@@ -98,13 +98,14 @@ export default function Integration({ config, params, setEvents, hideErrors, tim
 
       const lastDay = end.clone();
       lastDay.isDate = true;
-      if (start.isDate) {
-        // All-day events have an exclusive DTEND per the iCal spec, so step
-        // back one day to reach the last day actually covered by the event.
-        lastDay.day -= 1;
-      } else if (end.hour === 0 && end.minute === 0 && end.second === 0 && end.compare(start) > 0) {
-        // A timed event ending exactly at midnight ends the instant the next
-        // day begins, so it does not actually occupy that day.
+      // A timed event ending exactly at midnight ends the instant the next day
+      // begins, so that day is not actually covered by the event.
+      const endsAtMidnight =
+        !end.isDate && end.compare(start) > 0 && end.hour === 0 && end.minute === 0 && end.second === 0;
+
+      if (start.isDate || endsAtMidnight) {
+        // All-day events have an exclusive DTEND per the iCal spec; likewise a
+        // midnight end is exclusive. Step back to the last day actually covered.
         lastDay.day -= 1;
       }
       // Guard against zero/negative spans (e.g. same-day timed events, where
--- a/src/widgets/calendar/integrations/ical.jsx
+++ b/src/widgets/calendar/integrations/ical.jsx
@@ -102,6 +102,10 @@ export default function Integration({ config, params, setEvents, hideErrors, tim
         // All-day events have an exclusive DTEND per the iCal spec, so step
         // back one day to reach the last day actually covered by the event.
         lastDay.day -= 1;
+      } else if (end.hour === 0 && end.minute === 0 && end.second === 0 && end.compare(start) > 0) {
+        // A timed event ending exactly at midnight ends the instant the next
+        // day begins, so it does not actually occupy that day.
+        lastDay.day -= 1;
       }
       // Guard against zero/negative spans (e.g. same-day timed events, where
       // truncating to day granularity leaves start === end): emit the start day.
--- a/src/widgets/calendar/integrations/ical.jsx
+++ b/src/widgets/calendar/integrations/ical.jsx
@@ -82,10 +82,63 @@ export default function Integration({ config, params, setEvents, hideErrors, tim
     const rangeStart = ICAL.Time.fromJSDate(startDate.toJSDate());
     const rangeEnd = ICAL.Time.fromJSDate(endDate.toJSDate());
 
+    // Expands a single (occurrence) start into one entry per day it spans,
+    // clipped to the visible range. `duration` is the length of the event.
+    // For all-day (date-only) events, DTEND is exclusive per the iCal spec,
+    // so the last displayed day is the day before dtend.
+    const expandDays = (start, duration, occurrences) => {
+      // Compute the last covered instant, then normalize to day granularity so
+      // the per-day loop and range clipping work identically for all-day and
+      // timed events (downstream views compare at day granularity anyway).
+      const end = start.clone();
+      end.addDuration(duration);
+
+      const startDay = start.clone();
+      startDay.isDate = true;
+
+      const lastDay = end.clone();
+      lastDay.isDate = true;
+      if (start.isDate) {
+        // All-day events have an exclusive DTEND per the iCal spec, so step
+        // back one day to reach the last day actually covered by the event.
+        lastDay.day -= 1;
+      }
+      // Guard against zero/negative spans (e.g. same-day timed events, where
+      // truncating to day granularity leaves start === end): emit the start day.
+      if (lastDay.compare(startDay) < 0) {
+        lastDay.day = startDay.day;
+        lastDay.month = startDay.month;
+        lastDay.year = startDay.year;
+      }
+
+      const rangeStartDay = rangeStart.clone();
+      rangeStartDay.isDate = true;
+      const rangeEndDay = rangeEnd.clone();
+      rangeEndDay.isDate = true;
+
+      const firstDayNum = startDay.dayOfYear();
+      const firstYear = startDay.year;
+      for (const day = startDay; day.compare(lastDay) <= 0; day.day += 1) {
+        const isFirst = day.year === firstYear && day.dayOfYear() === firstDayNum;
+        if (day.compare(rangeStartDay) < 0 || day.compare(rangeEndDay) > 0) {
+          continue;
+        }
+        // Preserve the original start time on the first day so timed events keep
+        // their time-of-day (used by the `showTime` option); subsequent days of
+        // a multi-day event are represented as plain dates.
+        occurrences.push(isFirst ? start.clone() : day.clone());
+      }
+    };
+
     const getOcurrencesFromRange = (event) => {
+      const duration = event.dtend.subtractDate(event.dtstart);
+
       if (!event.rrule) {
-        if (event.dtstart.compare(rangeStart) >= 0 && event.dtend.compare(rangeEnd) <= 0) {
-          return [event.dtstart];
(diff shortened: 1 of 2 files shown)


shamoon:
Unless you can explain exactly what this PR does and what the ramifications are, please close it. The calendar code is complex and fragile, no one has reported an issue and I’m not willing to stake the functionality on a claude PR. Thanks

aschaber1:
> Unless you can explain exactly what this PR does and what the ramifications are, please close it. The calendar code is complex and fragile, no one has reported an issue and I’m not willing to stake the functionality on a claude PR. Thanks

Of course, I understand your concern. My issues (and if you'd like I can open issues on GitHub) are:
1. with multi-day events. For example, our company calendar has multi-day events for events like PTO, ... and
2. The filter required the whole event (start AND end) to sit inside the ±3-month window, so an event overlapping either boundary was dropped entirely. 

so the changes are:

## 1. multi day events

However right now only the first day of the multi-day event is displayed:

```js
if (event.dtstart.compare(rangeStart) >= 0 && event.dtend.compare(rangeEnd) <= 0) {
  return [event.dtstart];   // only ever the start day
}
```

which causes the calendar to look like this:
image

however it should look like this (yeah it's busy, but that's the truth :D )
image

codewise the change is to expand each event into one entry per covered day, clipped to the visible window. treat DTEND as exclusive for all-day events and midnight-ending timed events. apply the same expansion to each recurrence.

## 2. long event filter update

Range filter changed from "fully contained" to "overlaps." This means events that exceed the ±3-month boundary on either side, which were previously dropped, now appear.

## ramifications

here are the scenarios I looked at:

Changed:
1. Multi-day events now render on every day instead of just day one. This is the intended fix but IS a visible behavior change for anyone who has multi-day events.
1. Recurring events now expand by duration (multi-day recurrences show all days). Single-day recurrences unchanged.
1. Range filter changed from "fully contained" to "overlaps." This means events that straddle the ±3-month boundary, which were previously dropped, now appear.

Unchanged:
1. Single-day timed events: unchanged (1 entry). Verified by the pre-existing test still passing.
1. All-day single-day events: effectively unchanged (1 entry), but now go through the exclusive-DTEND logic.

## Known caveats
- Date-only events convert via ICAL.Time.toJSDate(), which yields local-midnight. The expansion inherits this existing conversion; I did not change timezone handling.
- The "timed event ending exactly at midnight is exclusive" check reads the end time in the event's own timezone, which may differ from the display timezone. This is an edge case I'm happy to revisit.

shamoon:
That is certainly a regurgitation of Claude's output. We will decline the PR as again, I am not comfortable with your process. Next time perhaps just raise the problem youre having rather than opening a PR. Here's Codex's explanation of why this PR is not a good idea. As I said, you do not understand the ramifications of the changes.

> The core diagnosis is likely real, but this implementation is not safe to merge. I would request changes or close it and ask for a much smaller replacement.
> 
> ## Blocking findings
> 
> 1. Missing `DTEND` turns ordinary events/tasks into events spanning until “now”
> 
> `buildEvent()` substitutes `ICAL.Time.now()` when `DTEND`, `DUE`, or `COMPLETED` is absent. The new code then treats `dtend - dtstart` as a real duration and expands it across every day.
> 
> A valid past `VEVENT` containing only:
> 
> ```ical
> DTSTART:20260701T090000Z
> ```
> 
> will therefore appear on every day from July 1 until today. A recurring event without `DTEND` is worse: that accidental duration is applied to every recurrence. `VTODO` items without a due date have the same problem.
> 
> This is not malformed input. RFC 5545 explicitly defines a date-only `VEVENT` without `DTEND`/`DURATION` as one day and a date-time event without them as zero-duration. It also permits `DURATION` instead of `DTEND`, which the integration ignores. [[RFC 5545](https://www.rfc-editor.org/rfc/rfc5545.html)](https://www.rfc-editor.org/rfc/rfc5545.html)
> 
> The dangerous calculation is at [[ical.jsx lines 141–166](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L141-L166)](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L141-L166). This alone is sufficient to reject the PR.
> 
> 2. Day expansion is performed in one timezone and rendered in another
> 
> The implementation converts the event start/end to date-only `ICAL.Time` values in the event’s timezone, then calls `toJSDate()`. Rendering groups that JavaScript date by the browser-local day. The supplied widget `timezone` is never used by the new logic.
> 
> For example, in Los Angeles, an expanded `2099-01-06` UTC date converts to January 5 at 16:00 local. This can:
> 
> - put continuation entries on the previous displayed day;
> - produce two copies on one day;
> - omit the actual final local day;
> - make the midnight-exclusive rule incorrect from the display timezone’s perspective.
> 
> The problematic normalization is [[lines 89–137](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L89-L137)](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L89-L137). Every new test uses UTC input and never verifies a non-UTC display timezone, despite passing `timezone="utc"` to code that does not consult that prop.
> 
> 3. It invents semantics for invalid and negative spans
> 
> The “zero/negative span guard” silently converts any negative duration into a one-day event:
> 
> ```js
> if (lastDay.compare(startDay) < 0) {
>   lastDay.day = startDay.day;
>   ...
> }
> ```
> 
> That does not merely defend against truncation. It also hides reversed dates and the missing-`DTEND` problem above. The test for a zero-length date-only event codifies invented behavior instead of testing RFC behavior. See [[lines 111–117](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L111-L117)](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L111-L117).
> 
> 4. Range-end semantics are internally inconsistent
> 
> Non-recurring events now use `dtstart <= rangeEnd` and expansion includes `rangeEnd`, while recurring occurrences use `next < rangeEnd`.
> 
> Consequently:
> 
> - a non-recurring event can be emitted on the end boundary;
> - the identical recurring occurrence is excluded;
> - the clipping test explicitly expects February 1 even though the supplied end is February 1 at midnight.
> 
> See [[lines 130–158](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L130-L158)](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L130-L158). The PR needs to establish whether the range is closed or half-open and apply that consistently.
> 
> ## Significant downstream ramifications the author missed
> 
> - Agenda `maxEvents` counts expanded entries, not distinct events. One long event can consume the entire agenda and hide later events.
> - Continuation entries are stored as midnight dates. With `showTime`, the existing view will display them as `00:00`; “plain dates” do not survive in the output model.
> - Continuation days are considered completed immediately after midnight because completion is calculated from the synthetic occurrence timestamp, not the event end. See [[lines 192–204](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L192-L204)](https://github.com/gethomepage/homepage/blob/dc671307d2c0bb9705d7aca3133e456f1fe1f58d/src/widgets/calendar/integrations/ical.jsx#L192-L204).
> - The same UID and timestamp are used as the key. Overlapping recurrences expanded onto the same date can overwrite one another.
> - The change applies indiscriminately to `VTODO`, although the reported problem concerns multi-day `VEVENT`s.
> - Recurrence duration exceptions and `RDATE` period-specific durations are not modeled, yet the PR presents recurrence expansion as generally solved.
> 
> ## Scope and test-quality assessment
> 
> The underlying regression is plausible and historically supported: Homepage previously had explicit multi-day expansion, which was lost during the move to `ical.js`. So I would not claim the initial multi-day diagnosis is false.
> 
> But the proposed response is wildly disproportionate:
> 
> - 68 production lines and 467 test lines;
> - five commits generated within roughly 45 minutes;
> - unrelated VTODO coverage added to satisfy Codecov;
> - a brittle “performance test” spying on every `ICAL.Time.compare()` call;
> - no tests for missing `DTEND`, `DURATION`, DST, differing event/display timezones, `VTODO` without `DUE`, range-end consistency, or agenda behavior.
> 
> The tests mostly restate the implementation’s assumptions. They do not validate interoperability or downstream UI behavior.

aschaber1:
Okay, thank you for your feedback and looking into this.

PR #6930, opened 2026-08-02
Fix: Disable using proxies in Docker HEALTHCHECK

Bytenka:
Add the `-Y off` flag to the container HEALTHCHECK wget directive to disable using proxies, even if set in the environment.

## Proposed change

The current HEALTHCHECK command uses the Busybox wget binary. This binary supports the `http(s)_proxy` environment variables, but not `no_proxy`. In environments where a mandatory proxy server is used and `no_proxy=127.0.0.1` is set to exclude loopback traffic, the HEALTHCHECK will incorrectly fail.

Disabling the use of proxies within the wget call fixes the issue. This is safe because the HEALTHCHECK is executed in the container, and only wants to reach 127.0.0.1. Having proxies enabled here is unnecessary.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ ENV PORT=3000
 EXPOSE $PORT
 
 HEALTHCHECK --interval=10s --timeout=3s --start-period=20s \
-  CMD wget --no-verbose --tries=1 --spider http://127.0.0.1:$PORT/api/healthcheck || exit 1
+  CMD wget --no-verbose --tries=1 --spider -Y off http://127.0.0.1:$PORT/api/healthcheck || exit 1
 
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["node", "server.js"]


shamoon:
Hi, thanks. Since:

1. No one has [ever] reported this and
2. This can already be overridden per deployment in your own Compose file

Before changing the default image behavior to a flag that only applies to a tiny subset of users, could you provide a minimal reproduction or healthcheck output demonstrating the failure with http_proxy and no_proxy=127.0.0.1? You could also make this more general with e.g. `CMD http_proxy= https_proxy=...`

Bytenka:
Sure! Let me turn this into a proper bug report then.

Here's the environment of the container (I use podman instead of docker, this works the same):
```
$ podman exec -it homepage env
HTTPS_PROXY=<redacted>
no_proxy=localhost,127.0.0.1,::1,<redacted>
NODE_VERSION=22.22.3
YARN_VERSION=1.22.22
HOSTNAME=::
SHLVL=1
PORT=3000
HOME=/root
NO_PROXY=localhost,127.0.0.1,::1,<redacted>
PGID=1000
container=podman
https_proxy=<redacted>
http_proxy=<redacted>
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PUID=1000
PWD=/app
HTTP_PROXY=<redacted>
NODE_ENV=production
HOMEPAGE_ALLOWED_HOSTS=*
```
I have `http_proxy`, `https_proxy`, `HTTP_PROXY`, and `HTTPS_PROXY` all set to the same value.
I have `no_proxy` and `NO_PROXY` set to ignore loopback URIs.

From `podman inspect`, I grab the current HEALTHCHECK comment, and run it in the container:
```
$ podman exec -it homepage sh
/app # 
/app # wget --no-verbose --tries=1 --spider http://127.0.0.1:$PORT/api/healthcheck
Connecting to <redacted> (<redacted>)
wget: server returned error: HTTP/1.1 403 Forbidden
```

Since `http_proxy` is defined, wget picks it up and tries to connect to the proxy. The proxy refuses because 127.0.0.1 is not allowed, as it would mean connecting to the loopback interface _of the proxy server itself_.

Running wget with the `http_proxy` variable cleared works as expected.
```
/app # http_proxy= wget --no-verbose --tries=1 --spider http://127.0.0.1:$PORT/api/healthcheck
Connecting to 127.0.0.1:3000 (127.0.0.1:3000)
remote file exists
```

This is equivalent to running wget with the `-Y off` flag to turn off all proxies:
```
/app # wget --no-verbose --tries=1 --spider -Y off http://127.0.0.1:$PORT/api/healthcheck
Connecting to 127.0.0.1:3000 (127.0.0.1:3000)
remote file exists
```

The container resolves wget from the PATH to the BusyBox version, as seen in the help:
```
/app # wget --help
BusyBox v1.37.0 (2025-12-16 14:19:28 UTC) multi-call binary.
...
```

Now, since the `no_proxy` variable is defined and 127.0.0.1 is listed, wget should bypass the proxy in this case, but it does not. This is because the wget implementation of BusyBox does not support parsing `no_proxy` at all. This can be checked here: https://github.com/vda-linux/busybox_mirror/blob/master/networking/wget.c
For reference, here's a extract of the manpage for the full wget version:
```
ENVIRONMENT
     Wget supports proxies for both HTTP and FTP retrievals.  The standard way to specify proxy location, which Wget recognizes, is using the following environment variables:

     http_proxy
     https_proxy
         If set, the http_proxy and https_proxy variables should contain the URLs of the proxies for HTTP and HTTPS connections respectively.

     ftp_proxy
         This variable should contain the URL of the proxy for FTP connections.  It is quite common that http_proxy and ftp_proxy are set to the same URL.

     no_proxy
         This  variable  should contain a comma-separated list of domain extensions proxy should not be used for.  For instance, if the value of no_proxy is .mit.edu, proxy will not be
         used to retrieve documents from MIT.
```

Changing the default behavior of the command to add `-Y off` means that this call to wget will always ignore the `http_proxy` environment variable. There is no impact to anything other than the HEALTHCHECK command itself. And since the command pings 127.0.0.1, the only affected users are those how rely on the _proxy server_ to allow CONNECT to _it's own_ 127.0.0.1:3000, and then expect some other service to answer with HTTP 200 on the `/api/healthcheck` location.
Either this affects no one, or the people affected have their proxies severely misconfigured. Besides, even if they made it work with proxies enabled (ie. the proxy and homepage are on the same server/network namespace, the container uses `--network=host`, and the proxy allows CONNECT to 127.0.0.1 on port 3000), it will also work without, no change needed. And for the majority of users which don't use any kind of proxy, this change doesn't affect them at all.

While overriding this locally is possible, it wouldn't be solved with a special CMD instruction, since this is related to the HEALTHCHECK setup. Besides, having `http_proxy` defined for the CMD is desirable for some homepage features. The HEALTHCHECK only runs on the loopback interface _inside_ the container, so it never needs access to external resources, unlike the CMD.

Sure, I could make a custom container image with just the HEALTHCHECK command edited to have this fixed, but that would imply maintenance on my side for a change that, to me, definitely looks like bug.
I would have been the first one to open an issue about this here, but I figured I'd just make a pull request directly. This fix very similar to what already happened on other projects, for reference:
- https://github.com/Difegue/LANraragi/issues/1040
- https://github.com/google/cadvisor/pull/3580

shamoon (approved):
👍

PR #6963, opened 2026-08-07
Performance: reduce calls to kubernetes api-server

emouawad:
## Proposed change

⚠️ Note: RBAC: A ClusterRole / ClusterRoleBinding is required now when gateway: true - whereas a namespaced RBAC RoleBinding was enough.

Closes #5830 
Closes #5969

This is a performance enhancement that switches from looping on Kubernetes namespaces then calling get CRD for each namespace to using 1 API call to list HttpRoute CRD on all the cluster.

For a cluster that has 100+ namespaces this is noticeable as the CPU / Memory in the older implementation was 3-4 times greater and homepage was lagging / taking longer to respond.

image

## Type of change

Enhancement (non-breaking change which fixes performance on cluster with lots of namespaces)

## Checklist:

- No configuration change is needed
- RBAC: A ClusterRole / ClusterRoleBinding is required now when `gateway: true` - whereas a namespaced RBAC RoleBinding was enough.
- Unit tests updated and working
- Tested on a live cluster that had the previous issue

### DIFF
--- a/src/utils/kubernetes/httproute-list.js
+++ b/src/utils/kubernetes/httproute-list.js
@@ -7,49 +7,29 @@ const logger = createLogger("httproute-list");
 const kc = getKubeConfig();
 
 export default async function listHttpRoute() {
-  const crd = kc.makeApiClient(CustomObjectsApi);
-  const core = kc.makeApiClient(CoreV1Api);
   const { gateway } = getKubernetes();
-  let httpRouteList = [];
 
   if (gateway) {
-    // httproutes
-    const getHttpRoutes = async (namespace) =>
-      crd
-        .listNamespacedCustomObject({
+    const crd = kc.makeApiClient(CustomObjectsApi);
+
+    const httpRoutes = await crd
+        .listClusterCustomObject({
           group: HTTPROUTE_API_GROUP,
           version: HTTPROUTE_API_VERSION,
-          namespace,
           plural: "httproutes",
         })
         .then((response) => {
-          return response.items;
+          logger.debug("Retrieved httproutes: %d", response?.items?.length ?? 0);
+          return response?.items;
         })
         .catch((error) => {
           logger.error("Error getting httproutes: %d %s %s", error.statusCode, error.body, error.response);
           logger.debug(error);
           return null;
         });
-    // namespaces
-    const namespaces = await core
-      .listNamespace()
-      .then((response) => response.items.map((ns) => ns.metadata.name))
-      .catch((error) => {
-        logger.error("Error getting namespaces: %d %s %s", error.statusCode, error.body, error.response);
-        logger.debug(error);
-        return null;
-      });
 
-    if (namespaces) {
-      const httpRouteListUnfiltered = await Promise.all(
-        namespaces.map(async (namespace) => {
-          const httpRoutes = await getHttpRoutes(namespace);
-          return httpRoutes;
-        }),
-      );
-
-      httpRouteList = httpRouteListUnfiltered.flat().filter((httpRoute) => httpRoute);
-    }
+    return httpRoutes;
   }
-  return httpRouteList;
+
+  return [];
 }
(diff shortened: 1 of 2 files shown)


emouawad:
@shamoon Can you look at this?
Automation closed it but i believe this is an important performance enhancement.
I fixed the linting issue in my branch, if this PR is reponed i can continue working on it to pass the requirements.
Thanks

shamoon:
You triggered all kinds of AI slop flags, including committing with different users, bypassing the issue template, etc.

src/utils/kubernetes/httproute-list.js
```diff
@@ -7,49 +7,29 @@ const logger = createLogger("httproute-list");
...
+    const crd = kc.makeApiClient(CustomObjectsApi);
+
+    const httpRoutes = await crd
+      .listClusterCustomObject({
+        group: HTTPROUTE_API_GROUP,
+        version: HTTPROUTE_API_VERSION,
+        plural: "httproutes",
+      })
+      .then((response) => {
+        logger.debug("Retrieved httproutes: %d", response?.items?.length ?? 0);
```
shamoon:
```suggestion
```

noise

src/utils/kubernetes/httproute-list.js
```diff
@@ -7,49 +7,29 @@ const logger = createLogger("httproute-list");
...
+        return response?.items;
```
shamoon:
```suggestion
        return response?.items ?? [];
```

this will cause a crash later

emouawad:
> You triggered all kinds of AI slop flags, including committing with different users, bypassing the issue template, etc.

Ah sorry :( my first contribution ever to any open source project; lol thanks for cutting me some slack!

Can you help me out - what is needed for the anti-slop check?

src/utils/kubernetes/httproute-list.test.js
```diff
@@ -87,20 +74,4 @@ describe("utils/kubernetes/httproute-list", () => {
     expect(logger.error).toHaveBeenCalled();
     expect(logger.debug).toHaveBeenCalled();
   });
-
-  it("skips namespaces whose httproute queries fail", async () => {
```
shamoon, line 91: Deleting a test with no explanation is an obvious red flag. Do you know what it did? What removing it does, or more specifically what this AI-generated PR changes about it?

shamoon (changes requested):

src/utils/kubernetes/httproute-list.test.js
emouawad: Yes  - this test is no longer needed.

Previous implementation was looping on each namespace: n namespaces meant n api calls to kubernetes api server 

This implementation isn't using namespaces at all therefore this case is irrelevant as a test

shamoon: Agreed the test can't survive the refactor, my point is about what it was guarding. This changes behavior for setups granting httproute access via namespaced RoleBindings instead of a ClusterRoleBinding: they go from partial results to an empty list. Yes the documented setup uses a ClusterRole, so that might be an acceptable trade off for the performance improvement but 'no configuration change is needed' isnt really accurate, and should at least be noted in the PR

shamoon:
> Can you help me out - what is needed for the anti-slop check?

Committing as yourself, not as some weird other presumably LLM user, using the project's issue template, all of the basic etiquette that worked with OSS for many years before LLMs started generating... you know what.

src/utils/kubernetes/httproute-list.test.js
emouawad: Fair point, note this change aligns with the traefik implementation which is using listClusterCustomObject and as you mentioned the docs mention a ClusterRole/Binding

I will update my PR to reflect the RBAC change and update the docs?(https://github.com/gethomepage/homepage/blob/dev/docs/installation/k8s.md).

shamoon: Since it’s probably an edge case just noting it in the PR will suffice

emouawad: I have update the PR description as suggested

> RBAC: A ClusterRole / ClusterRoleBinding is required now when `gateway: true` - whereas a namespaced RBAC RoleBinding was enough.

Let me know if anything else is still needed - thanks

shamoon (approved):

PR #7054, opened 2026-08-24
Fix/Improvement for calendar widget not being able to reference an *arr widget in the same service

MattFryer:
## Proposed change

When a calendar widget references an *arr integration (Sonarr, Radarr, Lidarr and Readarr), it can't reference their widget if they are located in the same service. This happens because the integration's proxy request always inherits the calendar's own index rather than the target widget's index. This means they always resolve back to the calendar widget itself. To fix this, the proxy request now includes the widget's expected type. getServiceWidget then falls back to a type match search when the existing index search doesn't find a match.

Although the calendar widget uses the generic proxy handler, I've added the type to all four handlers for consistency (generic, credentials, jsonrp and  synology).

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:

- [X] If applicable, I have added corresponding documentation changes. **N/A**
- [X] If applicable, I have added or updated tests for new features and bug fixes (see [testing](https://gethomepage.dev/widgets/authoring/getting-started/#testing)).
- [X] If applicable, I have reviewed the [feature / enhancement](https://gethomepage.dev/widgets/authoring/getting-started/#new-feature-guidelines) and / or [service widget guidelines](https://gethomepage.dev/widgets/authoring/getting-started/#service-widget-guidelines). **N/A**
- [X] I have checked that all code style checks pass using [pre-commit hooks](https://gethomepage.dev/widgets/authoring/getting-started/#code-formatting-with-pre-commit-hooks) and [linting checks](https://gethomepage.dev/widgets/authoring/getting-started/#code-linting).
- [X] In the description above I have disclosed the use of AI tools in the coding of this PR. **N/A**

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -39,7 +39,9 @@
         "wait": "Please wait"
     },
     "search": {
-        "placeholder": "Search…"
+        "placeholder": "Search…",
+        "provider": "Search engine: {{name}}",
+        "select_provider": "Select search engine"
     },
     "resources": {
         "cpu": "CPU",
--- a/src/components/widgets/search/search.jsx
+++ b/src/components/widgets/search/search.jsx
@@ -182,6 +182,7 @@ export default function Search({ options }) {
               focus:border-theme-500 dark:focus:border-white/50
               border border-theme-300 dark:border-theme-200/50"
               placeholder={t("search.placeholder")}
+              aria-label={t("search.placeholder")}
               onChange={(event) => {
                 setQuery(event.target.value);
               }}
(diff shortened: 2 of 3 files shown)


shamoon (changes requested):
Thanks for the PR, however:

1. This is too broad, we're not going to change all of these handlers for one rather specific issue.
2. **`widgets.find()` takes the first type match.** so now with two sonarr widgets on one service you can't address the second one, a regression.
3. `ical` integration sends `type=ical`, which never matches a widget type, so every ical refresh does a fallback lookup that will never succeed.

This should almost certainly be handled in src/widgets/calendar/component.jsx by adding index to the newly-created `widget`, like:

```js
  index: integration.service_name
    ? widget.index
    : (service.widgets?.find((w) => w.type === integration.type)?.index ?? widget.index),
```

MattFryer:
I'd convinced myself it couldn't be resolved in the calendar component. I'll take a look and give it a go. Thanks for the feedback.

Issue #7113, opened 2026-09-08
[Bug] Jellyfin widget 503 error after upgrading to Jellyfin 12.0 (removed /emby/ prefix)


diazpablo:
### Description

After upgrading Jellyfin to 12.0 (stable release, `jellyfin/jellyfin:latest`), the Jellyfin widget fails with a 503 error. Jellyfin 12.0 completely removed the legacy `/emby/` and `/mediabrowser/` route prefixes (confirmed in the official 12.0 release notes: https://jellyfin.org/posts/jellyfin-release-12.0/), but Homepage's Jellyfin widget still calls the hardcoded `emby/Items/Counts?api_key={key}` endpoint.

Other Jellyfin client projects (e.g. tsukinaha/tsukimi#676, nothing2obvi/pixelfin v1.0.6) hit the same issue and fixed it by switching to native, non-`/emby/`-prefixed endpoints and the modern `Authorization` header instead of `X-Emby-Token`/`api_key` query params.

### Steps to reproduce

1. Run Jellyfin 12.0 (`jellyfin/jellyfin:latest`)
2. Configure the Jellyfin widget in Homepage's `services.yaml`
3. Widget fails to load; logs show a 503 error

### Environment

- Homepage version: 2.2.0 (latest as of this report)
- Jellyfin version: 12.0 (stable)
- Installation: Docker Compose

### Expected behavior

Widget should use the native API endpoints (e.g. `Items/Counts` instead of `emby/Items/Counts`) so it works with both Jellyfin 12.0 and older 10.x versions, similar to how other Jellyfin API clients have adapted.

codebude:
## Confirmed + immediate workaround

Same problem here (Homepage 2.2.0, Jellyfin 12.0 from `jellyfin/jellyfin:latest`). The widget logs show it calling the removed legacy routes:

```
<jellyfinProxyHandler> HTTP Error 404 calling .../emby/Sessions?api_key=...
<jellyfinProxyHandler> HTTP Error 404 calling .../emby/Items/Counts?api_key=...
```

Jellyfin 12 returns 404 (not 401) for `/emby/...`, and the empty body surfaces in the widget UI as `API Error: Failed to execute json on Response: Unexpected end of JSON input`.

### Workaround available today — no Homepage update needed

The widget already ships parallel **V2 endpoints** that use the native (non-`/emby/`) paths, see `src/widgets/jellyfin/widget.js`:

```js
SessionsV2:  { endpoint: "Sessions" }
[...]

PR #7150, opened 2026-09-16
Enhancement: auto-fallback to Jellyfin v2 API when legacy emby/ routes 404

gautambaghel:
## Proposed change

Jellyfin 12 (10.12) removed the legacy `/emby/`-prefixed compatibility routes. #6252 added a version `2` widget mode that targets the new routes, but it is opt-in and defaults to `version: 1`. As a result, existing configurations start returning **404** on `emby/Items/Counts` and `emby/Sessions` after a Jellyfin upgrade until the user manually discovers and sets `version: 2`.

This change makes the proxy self-healing: when a legacy `emby/` request returns a `404`, it transparently retries the equivalent v2 endpoint (strips the `emby/` prefix and the `api_key` query param, since the `Authorization: MediaBrowser Token=...` header already carries auth). If the v2 route also 404s, the original response is preserved.

Existing `version: 1` configs keep working after a Jellyfin 12 upgrade with no config change, and `version: 2` still works as an explicit override that skips the fallback.

Closes the recurring 404 reports from users on Jellyfin 10.12 / 12.x.

## Type of change

## Checklist:

- [x] I have reviewed the service widget guidelines.
- [x] I have checked that all code style checks pass (eslint + prettier).
- [x] Tests added for both the successful fallback and the "v2 also missing" case (`src/widgets/jellyfin/proxy.test.js`).
- No AI-tool disclosure beyond standard editor assistance.


gautambaghel:
Closing this — the Jellyfin widget already supports the non-`/emby/` (native) endpoints via `version: 2`. Setting `version: 2` in the widget config uses the native `Items/Counts`/`Sessions` endpoints with an `Authorization: MediaBrowser Token="..."` header instead of the legacy `/emby/`-prefixed routes, which is the intended fix for Jellyfin >= 12 (10.12). Since this is already handled by existing configuration, the auto-fallback proposed here isn't necessary.

gautambaghel:
As mentioned in this issue https://github.com/gethomepage/homepage/issues/7113

PR #7167, opened 2026-09-22
Feature: Custom API information widget

spskeldon:
## Proposed change

Adds Custom API as an information widget, so a customapi block in widgets.yaml can show values from any JSON API in the header next to the other info widgets. Requested in discussion #6793.

The request runs on the server through a new route under /api/widgets, the same way the Glances info widget works. The URL, credentials, headers and request body never reach the browser; the route only takes the widget index. The mapping and formatting functions were private to the service widget and the header widget needs two of them, so they moved unchanged into a shared mappings.js that both widgets import. No other file imports it and the service widget's behaviour is the same. Resource now renders without an icon when none is given, and each mapping can name one.

Two choices you may want changed: refreshInterval is configurable with a 1 second floor, as in the service widget, and the service widget's special case that renders a mapped error field as data is not carried over.

Custom API information widget under a Glances widget, showing a CPU temperature with a thermometer icon

I drafted this patch with Claude. I set the requirements, reviewed every changed file, ran lint and the test suite, and tested the widget in a browser against a live Prometheus.

## Type of change

- [ ] New service widget
- [ ] Documentation only
- [ ] Other (please explain)

## Checklist:


shamoon:
Thanks, Im not comfortable with AI-generated 700+ line diffs from first-time contributors. 

Unfortunately, in the age of AI, a PR itself no longer provides any level of trust in authorship. Id suggest making first time contributions smaller scale in the future, with much clearer proof that it is more than prompt output.

Heres an LLM review, to match, with the problems, all suggestive of an LLM-crafted and minimally reviewed PR.

> Problems
> 
> 1. Refactor churn (about 400 of the lines)
> 
> It moves all four helpers into a new mappings.js: 150 lines deleted and 151 added.
> It then adds a 113-line mappings.test.js for code that didn't change, including getColor, which the info widget doesn't even use.
> 
> 2. Scope creep: per-mapping icons
> 
> Resource now renders without an icon, which touches a shared component and adds a new test.
> It adds a module-level Map cache that builds a component per icon string "so SWR refreshes don't remount". That's overengineering for an optional feature nobody asked for.
> 
> 3. The proxy logic is a second copy of genericProxyHandler, and it has already drifted
> 
> It loses the trailing-slash fix that generic.js has for customapi, so URLs ending in / behave differently in the service widget and the info widget.
> It rejects any status other than 200. Generic does too, but the behaviour and error shapes now live in two places.
> It's acceptable as a small standalone route. It just shouldn't pretend to match the service widget, which is what the docs claim.
> 
> 4. cleanWidgetGroups / getPrivateWidgetOptions
> 
> headers and requestBody are now stripped from every info widget, and getPrivateWidgetOptions returns headers/method/requestBody for every type.
> No existing info widget uses those keys, so nothing breaks. It's still a global change made for one widget.
> 
> 5. Small bugs
> 
> key={mapping.label}: two mappings with the same label, or with none, collide. Use the index.
> Test bloat: 142 lines of route tests, 97 lines of component tests, and 33 lines of helper tests. Several of them only test the mocks, for example checking that the URL is passed to httpProxy as a URL.
> The docs page repeats most of the service widget docs.

shamoon:
I added you to one of the commits for #7168

spskeldon:
Thanks @shamoon, I totally understand the hesitation, and appreciate the explanation and code review. Thank you also for #7168, I read it over and I think I understand what you did and really appreciate you taking the time to build it correctly!

PR #7176, opened 2026-09-23
Add Easypanel one-click deploy option to install docs

Ahson-Shaikh:
Hi — this is Ahson Shaikh, DevOps Engineer at Easypanel. [Easypanel](https://easypanel.io) is a self-hosted Docker deployment platform, and homepage has a one-click deployment template there: https://easypanel.io/templates/homepage

Added a new "Easypanel" card to the installation index, matching the existing Docker/Kubernetes/UNRAID/source cards, plus a dedicated docs page describing the template (persistent config volume + Docker socket mount, both handled automatically).

Deploy-tested via the template's own defaults before opening this PR — booted clean:

Homepage running in the Easypanel panel


shamoon:
Thanks but this is essentially advertising, and not something we'd be comfortable including as it implies a level of endorsement that is not equitable across providers.

Separately, if your hope is to turn a profit by deploying our humble FOSS app, there are ways you can contribute back, monetarily or otherwise!

